PR #7, opened 2019-07-05
Add initial run for local files

liamawhite:
The UX is pretty awful/has no value add right now. This is part of an iterative process towards a bare-bones `getenvoy list/pull/run`.

### DIFF
--- a/pkg/binary/run.go
+++ b/pkg/binary/run.go
@@ -17,14 +17,18 @@ package binary
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"syscall"
 )
 
+// Run execs the file at the path with the args passed
 func Run(path string, args []string) error {
 	if _, err := os.Stat(path); err != nil {
 		return fmt.Errorf("unable to stat %q: %v", path, err)
 	}
-	if err := syscall.Exec(path, append([]string{"envoy"}, args...), os.Environ()); err != nil {
+	_, filename := filepath.Split(path)
+	// #nosec -> passthrough by design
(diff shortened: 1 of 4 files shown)


pkg/binary/run.go
```diff
@@ -0,0 +1,31 @@
...
+	"fmt"
+	"os"
+	"syscall"
+)
+
+func Run(path string, args []string) error {
+	if _, err := os.Stat(path); err != nil {
+		return fmt.Errorf("unable to stat %q: %v", path, err)
+	}
+	if err := syscall.Exec(path, append([]string{"envoy"}, args...), os.Environ()); err != nil {
```
taiki45: Q: Can we make sure the `path` will always end with `envoy` filename? If not, we should use a filename from the `path` for the first element of the args for execve(2).

liamawhite: We should not assume this, good catch.

pkg/cmd/root.go
```diff
@@ -21,9 +21,11 @@ import (
...
 func NewRoot() *cobra.Command {
-	return &cobra.Command{
-		Use:   "getenvoy",
-		Short: "getenvoy",
-		Long:  "getenvoy",
-	}
+	return rootCmd
+}
+
+var rootCmd = &cobra.Command{
```
taiki45: nit: Are these changes intentional or mis-changes from rebasing? I think this should be separately committed for the record and for the better reviews (OK for this time).

liamawhite: These are intentional changes due to a rebase.

Issue #9, opened 2019-07-11
Enable users to list GetEnvoy builds using the CLI


liamawhite:
Depends on https://github.com/tetratelabs/getenvoy-ci/issues/163.

This is mostly done already [here](https://github.com/tetratelabs/getenvoy/tree/master/pkg/manifest). We just need to hook it up with a command.

Ensure to add a flag to allow a user to call out to any manifest location but default to ours on Bintray.

PR #21, opened 2019-07-19
Hook up list command

liamawhite:
Resolves https://github.com/tetratelabs/getenvoy/issues/9.


pkg/manifest/print.go
```diff
@@ -17,15 +17,21 @@ package manifest
...
 
+	"github.com/golang/protobuf/jsonpb"
 	"github.com/tetratelabs/getenvoy-package/api"
 )
 
-// PrettyPrint tabwrites the passed manifest to the passed writer
-func PrettyPrint(writer io.Writer, manifest *api.Manifest) error {
+// Print retrieves the manifest from the passed URL and writes it to the passed writer
+func Print(writer io.Writer, manifestURL string) error {
+	manifest, err := fetch(manifestURL)
```
taiki45: Hmm, I don't think fetching should be in the print package and the Print func... It seems not relevant to "print"s. What is the downside of the fetch package?

liamawhite: As a consumer of the manifest API I just want to print the manifest, how to retrieve the manifest is an implementation detail. It could be at a URL, it could be a local file, etc.

To make it clearer, I should rename the argument to just `manifestLocation`. Then, error on Non-URL inputs saying that only URL locations are supported currently.

taiki45: I see, then make sense to me 👍

Issue #10, opened 2019-07-11
Enable users to pull Envoy builds via an explicit manifest reference.


liamawhite:
`getenvoy fetch standard:1.11.0/debian`

See https://github.com/tetratelabs/getenvoy/blob/master/pkg/cmd/run.go#L33 for format. User's should default to their own OS but can optionally specify an operating system and version to pull that.

This should pull to  `~/.getenvoy/builds/filter-profile/version/os/os-version` if they request a specific OS e.g. `~/.getenvoy/builds/standard/1.11.0/ubuntu/xenial`

We should check if it's not already local and only attempt if not present.

PR #25, opened 2019-07-26
Move to new manifest api

liamawhite:
Resolves https://github.com/tetratelabs/getenvoy/issues/10. Depends on https://github.com/tetratelabs/getenvoy-package/pull/26.


pkg/binary/getenvoy/fetch.go
```diff
@@ -0,0 +1,103 @@
...
+	tarball := filepath.Join(dst, "envoy.tar.gz")
+	f, err := os.OpenFile(tarball, os.O_CREATE|os.O_WRONLY, 0600)
+	if err != nil {
+		return "", err
+	}
+	defer func() { _ = f.Close() }()
+
+	bar := progressbar.NewOptions64(resp.ContentLength, progressbar.OptionSetDescription("[Fetching Envoy]"))
+	out := io.MultiWriter(f, bar)
+	_, err = io.Copy(out, resp.Body)
```
taiki45: Q: don't we check a reponse status code before using a response body?

pkg/binary/getenvoy/fetch.go
```diff
@@ -0,0 +1,103 @@
...
+			if _, err := io.Copy(envoy, f); err != nil {
+				return err
+			}
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+	if !found {
+		return errors.New("unable to find Envoy binary in downloaded tarball")
```
taiki45: nit: Outputting tarball string (path?) can be useful.

pkg/binary/getenvoy/fetch.go
liamawhite: yes, we should!

pkg/binary/getenvoy/fetch.go
liamawhite: I don't think it's useful in this case. The defer ensures that the tarball is cleaned up. The URL would be useful but isn't in scope.

I could wrap any errors returned by this function to include the URL though.

go.mod
```diff
@@ -16,3 +17,5 @@ require (
 	github.com/tetratelabs/getenvoy-package v0.0.0-20190718134531-9487f25b3273
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
 )
+
+replace github.com/tetratelabs/getenvoy-package => ../getenvoy-package
```
taiki45: Q: is this on purpose?

liamawhite: Ah, nope!

This was so I could use my updates from the local file directory. It won't pass lint until I remove it so when the other PR is merged I'll fix.

PR #27, opened 2019-07-30
support run for reference and local file

liamawhite:
Resolves https://github.com/tetratelabs/getenvoy/issues/26


pkg/cmd/run.go
```diff
@@ -49,7 +51,12 @@ getenvoy run standard:1.10.1 -- --help
 			if err != nil {
 				return err
 			}
-			return runtime.Run(args[0], args[1:])
+			key, err := manifest.NewKey(args[0])
+			if err != nil {
+				log.Info("binary arg is not in the manifest reference format - attempting to run as path")
+				return runtime.RunPath(args[0], args[1:])
```
taiki45: Q: Do we need this polymorphism for `binary` CLI argument? How is this useful to users instead of directly running the binary? If so, I think this behavior has to be with another command-line argument/option like `--binary-path X`

liamawhite: IMHO, the polymorphism is a better UX. If we make it a flag there's confusion over what goes in the args.

It's not useful at the moment, but it will be when we add debugging, traffic int and bootstrapping functionality.

taiki45: Make sense, adding as a flag seems not good... I understand the use case for this, then I think we should implement this (run path feature) as another subcommand, not in the same command as the run command, like `run-path` command (the command name is terrible though, just for example). So that the argument of the run command will be more clear.

liamawhite: A separate command is less discoverable than polymorphism. 

We still have the same issue just shifted to subcommand. Instead of, what do I put in this arg? Its, which command should I use?

taiki45: TLDR; This case seems one of anti-pattern of implicit conversion, to me. WDYT?

Ah, my explanation was not clear. What I'm worrying/thinking about here is the current `run` feature and the current `run-path` feature are fundamentally different: the functionality is near like the run feature includes the run-path feature, but their interfaces are different with each other. In cases like this, the polymorphism in interfaces can be confusing in general. This is because the current code has to print an attention message "binary arg is not in...", and this shows the current run command is primarily for the run feature. The run-path feature is optional for the current run command and this situation is often confusing for users, in my past expriences.

In other words, let's think from the argument side: the argument above, a. "getenvoy binary selector" and, b. "the path to envoy binary", the (a) and (b) is fundamentally same? In my opinion, they are not same.

liamawhite: Ah, so you were suggesting making it explicit conversion originally? `getenvoy run path/to/envoy --path`

My counterpoint would be as a user I want to type as little as possible. This could be re-written in a way whereby that log is not required, but I don't think that's your point.  

The trade-offs here are verbosity, discoverability and comprehensibility.

`run`
Trades-off comprehensibility for conciseness and discoverability. Adding a `--path` flag adds verbosity in return for more comprehensibility.

`run && run-path`
Trades-off more verbosity and less discoverability for more comprehensibility.

My argument would be that in all situations other than a CLI, I would agree with you. Increased verbosity is worth the trade-off for comprehensibility. However, with CLIs, verbosity is much more important as I have to type everything out so often.

My suggestion would be that I write better usage documentation and error handling as part of the command to make it clear that there are two different options. Making it concise, discoverable and more comprehensible.

taiki45: OK, my point was not about explicit/implicit conversion, it's about a combination of polymorphism and implicit conversion of inputs. But, I don't think this discussion is much valuable to change the CLI interface since the tradeoff is okay to take. So let's try with the initial design and if it will cause problems, let's change it in v2 or something.

taiki45: FYI, @lizan @dio for any concerns/comments about this if any.

taiki45: Ah, I forgot, +1 for better usage documentation and error handling.

liamawhite: Agreed. We can evaluate everything before we do the first release.

PR #31, opened 2019-08-04
Pull binary if not local and AdminAPI tests

### LINKED ISSUE #12: Pull Envoy build if a user runs a manifest reference that is not already local
See https://github.com/tetratelabs/getenvoy/blob/master/pkg/cmd/run.go#L49

### LINKED ISSUE #32: On Termination GetEnvoy collects admin API data from Envoy.

liamawhite:
Resolves #12 and tests to resolve #32 as the former makes the latter easier.


pkg/binary/envoy/fetch_test.go
```diff
@@ -35,67 +35,94 @@ func TestRuntime_Fetch(t *testing.T) {
...
 		envoyLocation    string
+		libLocation      string
+		alreadyLocal     bool
 		responseStatus   int
 		wantErr          bool
+		wantServerCalled bool
 	}{
 		{
-			name:             "Downloads and untars envoy to local/key",
+			name:             "Downloads and untars envoy and libs to local/key/bin and local/key/lib",
```
lizan: nit: runtime libraries

PR #33, opened 2019-08-09
Access log and stderr capture

liamawhite:
Resolves https://github.com/tetratelabs/getenvoy/issues/34


pkg/binary/envoy/debug/log.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/tetratelabs/getenvoy/pkg/binary/envoy"
+	"github.com/tetratelabs/log"
+)
+
+// EnableEnvoyLogCollection is a preset option that registers collection of Envoy Access Logs
+//
+// This is not supported on non-Linux platforms as Envoy is not able to access /dev/stdout unless
+// *exec.Cmd.Stdout == os.Stdout. This makes it impossible to add the MultiWriter and start Envoy.
+//
+// The reason for this is not fully understood and not worth investing time yet just for macOS.
```
dio: Wild idea, probably we can also have an option on having the binary as ALS grpc server and stream the log there. Worth a TODO?

liamawhite: Yes, I think this is what we want to do eventually. We would need xDS mutation for it to work though.

PR #35, opened 2019-08-11
add node level ps collection

liamawhite:


lizan:
LGTM at code level. A high level question:

Have you considered using [psutil](https://github.com/giampaolo/psutil) alike library to collect this more programatically? Seems [this](https://github.com/shirou/gopsutil) is a golang equivalent.

liamawhite:
@lizan that is a much better idea 👍

PR #37, opened 2019-08-14
autobuild release on new tag

liamawhite:
I have no way of verifying if this works other than trying it... ¯\_(ツ)_/¯


dio:
Let's try it!

Issue #40, opened 2019-08-14
Collect status of network interfaces pre-termination


liamawhite:
This should be a [node level](https://github.com/tetratelabs/getenvoy/blob/master/pkg/binary/envoy/debug/node.go) collection. Output should look similar to (or the same as) `ifconfig`.
The aim is to keep external dependencies to a minimum, so [this library](https://godoc.org/github.com/shirou/gopsutil/net) should be used.

```sh
root@4f51979ee734:/# ifconfig
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 172.17.0.2  netmask 255.255.0.0  broadcast 172.17.255.255
        ether 02:42:ac:11:00:02  txqueuelen 0  (Ethernet)
        RX packets 12699  bytes 18362151 (18.3 MB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 6454  bytes 353781 (353.7 KB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.1  netmask 255.0.0.0
        loop  txqueuelen 1  (Local Loopback)
        RX packets 0  bytes 0 (0.0 B)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 0  bytes 0 (0.0 B)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```

ZackButcher:
If we use an SDK we need to be careful to make sure we fail gracefully if e.g. we don't have permissions to read the host level network config.

liamawhite:
Yup. By design, any errors returned by these functions are logged by the framework and then it continues.

PR #55, opened 2019-09-06
Implement network interfaces logging at exit

taiki45:
Signed-off-by: Taiki Ono <taiki@tetrate.io>

Fixes https://github.com/tetratelabs/getenvoy/issues/40

`Interfaces()` is just a wrapper of golang std's `Interfaces()` and it safely picks up interfaces. https://github.com/shirou/gopsutil/blob/v2.19.8/net/net.go#L212 And we can use JSON format to log interfaces at that time safely https://github.com/shirou/gopsutil/blob/v2.19.8/net/net.go#L193.

Output example:

```
[{"mtu":65536,"name":"lo","hardwareaddr":"","flags":["up","loopback"],"addrs":[{"addr":"127.0.0.1/8"},{"addr":"::1/128"}]},{"mtu":1500,"name":"eno1","hardwareaddr":"","flags":["up","broadcast","multicast"],"addrs":[{"addr":"192.168.11.7/24"},{"addr":""},{"addr":""},{"addr":""}]},{"mtu":1500,"name":"docker0","hardwareaddr":"","flags":["up","broadcast","multicast"],"addrs":[{"addr":"172.17.0.1/16"}]}]
```

### DIFF
--- a/pkg/binary/envoy/debug/node.go
+++ b/pkg/binary/envoy/debug/node.go
@@ -122,7 +122,7 @@ func safeProc(p *process.Process) *proc {
 }
 
 func networkInterfaces(r binary.Runner) error {
-	f, err := os.Create(filepath.Join(r.DebugStore(), "node/network_interface.txt"))
+	f, err := os.Create(filepath.Join(r.DebugStore(), "node/network_interface.json"))
 	if err != nil {
 		return fmt.Errorf("unable to create file to write network interface output to: %v", err)
 	}
--- a/pkg/binary/envoy/debug/node.go
+++ b/pkg/binary/envoy/debug/node.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"text/tabwriter"
 
+	"github.com/shirou/gopsutil/net"
 	"github.com/shirou/gopsutil/process"
 	"github.com/tetratelabs/getenvoy/pkg/binary"
 	"github.com/tetratelabs/getenvoy/pkg/binary/envoy"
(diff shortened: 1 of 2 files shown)


pkg/binary/envoy/debug/node.go
```diff
@@ -118,3 +120,19 @@ func safeProc(p *process.Process) *proc {
 		cmd:      cmd,
 	}
 }
+
+func networkInterfaces(r binary.Runner) error {
+	f, err := os.Create(filepath.Join(r.DebugStore(), "node/network_interface.txt"))
```
liamawhite: If the file is json can we make this `.json`?

pkg/binary/envoy/debug/node_test.go
```diff
@@ -41,3 +41,21 @@ func Test_ps(t *testing.T) {
 		}
 	})
 }
+
+func Test_network_inerfaces(t *testing.T) {
```
liamawhite: This test should be integrated with the one above as they do the same things. Add a table-driven test for each file and use only a single run.

taiki45: Right 👍

taiki45:
@liamawhite Good catch, updated the PR description.

liamawhite:
@taiki45 FYI, that output example isn't valid JSON. The array elements are missing trailing commas.

taiki45:
@liamawhite ah, I filtered hardware addresses and v6 addresses, so it's just my mistake. updated.

PR #58, opened 2019-09-06
WIP: Issue #1703 - envoy config dump - data obfuscation

jhengy:
# Task Accomplished
- function added to 
  - filter an array of logs by format string and a predefined map validFieldNames
  - hash the processed array of logs using sha256
  - return the processed logs as an array of strings

# Todos
- salting the hash
- Write test for the function


jhengy:
Problem with this PR, will close and submit a new one

PR #59, opened 2019-09-06
WIP: Issue #1703 - envoy config dump - data obfuscation

jhengy:
# Task Accomplished
- function added to 
  - filter an array of logs by `format` and a predefined map `containsPII`
  - hash the processed array of logs using sha256
  - return the processed logs as an array of strings and an error instance

# Todos
- salting the hash
- Write test for the function


pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,72 @@
...
+	fieldNames, ok := shell.Split(formatStr)
+
+	if !ok {
+		panic(fmt.Sprint("error in splitting format string: %s", formatStr))
+	}
+
+	testForValidLog :=  func(log string) bool {
+		fields, ok := shell.Split(log)
+		if !ok {
+			panic(fmt.Sprint("error in splitting log: %s", log))
```
ZackButcher: Like above, we should replace this with a `return []string{}, fmt.Errorf(...)`.

However, in this case we don't want to abort the entire function just because we can't process one log line. Instead, we should just log and `continue`:

```
if !ok {
	log.debugf("skipping log line because we could not split it: %q", log)
	continue
}
```

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		// TODO: consider using regex to check fields instead of merely verifiying the number of fields
+		return len(fields) == len(fieldNames)
+	}
+	validLogs := filter(logsArr, testForValidLog)
```
ZackButcher:
```suggestion
	valid := filter(logsArr, testForValidLog)
```

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// filter and hash each log in the array of valid logs with relevant fields
+	for i,_ := range(validLogs) {
```
ZackButcher: `range` is not a function: `for i, _ := range validLogs { ... }`

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,72 @@
+package debug
+
+import (
+	"fmt"
+	"bitbucket.org/creachadair/shell"
+	"crypto/sha256"
+)
+
+// filter the array of logs using formatStr and validFieldNames, hash the resulting array of logs and return the final array
```
ZackButcher: Idiomatic method and member comments in golang start with the name of the thing being documented. So this should be `// process filters the array of logs ...`. The linter didn't complain because this is not a public method, but convention we keep private methods matching public ones.

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,72 @@
...
+			panic(fmt.Sprint("error in splitting log: %s", log))
+		}
+		// TODO: consider using regex to check fields instead of merely verifiying the number of fields
+		return len(fields) == len(fieldNames)
+	}
+	validLogs := filter(logsArr, testForValidLog)
+
+	// filter and hash each log in the array of valid logs with relevant fields
+	for i,_ := range(validLogs) {
+		fields, ok := shell.Split(validLogs[i])
```
ZackButcher: rather than splitting once when we call `filter` then a second time when we process them here, let's just remove the filter and check that the log line is valid in this single loop: in this loop, just `parts, ok := shell.Split(log)` and `if !ok || !valid(fieldNames, parts) { log.debugf(...); continue }`. Remember that the number of logs can be very large, so we don't want to touch each log line more than once if we can avoid it.

pkg/binary/envoy/debug/pii.go
ZackButcher: The reason golang uses the `for index, element := range array` scheme is to avoid `array[index]` everywhere in the body of the for loop. In this case, you should update the loop like `for i, log := range logs` and replace all instances of `validLogs[i]` with `log`. In fact, I think you'll find that if you do this, you won't even need to declare the `i` variable at all, and you can just `for _, log := range logs { ... }`

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,72 @@
...
+		if !ok {
+			panic(fmt.Sprint("error in splitting a valid log: %s", validLogs[i]))
+		}
+
+		// filter fields by validFieldNames
+		filteredFields := []string{}
+		for j,_ := range(fields) {
```
ZackButcher: Same here: `for _, field := range fields {`

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,72 @@
...
+			if (validFieldNames[fieldNames[j]]) {
+				filteredFields = append(filteredFields,fields[j])
+			}
+		}
```
ZackButcher: We don't want to remove the PII fields then hash the entire log line, we only want to hash the PII fields. So this loop should look something like:

```go
for j, field := range fieldNames {
	if !containsPII(field) {
		// in golang, we prefer to keep the function as un-nested as possible. Write the other version and compare
		continue
	}
	// safe to index into fields with j because we checked fieldNames and fields have the same length above
	fields[j] = hash(fields[j])
}
processed = append(processed, strings.Join(fields))
```

ZackButcher: Also note that I flip this around and drive it by the format string, rather than driving it by the log line. This makes more sense, as you're matching up the log line to the format, not the format to the log.

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,72 @@
...
+	testForValidLog :=  func(log string) bool {
+		fields, ok := shell.Split(log)
+		if !ok {
+			panic(fmt.Sprint("error in splitting log: %s", log))
+		}
+		// TODO: consider using regex to check fields instead of merely verifiying the number of fields
+		return len(fields) == len(fieldNames)
+	}
+	validLogs := filter(logsArr, testForValidLog)
+
```
ZackButcher: Here is where we should declare our output variable. Traditionally, we'd either call this `out`, or something that made sense semantically. Your `processedLogs` is on the right track, but too long: it should be either `processed` or `out`.

```go
processed := make([]string, 0, len(validLogs))
```

Note that declaring it here means we can correctly-size the array from the get-go: we know exactly how many elements will be in it, so we can allocate enough memory up front and avoid having to copy the array while we're in this loop.

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,72 @@
...
+	for _,s := range strSlice {
+		if test(s) {
+			filteredStrSlice = append(filteredStrSlice, s)
+		}
+	}
+	return
+}
+
+// TODO: salting the hash
+func hash(s string) (hashedString string) {
```
ZackButcher: Here's another example of a downside of named outputs: your function is actually more verbose as a result. If you get rid of it, things are more simple:
```go
func hash(s string) string {
	h := sha256.New()
	h.Write()
	return string(h.Sum(nil))
}
```

Secondly, `Sum` takes a data buffer, so you can do this all in one go:
```go
func hash(s string) string {
	return string(sha256.New().Sum([]byte(s)))
}

// or it might be more clear as:

func hash(s string) string {
	h := sha256.New()
	return string(h.Sum([]byte(s)))
}
```

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,72 @@
...
+
+		// hash log
+		validLog = hash(validLog)
+		processedLogs = append(processedLogs, validLog)
+	}
+
+	return
+}
+
+func filter(strSlice []string, test func(string) bool) (filteredStrSlice []string) {
```
ZackButcher: No need for this if you merge the two loops together.

.gitignore
```diff
@@ -1,3 +1,4 @@
 vendor
 dist
 getenvoy
+.idea
```
ZackButcher: newline at end of file

pkg/binary/envoy/debug/pii.go
ZackButcher: One final note: stylistically I might actually not use `continue` here and tuck the logic into the `if`, since it winds up only being 1 line:
```go
for j, field := range fieldNames {
	if containsPII(field) {
		// safe to index into fields with j because we checked fieldNames and fields have the same length above
		fields[j] = hash(fields[j])
	}
}
processed = append(processed, strings.Join(fields))
```

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,72 @@
...
+
+// filter the array of logs using formatStr and validFieldNames, hash the resulting array of logs and return the final array
+func process(logsArr []string, formatStr string, validFieldNames map[string]bool) (processedLogs []string) {
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Println("process func panicking, recovering")
+		}
+	} ()
+
+	fieldNames, ok := shell.Split(formatStr)
```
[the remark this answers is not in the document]
jhengy: The reason of why I put `fieldNames` rather than `fields` is that I'm trying to differentiate between names of the field and values of the field, i.e. field name of the field value `[2019-09-05T17:53:36.908Z]` is `[%START_TIME%]`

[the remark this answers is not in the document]
jhengy: Am I understanding it wrongly?

pkg/binary/envoy/debug/pii.go
jhengy: Can I clarify if the final logs should have all the none PII fields removed?

ZackButcher: We want to call some function (our `hash` function for now) on each piece of the PII data and put the functions result into the log line instead of the original PII data. In the example I showed, we split the string into parts and replace the parts with PII with the result of calling `hash`.

In other words, given
```go
// Given: SOURCE_IP is PII, RESPONSE_CODE is not
fmtString := `%RESPONSE_CODE% %SOURCE_IP%`
logLines := []string{
	"200 127.0.0.1",
	"500 10.0.1.1",
	"404 192.168.17.0",
}

out := process(fmtString, logLines)
// should return the log lines with PII part scrambles
reflect.DeepEqual(out, []string{
	"200 AAB5=",
	"500 AS0978B",
	"404 BKL9P84",
})
```
(In a future iteration we'll extract out the call to `hash` so we can substitute in `hash` for other functions - for example replace IP addresses with `IP_ADDRESS` rather than random data - but don't worry about that yet.)

ZackButcher:
Looks like you're getting some other code changes mixed in with your branch (the template stuff) - we'll have to talk about a rebase based workflow. In short: you should always:

```sh
git checkout configdump_obfuscation
git fetch upstream/master
git rebase upstream/master
```
This will replay the changes in your branch over top of master. This is cleaner than merging master into your working changes. It might be easier to close this PR and re-open with the latest changes in a fresh branch without the different sets of commits from separate PRs. A side effect of this is changing your branches history; so when you push again, you'll have to force push back to origin.

lizan:
@ZackButcher I'd disagree with rebase and force-push, instead just do regular merge in working branch, at the end we will squash so it should be fine. Just always do:

```
git fetch upstream/master
git merge upstream/master
```

Btw, feel free to close and open another PR by clearing history.

pkg/binary/envoy/debug/pii.go
jhengy: I see, thanks for the detailed explanation

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,54 @@
...
+		if len(fieldValues) == len(fieldNames) {
+			requiredvalues := []string{}
+			// pick the PII fields and hash the fields
+			for j, name := range fieldNames {
+				if containsPII[name] {
+					hash, err := hash(fieldValues[j])
+					if err != nil {
+						return []string{}, fmt.Errorf("error in hashing the field: %s", fieldValues[j])
+					}
+					requiredvalues = append(requiredvalues, (hash))
```
ZackButcher: We don't want to print only the hashed PII fields, we want the entire log line but only the PII bits hashed. So rather than declare a new array here, just take the array you split up already (`fieldValues`), update only the PII part (`fieldValues[j] = hash(fieldValues[j])`, then slave that to output (`out = append(out, shell.Join(fieldValues))`).

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,54 @@
...
+			}
+			fmt.Println(shell.Join(requiredvalues))
+			out = append(out, shell.Join(requiredvalues))
+		}
+	}
+	return out, nil
+}
+
+// TODO: salting the hash
+func hash(s string) (string, error) {
```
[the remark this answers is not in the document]
ZackButcher: 🤦‍♂️`Sum` doesn't work like this and the implementation I'm suggesting isn't correct. You _do_ need to call `h.Write([]byte(s))` - sorry about that. However, you can see in the implementation that this `Write` can never error (in fact, I am not aware of an implementation of `hash.Hash` who's `Write` returns an error), so you can safely ignore it in this case. I'd suggest something like:

```go
func hash(s string) string {
	h := sha256.New()
	_, _ := h.Write([]byte(s)) //nolint: errcheck
	return string(h.Sum(nil))
}
```
(notice there is no space: `//nolint` *not* `// nolint`)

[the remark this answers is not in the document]
jhengy: Alright, thanks. Yep, I realize the problem without calling `h.Write([] byte(s))`, that's why I didn't change the implementation.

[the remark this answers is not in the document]
jhengy: Oh, thanks, I didn't know we can disable lint check for a particular line using `nolint: errcheck`

jhengy:
PR closed, please refer to PR #63

PR #60, opened 2019-09-10
change router mode into loadbalancer mode

liamawhite:

### DIFF
--- a/pkg/binary/envoy/controlplane/istio.go
+++ b/pkg/binary/envoy/controlplane/istio.go
@@ -125,7 +125,7 @@ func retrieveIPs() ([]string, error) {
 
 func istioNode(cfg *envoy.Config) string {
 	mode := cfg.Mode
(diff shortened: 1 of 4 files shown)


pkg/binary/envoy/controlplane/istio.go
```diff
@@ -124,8 +124,12 @@ func retrieveIPs() ([]string, error) {
 }
 
 func istioNode(cfg *envoy.Config) string {
+	mode := cfg.Mode
+	if mode == envoy.LoadBalancer {
+		mode = "router"
```
taiki45: Q: is this “router” correct?

liamawhite: Yes, Istio calls it router. We want to explicitly **not** use that terminology to avoid confusion.

taiki45: 👌👌👌

PR #63, opened 2019-09-10
Issue #1703 - envoy config dump - data obfuscation

jhengy:
# Task Accomplished
- struct `Filter` that describes elements for filtering and hashing of logs
- `process` method for `Filter` that processes logs based on the specific `Filter` instance
- package level function `ProcessLogs` that filters logs with a default `Filter` instance
- test cases for the newly added lines

# Todos
- salting the hash


pkg/binary/envoy/debug/pii_test.go
```diff
@@ -0,0 +1,102 @@
...
+			want: Out{logs: []string{VALID_LOG}, err: nil},
+		},
+	}
+
+	// defining subtest with t.Run
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			actualLogs, actualErr := ProcessLogs(tc.logs, tc.format, tc.containsPII)
+
+			// hash the desired fields for each log in wanted logs
```
jhengy: I have a question about what's the best way to define the expected output for my test cases in the table. Instead of defining the desired output of my test cases as hashed values, I did the logic of hashing the expected logs here. May I know if this may cause any problems? Is there a better way to do it? The obvious way is to put the hash of the relevant PII fields altogether but I feel that it would be too much manual work if there are many test cases.

lizan:
@jhengy re: lint error, run `go mod tidy` once and commit the change. then you will see real issues:

```
pkg/binary/envoy/debug/pii_test.go:72:41: Using the variable on range scope `tc` in function literal (scopelint)
                        actualLogs, actualErr := ProcessLogs(tc.logs, tc.format, tc.containsPII)
                                                             ^
pkg/binary/envoy/debug/pii_test.go:75:24: Using the variable on range scope `tc` in function literal (scopelint)
                        for i, log := range tc.want.logs {
                                            ^
pkg/binary/envoy/debug/pii_test.go:82:9: Using the variable on range scope `tc` in function literal (scopelint)
                                        if tc.containsPII[fieldName] {
                                           ^
pkg/binary/envoy/debug/pii_test.go:90:5: Using the variable on range scope `tc` in function literal (scopelint)
                                tc.want.logs[i] = shell.Join(fields)
                                ^
pkg/binary/envoy/debug/pii_test.go:93:26: Using the variable on range scope `tc` in function literal (scopelint)
                        errorDiff := cmp.Diff(tc.want.err, actualErr)
                                              ^
pkg/binary/envoy/debug/pii_test.go:94:25: Using the variable on range scope `tc` in function literal (scopelint)
                        logsDiff := cmp.Diff(tc.want.logs, actualLogs)
                                             ^
pkg/binary/envoy/debug/pii_test.go:91:5: commentFormatting: put a space between `//` and comment text (gocritic)
                                //tc.want.logs[i] = "'" + log + "'" // adding single quote to the test case
                                ^
pkg/binary/envoy/debug/pii.go:60: G104: Errors unhandled. (gosec)
        h.Write([]byte(s)) //nolint: errcheck
pkg/binary/envoy/debug/pii_test.go:25: line is 544 characters (lll)
        var FORMAT_STRING string = "[%START_TIME%] \"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\" %RESPONSE_CODE% %RESPONSE_FLAGS% \"%DYNAMIC_METADATA(istio.mixer:status)%\" \"%UPSTREAM_TRANSP
ORT_FAILURE_REASON%\" %BYTES_RECEIVED% %BYTES_SENT% %DURATION% %RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)% \"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" \"%REQ(X-REQUEST-ID)%\" \"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HO
ST%\" %UPSTREAM_CLUSTER% %UPSTREAM_LOCAL_ADDRESS% %DOWNSTREAM_LOCAL_ADDRESS% %DOWNSTREAM_REMOTE_ADDRESS% %REQUESTED_SERVER_NAME%\n"
pkg/binary/envoy/debug/pii_test.go:32: line is 442 characters (lll)
        var VALID_LOG string = `[2019-09-05T17:53:36.908Z] "GET /productpage HTTP/1.1" 200 - "-" "-" 0 5183 47 46 "10.28.2.1" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.
3809.132 Safari/537.36" "4273c09c-2534-962a-b7da-04a04cf39e88" "35.199.144.198" "127.0.0.1:9080" inbound|9080|http|productpage.default.svc.cluster.local - 10.28.2.27:9080 10.28.2.1:0 outbound_.9080_._.productpa
ge.default.svc.cluster.local`
pkg/binary/envoy/debug/pii.go:24:1: comment on exported function `ProcessLogs` should be of the form `ProcessLogs ...` (golint)
// process the array of logs by filtering it using format and containsPII, Hash the resulting array of logs and return the final array.
^
pkg/binary/envoy/debug/pii.go:57:1: comment on exported function `Hash` should be of the form `Hash ...` (golint)
// TODO: salt the Hash
^
pkg/binary/envoy/debug/pii_test.go:33:6: don't use ALL_CAPS in Go names; use CamelCase (golint)
        var LOG_WITH_ONE_FIELD string = `[2019-09-05T17:53:36.908Z]`
            ^
golangci-lint failed.
One or more linters failed.
```

pkg/binary/envoy/debug/pii_test.go
ZackButcher: Yes, this is roughly the correct way to approach this test with the code as-written. However, this is _exactly_ the kind of test you want to avoid, as the test logic is as complex as (in fact, nearly identical to) the code under test.

I planned on us running into exactly this problem 🙂. There are two major things we'll end up changing with this code before we merge it, and this is one of them. Ability to test is the main reason I mentioned abstracting out `hash` in earlier comments. The answer here is that we should extract out the function that `ProcessLogs` calls on each piece of PII and pass that function in as a parameter to `ProcessLogs` (i.e. `actual := ProcessLogs(logs, format, containsPII)` to `actual := ProcessLogs(logs, format, containsPII, hash)`; remember that go has first class functions).

Then in our tests, we can substitute in some easy function, for example `function(piiData string) string { return "pii" }` or something similar. We can easily take a log line for a test case and substitute `pii` for the appropriate piece by hand.

Then your test should be as simple as:
```go
piiFunc := func(string) string { return "pii" }

for name, c := range tests {
	t.Run(name, func(t *testing.T) {
		actual, err := ProcessLogs(c.in, c.format, c.pii, piiFunc)
		if err != nil {
			// got an error, see if it was expected and if so if it matches
			if c.err == "" {
				t.Fatalf("ProcessLogs(%v, %v, %v, piiFunc) = %v, %v, expected no error", c.in, c.format, c.pii, actual, err)
			} else if !strings.Contains(err.Error(), c.err) {
				t.Fatalf("ProcessLogs(%v, %v, %v, piiFunc) = %v, %v, but expected err: %v", c.in, c.format, c.pii, actual, err, c.err)
			}
		} else if case.err != "" {
			// expected an error but didn't get one
			t.Fatalf("ProcessLogs(%v, %v, %v, piiFunc) = %v, %v, but expected err: %v", c.in, c.format, c.pii, actual, err, c.err)
		}
		if !reflect.DeepEqual(actual, c.out) {
			t.Fatalf("ProcessLogs(%v, %v, %v, piiFunc) = %v but expected output: %v", c.in, c.format, c.pii, actual, c.out)
		}
	}
}
```
(You can do some boolean algebra to simplify that block of `if` statements if you'd like - this is a common pattern you'll repeat in _many_ tests you write. I might even suggest saving a snippet in your IDE that creates a test skeleton with this core `for name, c := range cases` loop + the declaration of `cases := map[string]string{}{}`.)

pkg/binary/envoy/debug/pii_test.go
```diff
@@ -0,0 +1,128 @@
...
+		logs []string
+		err  error
+	}
+	// test cases wrapped up in a map where key is the test case name and value is the struct
+	// defining the inputs and expecte outcome
+	tests := map[string]struct {
+		logs        []string
+		format      string
+		containsPII map[string]bool
+		want        Out
```
ZackButcher: Don't do this, it adds a ton of overhead when you're building the test case. Instead just flatten it out (you can do something simple like put a newline in the struct definition, or a comment like `// output fields` separating the input data from output):
```go
tests := map[string]struct {
	logs   []string
	format string
	pii    map[string]bool

	outLogs []string
	err     string
}{
	"simple": {
		logs: []string{validLog, logWithOneField},
		format: format,
		pii: map[string]bool{"[%START_TIME%]": true},
		outLogs: []string{validLog},
		err: "",
	},
}
```
This is much clearer to read. Some other things to note here:
- Try to keep your field initializations consistent. Don't mix with one field on one line and two on another, etc. Either initialize the entire thing on one line (`logs: []string{validLog, logWithOneField},`) or have exactly one field per line (the entire `"simple"` object above).
- The closing bracket always goes on a new line
- Always leave a trailing comma when you're initializing a struct (that last `err: nil,`). You don't have to with a list literal (`[]string{validLog, logWithOneField}` is correct, avoid `[]string{validLog, logWithOneField,}`)
- We can simplify the name `containsPII` - we know in context what it is / how it's used, so we don't need a long name. Just calling it `pii` saves us quite a bit of visual space and doesn't lose us any understanding.
- You almost never want an `error` in your test case struct, as building a correct error is a real pain. Instead, you should use a string and a `string.Contains(case.error, actualError.Error())` to compare. Then you can just ensure that the text for the expected exit condition is being hit without having to build an entire `error` object. Everywhere that you have `case.err != nil` just becomes `case.err != ""`

pkg/binary/envoy/debug/pii_test.go
```diff
@@ -0,0 +1,128 @@
...
+	logWithOneField := `[2019-09-05T17:53:36.908Z]`
+
+	type Out struct {
+		logs []string
+		err  error
+	}
+	// test cases wrapped up in a map where key is the test case name and value is the struct
+	// defining the inputs and expecte outcome
+	tests := map[string]struct {
+		logs        []string
```
ZackButcher: nit: I'd call this `in` to be most idiomatic.

pkg/binary/envoy/debug/pii_test.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+	}
+
+	// defining subtest with t.Run
+	for name, tc := range tests {
+		logs := tc.logs
+		format := tc.format
+		containsPII := tc.containsPII
+		expectedLogs := tc.want.logs
+		expectedErr := tc.want.err
```
ZackButcher: Don't do this, just use the fields directly. This is why you gave the fields descriptive names. Instead you should rename `tc` to something more useful. Either `case`, so the text reads naturally (`actual, err := ProcessLogs(case.in, case.format, case.pii)`) or `c` by convention (mirroring `t`).

pkg/binary/envoy/debug/pii_test.go
```diff
@@ -0,0 +1,128 @@
...
+						if err != nil {
+							t.Fatalf("error in hashing expected field")
+						}
+						fields[j] = h
+					}
+				}
+				expectedLogs[i] = shell.Join(fields)
+			}
+			errorDiff := cmp.Diff(expectedErr, actualErr)
+			logsDiff := cmp.Diff(expectedLogs, actualLogs)
```
ZackButcher: For now use `reflect.DeepEqual`, this does not work as you might expect. In particular, it does not compare private fields of structs. This won't trip you up in this case, but can definitely cause problems down the line (it works for Google because all the structs they're dealing with are protobufs where all of the fields that matter are public; that's not necessarily true for us in our tests). But see my suggested rewrite to avoid this part.

pkg/binary/envoy/debug/pii_test.go
```diff
@@ -0,0 +1,128 @@
...
+		" \"%UPSTREAM_TRANSPORT_FAILURE_REASON%\" %BYTES_RECEIVED% %BYTES_SENT% %DURATION% %RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)%" +
+		" \"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" \"%REQ(X-REQUEST-ID)%\" \"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HOST%\"" +
+		" %UPSTREAM_CLUSTER% %UPSTREAM_LOCAL_ADDRESS% %DOWNSTREAM_LOCAL_ADDRESS% %DOWNSTREAM_REMOTE_ADDRESS% %REQUESTED_SERVER_NAME%\n"
+
+	formats, ok := shell.Split(format)
+	if !ok {
+		t.Fatalf("error in splitting the format: %s", format)
+	}
+
+	validLog := "[2019-09-05T17:53:36.908Z] \"GET /productpage HTTP/1.1\" 200" +
```
ZackButcher: You can use string literals with ` to avoid having to double-escape the strings here:
```go
validLog := `[2019-09-05T17:53:36.908Z] "GET /productpage HTTP/1.1" 200` +
...
```

ZackButcher: And note that max string length in golang is 110 characters (enforced by our linters), so these strings can be a decent bit longer before you wrap.

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,63 @@
...
+	fieldNames, ok := shell.Split(format)
+	if !ok {
+		return []string{}, fmt.Errorf("error in splitting format string: %s", format)
+	}
+
+	out := make([]string, 0, len(logs))
+	for _, log := range logs {
+		fieldValues, ok := shell.Split(log)
+		if !ok {
+			return []string{}, fmt.Errorf("error in splitting log: %s", log)
```
ZackButcher: Here we probably don't want to bail on processing everything. Instead we probably just want to skip this one log line and carry on with the others (so just log and `continue` rather than `return`).

pkg/binary/envoy/debug/pii_test.go
```diff
@@ -0,0 +1,132 @@
...
+					// expected no error but get error
+					t.Fatalf("ProcessLogs(%v, %v, %v, piiFunc) = %v, %v, expected no error", tc.in, tc.format, tc.pii, actualLogs, actualErr)
+				} else if !strings.Contains(actualErr.Error(), tc.outErr) {
+					// actual error does not contain the expected error
+					t.Fatalf("ProcessLogs(%v, %v, %v, piiFunc) = %v, %v, but expected error: %v", tc.in, tc.format, tc.pii, actualLogs, actualErr, tc.outErr)
+				}
+			} else if tc.outErr != "" {
+				// expect error but no acutal error
+				t.Fatalf("ProcessLogs(%v, %v, %v, piiFunc) = %v, %v, but expected error: %v", tc.in, tc.format, tc.pii, actualLogs, actualErr, tc.outErr)
+			} else if !reflect.DeepEqual(actualLogs, tc.outLogs) {
```
ZackButcher: note: one option here is to drop the `else` and just keep this a bare `if !reflect.DeepEqual(actualLogs, ...)` - this means you always check the output even when err != nil, but it can be a decent practice (e.g. to ensure you always return an empty list when you return an error - as written we assert nothing about the returned value when the error is not nil).

pkg/binary/envoy/debug/pii_test.go
```diff
@@ -0,0 +1,132 @@
...
+		},
+	}
+
+	piiFunc := func(string) string { return "pii" }
+
+	// defining subtest with t.Run
+	//nolint:gocritic,scopelint
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			actualLogs, actualErr := ProcessLogs(tc.in, tc.format, tc.pii, piiFunc)
```
ZackButcher: `actual, err` - we know what's under test - the logs - and we know what `err` means - the error in this context. We don't need the extra context of `actual` in `actualErr` nor do we need the `Logs` suffix in `actualLogs` - both are still obvious in meaning.

ZackButcher:
Otherwise the code looks great, this last piece is a stylistic one that keeps things testable but hides a bunch of the details from a typical user just trying to consume the interface.

ZackButcher:
Ah, sorry - I lied a little bit: there's one other change we'll make in addition to introducing a `struct` to hang our state off of. We'll want to introduce a default, package level `Process` method. So for users there's two APIs: `pii.NewFilter().Process(logs)` and `pii.Process(logs)`. That should use the default PII filter and hash function, as well as the default Istio Envoy format string; 99% of the time that's all we'll need and the user's API is dead simple and can't be misused.

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,90 @@
...
+			}
+			out = append(out, shell.Join(fieldValues))
+		}
+	}
+	return out
+}
+
+// ProcessLogs process logs with the default filter an empty array of logs
+// and an error instance is returned in an event of error
+func ProcessLogs(logs []string) ([]string, error) {
```
taiki45: nit: I think the word "process" is a bit general for the actual task this function does. Instead, filter/mask words popped into my head (not strongly recommended though).

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Filter filters log fields using pii and modify all PII fields using f
+type Filter struct {
+	f      func(string) string
+	pii    map[string]bool
+	format []string
+}
+
+// NewFilter constructs a custom filter object
+func NewFilter(formatStr string, f func(string) string, pii map[string]bool) (Filter, error) {
```
taiki45: naming: it seems `f` lacks minimum information, so `hashFunc` or something might be better?

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// process logs with f the filter, assumes that filter has valid fields
+func (filter Filter) process(logs []string) []string {
+	out := make([]string, 0, len(logs))
+	for _, log := range logs {
+		fieldValues, ok := shell.Split(log)
+		if !ok {
+			fmt.Printf("error in splitting log: %s", log)
+			continue
```
taiki45: IMO: In this PII filtering area, we should filter/mask all unknown/invalid-format values rather than leaving them unprocessed to avoid accidental PII logging.

.gitignore
```diff
@@ -1,3 +1,4 @@
 vendor
 dist
 getenvoy
+.idea
```
taiki45: nit: this should be ignored in user/developer's local gitignore file like `~/.gitignore`.

taiki45:
@jhengy Just fyi, we should avoid force-pushing in this project? If it's on purpose, ignore my comment.

jhengy:
> @jhengy Just fyi, we should avoid force-pushing in this project? If it's on purpose, ignore my comment.

Yes, I did it on purpose after I did a `git rebase` on my local branch. The normal `git push` does not work after I rebase my local branch to master in an attempt to solve the branch out of date error on this PR

pkg/binary/envoy/debug/pii.go
ZackButcher: We omit the log entirely by calling `continue` here - so if we don't know we won't just hash the entire log line, we'll ignore it entirely and it won't appear in output.

Agree we _probably_ shouldn't `fmt.Printf` the log line itself (since it can contain PII)

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,90 @@
...
+
+var defaultFormat = `[%START_TIME%] "%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%"` +
+	` %RESPONSE_CODE% %RESPONSE_FLAGS% "%DYNAMIC_METADATA(istio.mixer:status)%" "%REQ(USER-AGENT)%"`
+var defaultPII = map[string]bool{
+	"[%START_TIME%]": true,
+}
+
+// Filter filters log fields using pii and modify all PII fields using hash
+type Filter struct {
+	hash   func(string) string
```
ZackButcher: Call this `f` - it's not always going to be a hash function. Calling it `hash` is misleading.

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,90 @@
...
+	// splitting formats and handle error
+	format, ok := shell.Split(formatStr)
+	if !ok {
+		return Filter{}, fmt.Errorf("error in splitting format string: %s", format)
+	}
+	return Filter{hash: hash, pii: pii, format: format}, nil
+}
+
+// process logs with hash the filter, assumes that filter has valid fields
+func (filter Filter) process(logs []string) []string {
```
ZackButcher:
```suggestion
func (filter Filter) Process(logs []string) []string {
```

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,90 @@
...
+		if !ok {
+			fmt.Printf("error in splitting log: %s", log)
+			continue
+		}
+
+		if len(fieldValues) == len(filter.format) {
+			// pick the PII fields and Hash the fields
+			for j, name := range filter.format {
+				if filter.pii[name] {
+					hash := filter.hash(fieldValues[j])
```
ZackButcher: don't bother declaring `hash`, just `fieldValues[j] = filter.hash(fieldValues[j])`

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,90 @@
...
+			out = append(out, shell.Join(fieldValues))
+		}
+	}
+	return out
+}
+
+// ProcessLogs process logs with the default filter an empty array of logs
+// and an error instance is returned in an event of error
+func ProcessLogs(logs []string) ([]string, error) {
+	filter, err := NewFilter(defaultFormat, defaultHash, defaultPII)
```
ZackButcher: Decent first stab at this, but there's no reason for us to instantiate a Filter every time someone calls this method. Instead you should declare `var defaultFilter, _ = NewFilter(...)` as a package level variable. Then this (which you should call `Process` to match the method name on `Filter` exactly) should just `return defaultFilter.Process(logs)`.

(We know ignoring that error at initialization time calling `NewFilter` is safe because we're handing the constructor our default values, which we know parse correctly and will never return an error.)

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,90 @@
...
+	"crypto/sha256"
+	"fmt"
+
+	"bitbucket.org/creachadair/shell"
+)
+
+// defaultHash returns the hashed value of s using sha256 defaultHash function
+// TODO: salt the Hash
+//nolint: unused,deadcode
+func defaultHash(s string) string {
```
ZackButcher: move to the end of the file, this is the least important thing in it

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,90 @@
...
+	h := sha256.New()
+	_, _ = h.Write([]byte(s))
+	return string(h.Sum(nil))
+}
+
+var defaultFormat = `[%START_TIME%] "%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%"` +
+	` %RESPONSE_CODE% %RESPONSE_FLAGS% "%DYNAMIC_METADATA(istio.mixer:status)%" "%REQ(USER-AGENT)%"`
+var defaultPII = map[string]bool{
+	"[%START_TIME%]": true,
```
ZackButcher: That was just an example; start time is not PII.

[Go ahead and put the entire list of Envoy directives here and we can go over them and mark true or false.](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log#command-operators)

pkg/binary/envoy/debug/pii.go
ZackButcher: The last thing we'll do in this PR is move this file into it's own package `pii`, so in the CLI it'd be called `pii.Process` which I think makes sense (wouldn't mind renaming to `filter` or `mask` though).

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Filter filters log fields using pii and modify all PII fields using hash
+type Filter struct {
+	hash   func(string) string
+	pii    map[string]bool
+	format []string
+}
+
+// NewFilter constructs a custom filter object
+func NewFilter(formatStr string, hash func(string) string, pii map[string]bool) (Filter, error) {
```
ZackButcher: You haven't had to deal with go packages yet so you haven't been trained to think in terms of visibility, but at the moment this interface is not usable as you intend. Primarily, there's no way for a user to access `defaultFormat`, `defaultHash`, or `defaultPII` outside of this package, so no user calling it would be able to instantiate a default Filter.

Further, the only times we don't use the default formatStr, hash, and PII are in tests. Our tests are in the same package so they can just construct a Filter directly themselves with the correct fields. The default.

So two options: 1) for now make this constructor take no arguments and return a new Filter with the default values - callers can't customize at all (less extreme version: use default hash and PII, but accept the format string from the caller; make the default format string a public variable of the package so callers can use it). 2) Add a second constructor:

```go
func New() (Filter, error) {
	return NewFilter(defaultFormat, defaultHash, defaultPII)
}
```

Either way we need to give people outside of this package a way to get an instance that has the default values.

ZackButcher:
Oh, and one other thing: let's go ahead and introduce our `pii` package. Since this is private to our tool, it should go in the `pkg` directory (this is a common go convention: stuff for others to consume goes in a top level package, stuff that's library code for this codebase is tucked into `pkg`). So we should have a new directory `getenvoy/pkg/pii`.

pkg/binary/envoy/debug/pii.go
ZackButcher: Missed on the first pass: receivers in golang are _always_ single letter, typically the first letter of the type. In this case it should be `(f Field)`

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,90 @@
...
+// defaultHash returns the hashed value of s using sha256 defaultHash function
+// TODO: salt the Hash
+//nolint: unused,deadcode
+func defaultHash(s string) string {
+	h := sha256.New()
+	_, _ = h.Write([]byte(s))
+	return string(h.Sum(nil))
+}
+
+var defaultFormat = `[%START_TIME%] "%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%"` +
```
ZackButcher: Let's call this `istioFormat`

pkg/binary/envoy/debug/pii.go
```diff
@@ -0,0 +1,90 @@
...
+	return Filter{hash: hash, pii: pii, format: format}, nil
+}
+
+// process logs with hash the filter, assumes that filter has valid fields
+func (filter Filter) process(logs []string) []string {
+	out := make([]string, 0, len(logs))
+	for _, log := range logs {
+		fieldValues, ok := shell.Split(log)
+		if !ok {
+			fmt.Printf("error in splitting log: %s", log)
```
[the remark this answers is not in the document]
ZackButcher: Ah, good time to bring up two other points:
1) If we're handling PII, we shouldn't print it to the terminal. So in this case we should just print we're skipping as `infof` or if we're in debug mode the full log `debugf`:
```go
fieldValues, ok := shell.Split(log)
if !ok {
	if log.DebugEnabled() {
		log.Debuf("error splitting log, skipping: %s", log)
	} else {
		log.Info("error splitting log, skipping")
	}
	continue
}
```

2) When we're handling strings to print them for consumption and we're using `fmt.Printf`, we should prefer `%q` to `%s`. `%q` prints the bytes passed as a string, printing them in quotes and escaping quotes in the string. `%s` just prints the bytes as-is. See the differences here: https://play.golang.org/p/xtbVjA6zA3n . Since this is logging, however, we should stick with `%s`. For example, think how frustrating it would be to try and debug the tool if it printed out `error splitting log: "asdf \"asdf\""` and you put that string into the tool again and it worked, because of the escapes `%q` added, but breaks with the _real_ string: `"asdf "asdf""`. For debug data, we want closer to the golang structs we're reasoning about when we write the code - you could even make the strong argument for using `%v` in this logging context.

.gitignore
```diff
@@ -1,3 +1,4 @@
 vendor
 dist
 getenvoy
+
```
taiki45: nit: accidental empty line

pkg/pii/pii.go
```diff
@@ -0,0 +1,136 @@
...
+	"fmt"
+
+	"bitbucket.org/creachadair/shell"
+
+	l "github.com/tetratelabs/log"
+)
+
+var logger = l.RegisterScope("pkg/pii", "filters log lines for PII data", 0)
+
+var defaultFilter, _ = Default()
```
ZackButcher: nit: make this a `var ( ... )` declaration:
```go
var (
	defaultFilter, _ = Default()

	istioFormat = `[%START_TIME%] "%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%"` +
		` %RESPONSE_CODE% %RESPONSE_FLAGS% "%DYNAMIC_METADATA(istio.mixer:status)%" "%REQ(USER-AGENT)%"`

	defaultPII = map[string]bool{ ... }
)
```

ZackButcher: (But leave the logger declaration out - use the block to group related variables, essentially)

pkg/pii/pii.go
```diff
@@ -0,0 +1,136 @@
...
+var istioFormat = `[%START_TIME%] "%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%"` +
+	` %RESPONSE_CODE% %RESPONSE_FLAGS% "%DYNAMIC_METADATA(istio.mixer:status)%" "%REQ(USER-AGENT)%"`
+var defaultPII = map[string]bool{
+	// keys without variables
+	"[%START_TIME%]":                           true,
```
ZackButcher: nit: throughout, let's use tickmark strings to make clear we're looking for these literal values and that we're not trying to do stuff like format codes with the `%` in the strings

```go
var defaultPII = map[string]bool{
	// keys without variables
	`[%START_TIME%]`:                           true,
	`[%BYTES_RECEIVED%]`:                       true,
	// ...
}
```

(tick marks meaning string literal in golang, whereas `"` can be any type of string)

pkg/pii/pii.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+			out = append(out, shell.Join(fieldValues))
+		}
+	}
+	return out
+}
+
+// ProcessLogs process logs with the default filter
+// an empty array of logs and an error instance is returned in the event of an error
+func ProcessLogs(logs []string) ([]string, error) {
```
ZackButcher: Rename to `Process` to exactly match the method name on `filter`

pkg/pii/pii.go
```diff
@@ -0,0 +1,136 @@
...
+	return defaultFilter.Process(logs), nil
+}
+
+// defaultHash returns the hashed value of s using sha256 defaultHash function
+// TODO: salt the Hash
+//nolint: unused,deadcode
```
ZackButcher: This should not be needed, right? (general rule is that we want to avoid directives in source code as much as possible as they rapidly go out of date)

PR #68, opened 2019-09-27
Issue #47: Collect files opened by envoy pre-termination (lsof)

### LINKED ISSUE #47: Collect files opened by envoy pre-termination (lsof)
This should be a [non-node level](https://github.com/tetratelabs/getenvoy/blob/master/pkg/binary/envoy/debug/) collection. Output should look similar to (or the same as) `lsof`.
The aim is to keep external dependencies to a minimum, so [this library](https://godoc.org/github.com/shirou/gopsutil/process) should be used.

```sh
root@103a9553e629:/# lsof
COMMAND PID USER   FD   TYPE DEVICE SIZE/OFF    NODE NAME
bash      1 root  cwd    DIR  0,168     4096 2153483 /
bash      1 root  rtd    DIR  0,168     4096 2153483 /
bash      1 root  txt    REG    8,1  1037528 2915087 /bin/bash
bash      1 root  mem    REG    8,1    47600  146118 /lib/x86_64-linux-gnu/libnss_files-2.23.so
bash      1 root  mem    REG    8,1    47648  146122 /lib/x86_64-linux-gnu/libnss_nis-2.23.so
bash      1 root  mem    REG    8,1    93128  146112 /lib/x86_64-linux-gnu/libnsl-2.23.so
bash      1 root  mem    REG    8,1    35688  146114 /lib/x86_64-linux-gnu/libnss_compat-2.23.so
bash      1 root  mem    REG    8,1  1868984  146069 /lib/x86_64-linux-gnu/libc-2.23.so
bash      1 root  mem    REG    8,1    14608  146082 /lib/x86_64-linux-gnu/libdl-2.23.so
bash      1 root  mem    REG    8,1   167240  146158 /lib/x86_64-linux-gnu/libtinfo.so.5.9
bash      1 root  mem    REG    8,1   162632  146049 /lib/x86_64-linux-gnu/ld-2.23.so
bash      1 root    0u   CHR  136,0      0t0       3 /dev/pts/0
bash      1 root    1u   CHR  136,0      0t0       3 /dev/pts/0
bash      1 root    2u   CHR  136,0      0t0       3 /dev/pts/0
bash      1 root  255u   CHR  136,0      0t0       3 /dev/pts/0
```

jhengy:
# Task accomplished
- fixes #47 
- example file written:
```json
[
 {
    "command": "envoy",
    "device": "",
    "fd": "0",
    "name": "/dev/pts/3",
    "node": "6",
    "pid": "7218",
    "size": "0",
    "type": "",
    "user": "hengyuan"
  },
  {
    "command": "envoy",
    "device": "",
    "fd": "1",
    "name": "/dev/pts/3",
    "node": "6",
    "pid": "7218",
    "size": "0",
    "type": "",
    "user": "hengyuan"
  },
  {
    "command": "envoy",
    "device": "",
    "fd": "2",
    "name": "/dev/pts/3",
    "node": "6",
    "pid": "7218",
    "size": "0",
    "type": "",
    "user": "hengyuan"
  },
  {
    "command": "envoy",
    "device": "",
    "fd": "3",
    "name": "anon_inode:[eventpoll]",
    "node": "",
    "pid": "7218",
    "size": "",
    "type": "",
    "user": "hengyuan"
  },
  {
    "command": "envoy",
    "device": "",
    "fd": "4",
    "name": "anon_inode:[eventfd]",
    "node": "",
    "pid": "7218",
    "size": "",
    "type": "",
    "user": "hengyuan"
  },
...
]
```

Signed-off-by: Hengyuan <jhy8890948@gmail.com>

### DIFF
--- a/pkg/binary/envoy/debug/lsof.go
+++ b/pkg/binary/envoy/debug/lsof.go
@@ -63,8 +63,8 @@ func retrieveOpenFilesData(r binary.Runner) error { //nolint:gocyclo
 
 	// filter Process instances of envoy
 	isEnvoy := func(p *process.Process) bool {
-		name, err := p.Name() //nolint:govet
-		if err != nil {
+		name, errName := p.Name()
+		if errName != nil {
 			log.Errorf("error in getting process name for %v", p)
 			return false
 		}
@@ -91,8 +91,8 @@ func retrieveOpenFilesData(r binary.Runner) error { //nolint:gocyclo
 		name, _ := envoy.Name()
 		pid := envoy.Pid
 
-		openFiles, err := envoy.OpenFiles() //nolint:govet
-		if err != nil {
+		openFiles, errOpen := envoy.OpenFiles()
+		if errOpen != nil {
 			log.Debugf("error in getting ofStat for %v\n", envoy)
 			continue
 		}
--- a/pkg/binary/envoy/debug/lsof.go
+++ b/pkg/binary/envoy/debug/lsof.go
@@ -55,7 +55,7 @@ func EnableOpenFilesDataCollection(r *envoy.Runtime) {
 
 // retrieveOpenFilesData writes statistics of open files associated with envoy instance(s) to a json file
 // if succeeded, return nil, else return an error instance
-func retrieveOpenFilesData(r binary.Runner) error { //nolint:gocyclo
+func retrieveOpenFilesData(r binary.Runner) error {
 	// get pid of envoy instance
 	pid, err := r.GetPid()
 	if err != nil {
--- a/pkg/binary/envoy/debug/lsof.go
+++ b/pkg/binary/envoy/debug/lsof.go
@@ -27,8 +27,8 @@ import (
 	"github.com/tetratelabs/log"
 )
 
-// Lsof defines the structure of statistics about a single opened file
-type Lsof struct {
+// OpenFileStat defines the structure of statistics about a single opened file
+type OpenFileStat struct {
 	// string enclosed in `` are known as struct tags
 	// this particular tag is used by json.Marshal() to encode Command field in a specific manner
 	// process dependent information
@@ -83,7 +83,7 @@ func retrieveOpenFilesData(r binary.Runner) error { //nolint:gocyclo
 	}
 	defer f.Close() //nolint
 
-	ofStatArr := make([]Lsof, 0)
+	result := make([]OpenFileStat, 0)
 	// print open file stats for all envoy instances
 	for _, envoy := range envoys {
 		// relevant fields of the process
@@ -98,7 +98,7 @@ func retrieveOpenFilesData(r binary.Runner) error { //nolint:gocyclo
 		}
 
 		for _, stat := range openFiles {
-			ofStat := Lsof{
+			ofStat := OpenFileStat{
 				Command: name,
 				Pid:     fmt.Sprint(pid),
 				User:    username,
@@ -109,21 +109,21 @@ func retrieveOpenFilesData(r binary.Runner) error { //nolint:gocyclo
 			var fstat syscall.Stat_t
 			if err := syscall.Stat(stat.Path, &fstat); err != nil {
 				// continue if the path is invalid
-				ofStatArr = append(ofStatArr, ofStat)
+				result = append(result, ofStat)
 				continue
 			}
 
 			ofStat.Node = fmt.Sprint(fstat.Ino)
 			ofStat.Size = fmt.Sprint(fstat.Size)
-			ofStatArr = append(ofStatArr, ofStat)
+			result = append(result, ofStat)
 		}
 	}
 
-	out, err := json.Marshal(ofStatArr)
+	out, err := json.Marshal(result)
 	if err != nil {
 		return fmt.Errorf("unable to convert to json representation: %v", err)
 	}
-	// write to file
+
 	fmt.Fprintln(f, string(out))
 
 	return nil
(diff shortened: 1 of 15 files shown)


jhengy:
I have yet to find a library that returns the `device` and `type` fields of the opened files. Does anybody know of any library that can do that?

pkg/binary/envoy/debug/lsof.go
```diff
@@ -0,0 +1,130 @@
...
+func retrieveOpenFilesData(r binary.Runner) error { //nolint:gocyclo
+	// get a list of processes
+	processes, err := process.Processes()
+	if err != nil {
+		return fmt.Errorf("error in getting process pids")
+	}
+
+	// filter Process instances of envoy
+	isEnvoy := func(p *process.Process) bool {
+		name, err := p.Name() //nolint:govet
```
liamawhite: whats the complaint here?

pkg/binary/envoy/debug/lsof.go
```diff
@@ -0,0 +1,130 @@
...
+		if err != nil {
+			log.Errorf("error in getting process name for %v", p)
+			return false
+		}
+		return name == "envoy"
+	}
+	envoys := make([]*process.Process, 0)
```
liamawhite: We should be able to get the pid from the command, can we just use that rather than search? It will be much cheaper on large VMs with thousands of processes.

pkg/binary/envoy/debug/lsof.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+	}
+
+	f, err := os.Create(filepath.Join(r.DebugStore(), "lsof/lsof.json"))
+	if err != nil {
+		return fmt.Errorf("unable to create file to write lisof output to: %v", err)
+	}
+	defer f.Close() //nolint
+
+	ofStatArr := make([]Lsof, 0)
```
liamawhite: try to avoid types in variable names

pkg/binary/envoy/debug/lsof.go
```diff
@@ -0,0 +1,130 @@
...
+	// print open file stats for all envoy instances
+	for _, envoy := range envoys {
```
liamawhite: We should only check one Envoy instance, this is a single Envoy debug level, not node.

pkg/binary/envoy/debug/lsof.go
liamawhite: also please try to avoid acronyms, makes it more difficult to parse code. Something like `openFileStats` should work here.

pkg/binary/envoy/debug/lsof.go
```diff
@@ -0,0 +1,130 @@
...
+		// relevant fields of the process
+		username, _ := envoy.Username()
+		name, _ := envoy.Name()
+		pid := envoy.Pid
+
+		openFiles, err := envoy.OpenFiles() //nolint:govet
```
liamawhite: Whats the complaint here? In general please add a justification for avoiding a linting error in comments. Defer error checking is the exception.

pkg/binary/envoy/debug/lsof.go
```diff
@@ -0,0 +1,130 @@
...
+			ofStat.Size = fmt.Sprint(fstat.Size)
+			ofStatArr = append(ofStatArr, ofStat)
+		}
+	}
+
+	out, err := json.Marshal(ofStatArr)
+	if err != nil {
+		return fmt.Errorf("unable to convert to json representation: %v", err)
+	}
+	// write to file
```
liamawhite: Comments should generally only explain why not what, Go is verbose by design to avoid what comments.

pkg/binary/envoy/debug/lsof.go
jhengy: @liamawhite The error is ` shadow: declaration of "err" shadows declaration at line 59`. I think the error is returned because I reused the identifier `err` in another function scope. Would it be better if I just change the variable name instead of ignoring the lint?

pkg/binary/envoy/debug/lsof.go
jhengy: This line gives the same error: `shadow: declaration of "err" shadows declaration at line 59`

pkg/binary/envoy/debug/lsof.go
liamawhite: yes, avoid shadowing

pkg/binary/envoy/debug/lsof.go
jhengy: @liamawhite I tried to find a function that returns me the pid of a process from its name using [this](https://godoc.org/github.com/shirou/gopsutil/process). However, it seems like the recommended library does not have such function. I tried googling for other libraries and found [this](https://godoc.org/k8s.io/kubernetes/pkg/util/procfs#PidOf).

jhengy: @liamawhite I have found out from Lizan that I can find the pid of the envoy instance from binary.Runner, will implement it and push changes

pkg/binary/envoy/debug/lsof.go
```diff
@@ -55,75 +56,59 @@ func EnableOpenFilesDataCollection(r *envoy.Runtime) {
 // retrieveOpenFilesData writes statistics of open files associated with envoy instance(s) to a json file
 // if succeeded, return nil, else return an error instance
 func retrieveOpenFilesData(r binary.Runner) error { //nolint:gocyclo
```
liamawhite: Please split the function into smaller ones rather than turn off linting.

jhengy: Alright, the lint now passes I remove the for loops in previous commits

PR #107, opened 2020-11-04
wasm: packing built binary into OCI-compliant images

musaprg:
This PR introduces the `getenvoy extension push` command to push a WASM extension to the OCI-compliant registry.

## TODO
- [x] PoC implementation
- [x] add unittest for the new lines
- [x] add e2e test


docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: '3'
+services:
+  registry:
+    image: registry:2
+    network_mode: host
+  localstack:
+    image: localstack/localstack
```
mathetake, line 7: looks like localstack's ECR support is not supported in the OSS version, isn't it? https://github.com/localstack/localstack#overview

pkg/cmd/extension/push/cmd.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// cmdOpts represents configuration options of the `push` command.
+type cmdOpts struct {
+	// Toolchain to use to build the *.wasm file.
+	Toolchain common.ToolchainOpts
+	// Extension to use to specify the built *.wasm file.
+	Extension runtime.ExtensionOpts
+	// Pusher to use to specify options for Pusher
+	Pusher wasmimage.PusherOpts
```
mathetake, line 34: Can we make these fields start with being lower cased?

pkg/cmd/extension/push/cmd_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+	var dockerDir string
+
+	BeforeEach(func() {
+		dir, err := filepath.Abs("../../../extension/workspace/toolchain/builtin/testdata/toolchain")
+		Expect(err).ToNot(HaveOccurred())
+		dockerDir = dir
+	})
+
+	var pathBackup string
```
mathetake, line 50: and I'm just curious do we really need to backup?

pkg/extension/wasmimage/wasmimage.go
```diff
@@ -0,0 +1,61 @@
...
+	"errors"
+	"fmt"
+	orascnt "github.com/deislabs/oras/pkg/content"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+type WasmImage struct {
```
mathetake, line 13: lowercase?

mathetake (changes requested):
- Could you please do go imports?
- I don't think it's good to push the raw Wasm binary: `pkg/cmd/extension/push/testdata/workspace/extension.wasm `

docker-compose.yml
musaprg: oh, sorry I missed it isn't provided in the community version. I'll remove it and write unit and e2e tests with a local registry.

musaprg:
> I don't think it's good to push the raw Wasm binary: pkg/cmd/extension/push/testdata/workspace/extension.wasm

@mathetake I think it's not a bad way because it is a unit test. However, the binary is too large so I will prepare another smaller binary if possible.

lizan:
> > I don't think it's good to push the raw Wasm binary: pkg/cmd/extension/push/testdata/workspace/extension.wasm
> 
> @mathetake I think it's not a bad way because it is a unit test. However, the binary is too large so I will prepare another smaller binary if possible.

For unit test it might be OK, but 6MiB seems too large.

pkg/extension/wasmimage/wasmimage.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	orascnt "github.com/deislabs/oras/pkg/content"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+type WasmImage struct {
```
Shikugawa, line 14: You can ignore this because it is just personal comment and may have wrong consideration.
I think `WasmImage` should be stateful at the global sight. For more detail, it should have `pushed`, `pending`, `not pushed` state. So I consider that we should have `WasmImage` interface and implement `push()` method, with injecting `Pusher` dependency, and also we can have `Pusher` interface, right? If I have wrong pointing out about this, please notify it. Thanks.

musaprg: > I think WasmImage should be stateful at the global sight. For more detail, it should have pushed, pending, not pushed state.

oh, yes, that seems better in terms of their status managements. But currently, I don't expect to push multiple images at the same time, so we might not need to do it because the job works synchronously. what do you think about that?

Shikugawa: The indication signal may be used handling signals like `SIGTERM`, which will execute fallback. I'm not sure about the detail of the library which you used, but if the library doesn't have the handling mechanism, we may implement this feature. But it can't be used in this PR.

pkg/cmd/extension/push/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+				if err != nil {
+					return err
+				}
+				image, err = toolchain.Package(imageRef)
+				if err != nil {
+					return err
+				}
+			}
+			pusher, err := wasmimage.NewPusher(opts.pusher.AllowInsecure, opts.pusher.UseHTTP)
+			desc, err := pusher.Push(image)
```
mathetake, line 108: why not just passing the image's path to `wasmimage.NewPusher`? Then when we no longer need to expose neither `NewWasmImage` nor  `WasmImage` struct.

pkg/cmd/extension/push/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+				image, err = wasmimage.NewWasmImage(imageRef, opts.extension.WasmFile)
+				if err != nil {
+					return err
+				}
+			} else {
+				workspace, err := workspaces.GetCurrentWorkspace()
+				if err != nil {
+					return err
+				}
+				toolchain, err := common.LoadToolchain(workspace, opts)
```
mathetake, line 98: unused variable

pkg/cmd/extension/push/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			imageRef := args[0]
+			var image *wasmimage.WasmImage
+			if opts.extension.WasmFile != "" {
+				image, err = wasmimage.NewWasmImage(imageRef, opts.extension.WasmFile)
+				if err != nil {
+					return err
+				}
+			} else {
+				workspace, err := workspaces.GetCurrentWorkspace()
```
mathetake, line 94: ditto

pkg/cmd/extension/push/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+				if err != nil {
+					return err
+				}
+				toolchain, err := common.LoadToolchain(workspace, opts)
+				if err != nil {
+					return err
+				}
+				image, err = toolchain.Package(imageRef)
```
mathetake, line 102: ref: https://github.com/tetratelabs/getenvoy/pull/107/files#r551768609
```suggestion
				image = toolchain.GetBuildOutputWasmFile()
```

pkg/cmd/extension/push/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+
+			if len(args) == 0 {
+				return errors.New("missing image-reference parameter")
+			}
+
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			imageRef := args[0]
+			var image *wasmimage.WasmImage
```
mathetake, line 87: var imagePath string
```suggestion
			imagePath := opts.extension.WasmFile
```

ref: https://github.com/tetratelabs/getenvoy/pull/107/files#r551768609

pkg/cmd/extension/push/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+			if opts.extension.WasmFile != "" {
```
mathetake, line 88: delete this block

Ref: https://github.com/tetratelabs/getenvoy/pull/107/files#r551768609

pkg/cmd/extension/push/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+				image, err = wasmimage.NewWasmImage(imageRef, opts.extension.WasmFile)
+				if err != nil {
+					return err
+				}
+			} else {
```
mathetake, line 93:
```suggestion
			if imagePath == "" {
```
Ref: https://github.com/tetratelabs/getenvoy/pull/107/files#r551768609

pkg/extension/wasmimage/pusher.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	resolver, err := auth.Resolver(context.Background(), client, useHTTP)
+	if err != nil {
+		return nil, err
+	}
+	return &Pusher{resolver: resolver}, nil
+}
+
+// Push pushes the image to the registry
+func (p *Pusher) Push(image *WasmImage) (ocispec.Descriptor, error) {
```
mathetake, line 79:
```suggestion
func (p *Pusher) Push(image *WasmImage, imageRef, imagePath string) (ocispec.Descriptor, error) {
```

pkg/extension/wasmimage/wasmimage.go
```diff
@@ -0,0 +1,76 @@
...
+	ref      string
+	name     string
+	contents []byte
+
+	store  *orascnt.Memorystore
+	layers []ocispec.Descriptor
+}
+
+// NewWasmImage returns a new WasmImage instance
+func NewWasmImage(ref string, path string) (*WasmImage, error) {
```
mathetake, line 38:
```suggestion
func newWasmImage(ref string, path string) (*WasmImage, error) {
```

pkg/extension/wasmimage/wasmimage.go
```diff
@@ -0,0 +1,76 @@
...
+	return &WasmImage{
+		ref:      ref,
+		name:     name,
+		contents: contents,
+		layers:   layers,
+		store:    store,
+	}, nil
+}
+
+func validateFile(path string) error {
```
mathetake, line 66:
```suggestion
func validatePath(path string) error {
```

pkg/extension/wasmimage/pusher.go
```diff
@@ -0,0 +1,88 @@
...
+	ctx := context.Background()
```
mathetake, line 80:
```suggestion
	ctx := context.Background()
	image, err := newWasmImage(imageRef, imagePath)
```

pkg/cmd/extension/push/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+				toolchain, err := common.LoadToolchain(workspace, opts)
+				if err != nil {
+					return err
+				}
+				image, err = toolchain.Package(imageRef)
+				if err != nil {
+					return err
+				}
+			}
+			pusher, err := wasmimage.NewPusher(opts.pusher.AllowInsecure, opts.pusher.UseHTTP)
```
mathetake, line 107: should handle the error

pkg/cmd/extension/push/cmd.go
musaprg: sounds good. I made them private and changed the signature of `pusher.Push`

musaprg:
for now, e2e tests on macOS fail due to https://github.com/tetratelabs/getenvoy/pull/112, but I basically don't care about that in this PR.

lizan (approved):

pkg/cmd/extension/push/cmd.go
```diff
@@ -0,0 +1,132 @@
...
+			return nil
+		},
+	}
+	cmd.PersistentFlags().StringVar(&opts.toolchain.Name, "toolchain", opts.toolchain.Name,
+		`Name of the toolchain to use, e.g. "default" toolchain that is backed by a Docker build container`)
+	cmd.PersistentFlags().BoolVar(&opts.pusher.AllowInsecure, "allow-insecure", opts.pusher.AllowInsecure, `Allow insecure registry`)
+	cmd.PersistentFlags().BoolVar(&opts.pusher.UseHTTP, "use-http", opts.pusher.UseHTTP, `Use HTTP for communication with registry`)
+	cmd.PersistentFlags().StringVar(&opts.extension.WasmFile, "extension-file", opts.extension.WasmFile,
+		`Use a pre-built *.wasm file`)
+	cmd.PersistentFlags().StringVar(&opts.extension.Config.Source, "extension-config-file", opts.extension.Config.Source,
```
mathetake, line 129: looks like unused

mathetake, line 129: looks like unused

pkg/cmd/extension/push/cmd_test.go
```diff
@@ -0,0 +1,138 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package push_test
+
+import (
```
mathetake, line 17: same for the style.

pkg/cmd/extension/root.go
```diff
@@ -17,6 +17,8 @@ package extension
 import (
 	"os"
 
+	"github.com/tetratelabs/getenvoy/pkg/cmd/extension/push"
```
mathetake, line 20: same

pkg/extension/wasmimage/mediatypes.go
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package wasmimage
+
+// media types defined in WASM Artifact Image Specification
+const (
+	ConfigMediaType       = "application/vnd.module.wasm.config.v1+json"
+	ContentLayerMediaType = "application/vnd.module.wasm.content.layer.v1+wasm"
```
mathetake, line 20: no need to export

pkg/extension/wasmimage/puller.go
```diff
@@ -0,0 +1,89 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package wasmimage
+
+import (
```
mathetake, line 17: same

pkg/extension/wasmimage/mediatypes.go
mathetake, line 20: no need to export

pkg/extension/wasmimage/pusher.go
```diff
@@ -0,0 +1,98 @@
...
+	UseHTTP       bool
+}
+
+// NewPusherOpts returns a default PusherOpts instance
+func NewPusherOpts() PusherOpts {
+	return PusherOpts{
+		AllowInsecure: false,
+		UseHTTP:       false,
+	}
+}
```
mathetake, line 49: I don't think this struct is necessary. I suggest just parse the parse flags with local variables and then pass them to `NewPusher` directly.

pkg/extension/wasmimage/pusher.go
```diff
@@ -0,0 +1,98 @@
...
+	// TODO(musaprg): separate these instructions into another functions
+	auth, err := docker.NewClient()
+	if err != nil {
+		return nil, err
+	}
+	resolver, err := auth.Resolver(context.Background(), client, useHTTP)
+	if err != nil {
+		return nil, err
+	}
+	return &Pusher{resolver: resolver}, nil
```
mathetake, line 78: unnecessary if statement
```suggestion
	return  auth.Resolver(context.Background(), client, useHTTP)
```

pkg/cmd/extension/push/cmd.go
```diff
@@ -0,0 +1,132 @@
...
+
+			cmd.Printf("Pushed %s\n", ref)
+			cmd.Printf("digest: %s size: %d\n", manifest.Digest, size)
+
+			return nil
+		},
+	}
+	cmd.PersistentFlags().StringVar(&opts.toolchain.Name, "toolchain", opts.toolchain.Name,
+		`Name of the toolchain to use, e.g. "default" toolchain that is backed by a Docker build container`)
+	cmd.PersistentFlags().BoolVar(&opts.pusher.AllowInsecure, "allow-insecure", opts.pusher.AllowInsecure, `Allow insecure registry`)
```
mathetake, line 125: I think we should describe what "insecure" means here. (e.g. allow insecure tls connection)

pkg/extension/wasmimage/pusher.go
mathetake: or put them in `cmdOpts` just like others.

pkg/extension/wasmimage/wasmimage.go
```diff
@@ -0,0 +1,76 @@
...
+	if _, err := os.Stat(path); err != nil {
+		return fmt.Errorf("%#v does not exist: %w", path, err)
+	}
+
+	if ext := filepath.Ext(path); ext != ".wasm" {
```
mathetake, line 71: nit: I see this is a little bit weird since whether "Wasm binary" or not is not determined by the extension, but rather 
checking if a binary's header starts with the magic number or not.

pkg/extension/wasmimage/pusher.go
```diff
@@ -0,0 +1,98 @@
...
+// Pusher knows how to push wasm images to OCI-compliant registries.
+type Pusher struct {
+	resolver remotes.Resolver
+}
+
+// NewPusher returns a new Pusher instance.
+func NewPusher(insecure, useHTTP bool) (*Pusher, error) {
+	client := http.DefaultClient
+
+	if insecure {
```
Shikugawa, line 60: Shouldn't we determine behavior when the option which specifies `allowInsecure` as `false`? I think we shouldn't provide the knob to configure not to use TLS or not.

pkg/cmd/extension/push/cmd.go
Shikugawa, line 125: I think this option couldn't be configured by user in the current implementation because of what I described below.

pkg/extension/wasmimage/puller.go
```diff
@@ -0,0 +1,89 @@
...
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+
+	"github.com/containerd/containerd/remotes"
+	"github.com/deislabs/oras/pkg/auth/docker"
+	orascnt "github.com/deislabs/oras/pkg/content"
+	"github.com/deislabs/oras/pkg/oras"
+)
+
+var (
+	pullOpts = []oras.PullOpt{
```
Shikugawa, line 35: Couldn't it define under `Puller`?

pkg/extension/wasmimage/puller.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	return &Puller{resolver: resolver}, nil
+}
+
+// Pull fetches the specified image from the registry
+func (p *Puller) Pull(imageRef, imagePath string) (ocispec.Descriptor, error) {
+	ctx := orasctx.Background()
+	store := orascnt.NewMemoryStore()
+
+	_, layers, err := oras.Pull(ctx, p.resolver, imageRef, store, pullOpts...)
```
Shikugawa, line 74: My opinion that `oras` can handle fallback mechanism when sent signals like `SIGTERM`, or disconnected by some of the reasons. Is it correct?

pkg/extension/wasmimage/pusher.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil, err
+	}
+	return &Pusher{resolver: resolver}, nil
+}
+
+// Push pushes the image to the registry
+func (p *Pusher) Push(imagePath, imageRef string) (manifest ocispec.Descriptor, size int64, err error) {
+	ctx := orasctx.Background()
+
+	var image *wasmImage
```
Shikugawa, line 85: Not needed

pkg/extension/wasmimage/pusher.go
```diff
@@ -0,0 +1,98 @@
...
+
+	image, err = newWasmImage(imageRef, imagePath)
```
Shikugawa, line 87:
```suggestion
	image, err := newWasmImage(imageRef, imagePath)
```

pkg/extension/wasmimage/puller.go
mathetake: I guess signals are not handled by them. I don't see any value to handle singles in this cli since the state holded by registries.

Shikugawa: @mathetake got it. sounds good.

pkg/extension/wasmimage/pusher.go
musaprg: You're right, but we need to allow it as there's some option to use getenvoy with some insecure registries (e.g., self-hosted registry signed with self-signed certification). This option is basically disabled and the user must enable it explicitly when the user is expected to access the insecure registry.

go.mod
```diff
@@ -34,8 +39,15 @@ require (
...
 	istio.io/api v0.0.0-20200227213531-891bf31f3c32
 	istio.io/istio v0.0.0-20200304114959-c3c353285578
+	rsc.io/letsencrypt v0.0.3 // indirect
 )
 
+replace github.com/Azure/go-autorest/autorest => github.com/Azure/go-autorest/autorest v0.11.15
+
+replace github.com/docker/docker => github.com/docker/docker v17.12.1-ce+incompatible
+
+replace github.com/hashicorp/consul => github.com/hashicorp/consul v1.3.1
```
mathetake, line 49: just curious, what are the purpose of these replacements?

pkg/cmd/extension/push/cmd.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (opts *cmdOpts) ApplyTo(interface{}) {}
+
+func (opts *cmdOpts) Validate() error {
+	if err := opts.toolchain.Validate(); err != nil {
+		return err
+	}
+
+	return nil
```
mathetake, line 62:
```suggestion
	return opts.toolchain.Validate()
```

go.mod
musaprg: These replacements are mainly for resolving the dependency issues caused by other packages.

musaprg:
Collected given issues, could you check again? @mathetake @Shikugawa

mathetake (approved):

PR #128, opened 2021-03-23
WIP: updates to latest dependencies

codefromthecrypt:
This attempts to update to latest dependencies. There is a glitch
locally with one test.

Also, controlplane's `istio_test` `Test_IstioGateway` is untranslatable
into current apis. We need to decide whether to create fixtures to
re-implement it, or leave it out. Considering it was already with a
hazard sign, I figured worthwhile asking:

```
// NOTE: This test will fail on macOS due to an issue with Envoy, the same issue as debug logging
```


codefromthecrypt:
I intentionally raised a head branch, so that folks can edit this as necessary. I walked into a bunch of out of date deps when updating the log driver and timeboxed some effort towards the others. Any advice or commits welcome, as I'm a bit over my head at this point!

pkg/binary/envoy/controlplane/istio.go
```diff
@@ -94,7 +94,6 @@ func writeBootstrap(r binary.Runner) error {
 	}
 	if _, err := bootstrap.New(bootstrap.Config{
 		Node:           istioNode(e.Config),
-		DNSRefreshRate: "60s",
```
codefromthecrypt, line 97: deleted symbol

pkg/binary/envoy/controlplane/istio_test.go
```diff
@@ -38,43 +28,3 @@ func TestMain(m *testing.M) {
 	}
 	os.Exit(m.Run())
 }
-
-// NOTE: This test will fail on macOS due to an issue with Envoy, the same issue as debug logging
-func Test_IstioGateway(t *testing.T) {
```
codefromthecrypt, line 43: most of the code here doesn't exist anymore. Not sure if we want to do this or delete the test. Seems possibly easier in docker?

codefromthecrypt (commented):
notes!

pkg/binary/envoy/controlplane/istio_test.go
yskopets: @liamawhite Could you please advise how to upgrade correctly Istio dependencies without breaking `getenvoy run` ?

I think, this unit test is only one aspect to consider.

More importantly, we need to make sure that
```shell
# Run as a gateway using an Istio controlplane bootstrap.
getenvoy run standard:1.11.1 --mode loadbalancer --bootstrap istio --controlplaneAddress istio-pilot.istio-system:15010
```
will not get broken.

yskopets:
@codefromthecrypt The most problematic part of this update are Istio dependencies.

I don't know this part of `getenvoy`.

@liamawhite  Could you please advise how to upgrade correctly Istio dependencies without breaking `getenvoy run` ?

codefromthecrypt:
ps I intentionally didn't mark this for review, yet. dependencies will settle down when other PRs merge, making this smaller and more focused on the istio related stuff.

codefromthecrypt:
I learned more about the istio test in #160 and think I can probably port it to the new libraries. However, there's 4 outstanding PRs of mine, and most reduce deps one way or another. I'm not touching this again until they are in. Once they are in, I'll see if I can retain the test in question.

codefromthecrypt:
#168 moves up to Istio 1.6 first

PR #129, opened 2021-03-24
Changes from statik to go:embed for example and extension templates

### LINKED ISSUE #126: Can go:embed replace statik?
Right now, we are having to work around unmerged change in statik. The build is also more fragile the more generation steps are involved.

Ideally, we can use go:embed. There are a number of projects who have migrated in this way. For example, https://github.com/sascha-andres/go-logsink/pull/36

If possible, we could remove generation commands from make and CI as a result.

Continue to maintain a fork as statik isn't really merging change we need.

**Additional context**
More examples https://github.com/search?q=statik+go%3Aembed&type=issues

codefromthecrypt:
This reduces build complexity by eliminating a generation step with go:embed. This implicitly reduces tech debt even more because not only were we using a stalled project, statik, but also a fork of it.

go:embed is not perfect, as it disallows the file name `go.mod`. The workaround is to rename it to `go.mod_` per https://github.com/golang/go/issues/45197. While this is imperfect an if-statement is a far better punch than a dependency on a fork of a stalled project which also requires a codegen step.

Fixes #126


data/extension/init/templates/tinygo/envoy.access_loggers/default/go.mod
```diff
@@ -1,8 +0,0 @@
-module {{ .Extension.Name }}
```
codefromthecrypt, line 1: fyi @opennota this is the file I had to nuke per https://github.com/golang/go/issues/41191#issuecomment-805398783

Will watch the build output but guessing it will fail expecting go.mod

codefromthecrypt:
One simple workaround could be to serve the go.mod files dynamically in go. This could stay until there's a go.mod solution in go:embed, and still get us off the abandoned library and more complicated build.

Each of the go.mod files are identical.
data/extension/init/templates/tinygo/envoy.access_loggers/default/go.mod
data/extension/init/templates/tinygo/envoy.filters.http/default/go.mod
data/extension/init/templates/tinygo/envoy.filters.network/default/go.mod
```
module {{ .Extension.Name }}

go 1.15

require (
	github.com/stretchr/testify v1.6.1
	github.com/tetratelabs/proxy-wasm-go-sdk v0.1.0
)
```

codefromthecrypt:
might be saved by the bell, as I mistook the root path used in statik. it was `data/example/init` not `data/extension/init`. The latter has go, but the former doesn't.

🤞

codefromthecrypt:
OK I think the go.mod in the extensions dir seems to be about our lint setup. If all tests pass except lint, it might be a better punch to avoid that lint check vs override go's behavior or rely on codegen.

In a bit, I'll raise an issue in go to document the surprising limitation about go.mod files. I can understand how someone could equate presence of go.mod meaning the directory is no longer in the module, but it is surprising enough to have an issue to point to and ideally docs about.

codefromthecrypt:
raised https://github.com/golang/go/issues/45197 on the constraint about go.mod. Not sure if having a `go.mod` is critical in tinygo or not. It seems that it trips our lint meanwhile.

need a hand on the lint failure as it isn't consistent running locally. I tried to exclude but guess I got something wrong.

```
ERRO Running error: context loading failed: failed to load program with go/packages: -: go: downloading github.com/stretchr/testify v1.6.1
```

yskopets:
> Not sure if having a go.mod is critical in tinygo or not.

The goal of `GetEnvoy Extension Toolkit` is help a user to kick off extension development that follows all best practices (as opposed to starting from a "Hello World" and learning painfully it's not production-ready in any way).

All modern `Go` project should use Go modules. So, we have to provide `go.mod` out-of-the-box one way or another.

yskopets:
> One simple workaround could be to serve the go.mod files dynamically in go. This could stay until there's a go.mod solution in go:embed, and still get us off the abandoned library and more complicated build.

Personally, I don't see what features of `go:embed` justify the need in such workarounds.

I earlier said that "`go:embed` provides exactly the same features as `statik` but has slightly different API". Apparently, it's not the case - `go:embed` is a less feature-complete solution.

Again, I'm just one vote. If other reviewers are fine with it, I won't object.

codefromthecrypt:
The following suggestion seems alright to me, provided we document it https://github.com/golang/go/issues/45197#issuecomment-805561759 basically rename go.mod to go.mod_

We should keep in mind that the file is already templated, so it isn't valid anyway.

codefromthecrypt:
FWIW I'd like to finish #130 before merging this anyway because testing is where I lost most time

data/extension/init/templates/tinygo/envoy.access_loggers/default/go.mod
codefromthecrypt: used a workaround to keep the file. rewrote desc with the outcome

data/extension/init/templates/tinygo/.gitignore
```diff
@@ -0,0 +1,2 @@
+# We rename go.mod to go.mod_ to workaround https://github.com/golang/go/issues/45197
```
codefromthecrypt, line 1: @mathetake this will only effect tinygo as I don't expect any other extension language to want a file named go.mod in it.

pkg/extension/init/init.go
```diff
@@ -102,7 +102,12 @@ func (s *scaffolder) walk(sourceDirName, destinationDirName string) (errs error)
 }
 
 func (s *scaffolder) visit(sourceDirName, destinationDirName string, sourceFileInfo os.FileInfo) (errs error) {
-	relOutputFileName := filepath.Join(destinationDirName, sourceFileInfo.Name())
+	baseOutputFileName := sourceFileInfo.Name()
+	// We rename go.mod to go.mod_ to workaround https://github.com/golang/go/issues/45197
```
codefromthecrypt, line 106: this is far far easier than writing an FS renamer, especially considering we are already walking the file tree. However, I linked the issue as it is soliciting a library to do that.

.golangci.yml
```diff
@@ -25,6 +25,7 @@ run:
   #   	vendor$, third_party$, testdata$, examples$, Godeps$, builtin$
   skip-dirs:
     - pkg/binary/envoytest
+    - data/extension/init/templates/tinygo  # skip templated go projects
```
codefromthecrypt, line 28: I must have syntax wrong.. as the linter (in CI) is still failing like this, presumably because it is peering into these dirs

```
ERRO Running error: context loading failed: failed to load program with go/packages: -: go: downloading github.com/stretchr/testify v1.6.1
```

codefromthecrypt:
good news is the tinygo tests pass now.

Not sure about the rust failures haven't looked deeply yet
```
           /tmp/cache/getenvoy/extension/rust-builder/cargo/registry/src/github.com-1ecc6299db9ec823/proxy-wasm-experimental-0.0.8/src/hostcalls.rs:919: undefined reference to `proxy_done'
          collect2: error: ld returned 1 exit status
```

codefromthecrypt:
will rebase after https://github.com/tetratelabs/getenvoy/pull/135 and look into adding unit tests that show the templates directory tree isn't empty due to faulty go:embed setup.

codefromthecrypt:
pulled the lint thing out to here #138 and will rebase later

codefromthecrypt:
changed merge base to `start-migration` until #130 is in, or I figure out what the glitches are. debugging before that is arduous.

codefromthecrypt:
Thanks for the tip by @carlmjohnson I didn't realize that dot directories are filtered in go:embed expressions. I renamed `.cargo` to `cargo` to work around this.

Summary is two if statements which would be common to users who understand go:embed pros and cons instead of a codegen phase powered by a fork of a fork. Dirty, but less dirty.

codefromthecrypt:
ps it isn't strictly required to rename the config file to config.toml, but I noticed that the old naming isn't encouraged anymore per https://doc.rust-lang.org/cargo/reference/config.html so fixed it while here.

codefromthecrypt:
will look into this again after #130 and #151 are merged because it is a lot easier to debug things that way.

codefromthecrypt:
I think we're having docker problems. I'll re-kick this tomorrow, but I think the code change is done now. There's a test coverage issue on "getenvoy extension init" which I'll do in a separate PR, so that there's nothing needed here.

codefromthecrypt:
#161 finishes backfilling the tests needed for here

codefromthecrypt:
circleci is no longer in use, so we can ignore that

go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	github.com/tetratelabs/getenvoy-package v0.0.0-20190730071641-da31aed4333e
 	github.com/tetratelabs/log v0.0.0-20190710134534-eb04d1e84fb8
 	github.com/tetratelabs/multierror v1.1.0
+	github.com/tetratelabs/proxy-wasm-go-sdk v0.12.0
```
codefromthecrypt, line 39: one glitch of renaming go.mod -> go.mod_ is that if you do `go mod tidy`, this pops into the root go.mod only because the examples import it. I'll look into it..

RATIONALE.md
```diff
@@ -0,0 +1,66 @@
...
+
+See https://pkg.go.dev/embed#hdr-Directives for more information.
+
+### Limitations of `go:embed` impact extension init templates
+`getenvoy extension init` creates a workspace directory for a given category and programming language. Some constraints
+of `go:embed` impact how these template directories are laid out, and the workarounds are file rename in basis. The
+impacts are noted below for information and future follow-up:
+
+#### `go:embed` doesn't traverse hidden directories, but Rust projects include a hidden directory 
+Our Rust examples use [Cargo][https://doc.rust-lang.org/cargo/reference/config.html] as a build tool. This stores
```
mathetake, line 31:
```suggestion
Our Rust examples use [Cargo](https://doc.rust-lang.org/cargo/reference/config.html) as a build tool. This stores
```

.github/workflows/release.yaml
```diff
@@ -137,5 +142,8 @@ jobs:
         timeout-minutes: 10  # fail fast if MacOS runner becomes to slow
 
       - name: "Run e2e tests using released `getenvoy` binary and published extension builder images"
+        # chmod to restore permissions lost in actions/download-artifact@v2
         # expand E2E_TOOLCHAIN_CONTAINER_OPTIONS here to allow shell interpolation
-        run: E2E_TOOLCHAIN_CONTAINER_OPTIONS="${{ matrix.args.toolchain-container-options }}" make e2e
+        run: |
+          chmod a+x build/bin/*/*/getenvoy
```
codefromthecrypt, line 148: this magically popped up again, so might as well fix it here.

mathetake: why do we need this..😕  should be passed without this since the flow is exactly the same as users do when they download and use. it

codefromthecrypt: sorry you are right. I didn't need to double copy/pasta

mathetake (approved):

codefromthecrypt:
Thanks for the review @mathetake!

Issue #127, opened 2021-03-23
Test the tinygo examples


codefromthecrypt:
**Is your feature request related to a problem? Please describe.**

In trying to understand which dependencies can upgrade, I tried to run tests manually in `data/extension/init/templates/tinygo/envoy.access_loggers/default` (after making the package name valid)

```bash
$ tinygo test
can't load test package: /Users/adrian/go/pkg/mod/github.com/stretchr/testify@v1.6.1/assert/assertions.go:14:2: package runtime/debug is not in GOROOT (/Users/adrian/Library/Caches/tinygo/goroot-go1.16.2-4cc3b348dc408e179579abba3f2d226a92f1a159b77d8068cee8b4bf5438ca85-syscall/src/runtime/debug)
```

**Describe the solution you'd like**
Similar to compilation (which uses docker), it would be handy to also run the test to provide assurance the test code is valid.

**Describe alternatives you've considered**
An alternative is a README with instructions on how to run the tests

**Additional context**

After digging around, it seems this is possibly a current problem being investigated. https://github.com/tinygo-org/tinygo/issues/1634

yskopets:
@codefromthecrypt We test all extension templates as part of our `e2e` tests.

* For source code see https://github.com/tetratelabs/getenvoy/tree/master/test/e2e
* For logs from a test run see https://github.com/tetratelabs/getenvoy/runs/2138055430?check_suite_focus=true

codefromthecrypt:
Thanks for the tips on the e2e tests. I was mainly curious about the unit
tests, ex:
https://github.com/tetratelabs/getenvoy/blob/master/data/extension/init/templates/tinygo/envoy.access_loggers/default/main_test.go

possibly I'm blind, but I don't see how that is run. If it is run, can you
mention where? If it isn't we should probably run it or delete it, ya?
(same pattern for all the others)

>

codefromthecrypt:
guessing maybe `extension test` implicitly runs `go test`. If "default toolchain" might be "go", we could perhaps improve the docs to say this as right now it just says docker.. https://www.getenvoy.io/reference/getenvoy_extension_test

```
			By("running `extension test` command")
			stdout, stderr, err := GetEnvoy("extension test").
				Args(e2e.Env.GetBuiltinContainerOptions()...).
				Exec()
```

yskopets:
> I don't see how that is run. If it is run, can you mention where?

`e2e` tests reproduce development flow of an extension developer:
* ```shell
[...]

PR #130, opened 2021-03-24
Migrates e2e tests from ginkgo,gomega to testing,testify

codefromthecrypt:
To call ginkgo,gomega indirect would be putting things lightly. A combination of an indirect testing library and long-running tests are a recipe for frustration. I've begun a process to remove ginkgo,gomega with a couple tests
that I had been looking at.

See #127

## Some benefits
* Before, we both had tests using normal testing,testify and also ginkgo,gomega. Now we only have way to write tests: testing,testify.
* testing,testify are more popular and easier to learn than ginkgo,gomega. testing is the standard library described in all golang books.  testify is a simple assertion library 10x more popular than gomega in terms of star count. Neither require learning BDD concepts.
* Before, our running of e2e tests was significantly different than our other non-gingko tests in the same project. It literally required compilation of a platform-specific binary named 'e2e', adding a lot of complexity to CI and local setup.
* Before, individual tests cannot be executed via normal means, such as Goland. Instead, you have to run the suite file. Now, you can execute and more importantly debug individual tests instead of via a suite wrapper.
* Before, writing tests not only required understanding of the subject, but also the ginkgo runtime. Otherwise, statements like `var _ = Describe("getenvoy extension init", func() {` seem pure magic.
* Before, the nature of test execution required a combination of understanding the go platform and also ginkgo
* Before, updating Ginkgo/gomega would skew the dependency tree, due to their dependencies on two protobuf libraries. Now, the test runtime does not interfere with main code.
* Before, the same test took more lines of code to achieve the same table test. Less code is less to maintain.
* Some other notes about migration off are captured in this issue https://github.com/openservicemesh/osm/issues/1704

## Notable differences

Gomega had a different way of logging. For example instead of using a logging library, it would use statements like this:
`By("changing to the output directory")` which translate into output `STEP: changing to the output directory`. This is mainly to satisfy BDD idioms and is meant to be interpreted amongst the many lines of normal log output. For cases where the points inside the test should be logged, we should use normal logging instead. However, I don't recommend making log statements unless they are meaningful and clarify progress.

Also, those not used to it will notice use of normal go `defer` commands instead of the before/after test hooks used in tools like gomega. This can take getting used to, but it is also the same as normal go code, so shouldn't be time lost. There are some cases where multiple tear-down steps are needed and those steps are more visible now. That's not necessarily bad because these typically are also the most complex tests and have a lot going on. Seeing the various setup requirements shows problem areas that help knowing when debugging.


codefromthecrypt:
Here's an example of the tests executing in Goland

Screen Shot 2021-03-24 at 3 57 38 PM

mathetake:
+1 Personally, I'm all in for this migration, and I've been frustrated too. 

I would like to have @yskopets's opinition.

yskopets:
@codefromthecrypt 

> A combination of an indirect testing library and long-running tests are a recipe for frustration.

Can you explain what it means ? What has changed after this PR ?

> While the rationale is fairly straightforward, the following issue captures a lot of it nicely

Personally, I don't agree with this rationale.

I don't see from the linked issue how much support it received.

It's labeled with `needs discussion` for a reason.

If I understand the linked issue correctly, it has only 1 argument - "Let's move to Go std library because Ginkgo/Gomega is not a Go std library."

Sure. Developers choose `Ginkgo/Gomega` when they want to do BDD-style testing (which is uniform across all programming languages). If Go std library cannot offer it, developers have to switch to other libraries that can.  

> I've begun a process to remove gomega with a couple tests that I had been looking at.

I'm looking at the replacement code of this PR and I don't understand what component has become better now:
* is it more readable ?
* does it give better assertion messages when the test fails ?
* does it leave a better trail in the test log ?

> I don't mind finishing the e2e suite

It's a team/community work. If majority of contributors are in favour of this change, I won't object.

Please be mindful, though, about people who will have to maintain these tests.

If you're not sticking around to maintain the refactored code, please consider the effect of such changes on other developers.

codefromthecrypt:
The highest value from this so far is that the nested test is directly executable now, as opposed to unrunnable in Goland, for example. I would like to hear from others besides you @yskopets because I have noticed a lot of projects copy/paste things even when they don't agree with it. There are likely more people who have only tolerated this gomega thing due to difficulty getting rid of it.

test/e2e/getenvoy_extension_test_test.go
```diff
@@ -16,59 +16,57 @@ package e2e_test
...
-
-	type testCase e2e.CategoryLanguageTuple
-
-	testCases := func() []TableEntry {
-		testCases := make([]TableEntry, 0)
-		for _, combination := range e2e.GetCategoryLanguageCombinations() {
-			testCases = append(testCases, Entry(combination.String(), testCase(combination)))
-		}
-		return testCases
+// TestGetEnvoyExtensionTest runs the equivalent of "getenvoy extension test" for a matrix of extension.Categories and
```
codefromthecrypt, line 27: for the doubters.. I've increased the documentation and there are less and more direct output here. I highly doubt someone could say this is a net loss vs what was here before.

codefromthecrypt:
@mathetake I'd like to merge what's here. Later, I will do a PR to finish the rest, but landing incremental progress makes debugging #129 far easier, which is what started all of this really.

codefromthecrypt:
a different test failed while looking at go:embed and debuggability problems are the same as the one I've already converted. I'll go ahead and finish this PR vs doing things piece-wise. I don't usually like huge PRs, but I like whack-a-mole less

codefromthecrypt:
found this bug while visiting the push tests I'll rebase once it is merged https://github.com/tetratelabs/getenvoy/pull/147

codefromthecrypt:
will be another day or two to finish. the "run" tests are rather gnarly so saved them for last :p

codefromthecrypt:
the "getenvoy extension run" unit tests are now done. I still need to do the e2e for that, and that's probably where I'll stop as the "getenvoy extension" tests were the more difficult ones I ran into, leading to this issue. I'll update soon and probably separate out the "unit tests" from the e2e.

I quote "unit tests" as they aren't really like normal unit tests, as they do invoke OS commands etc. The whole unit test run could be much faster if this was faked out, resulting in less temp dirs and process calls. I won't have time to address that, though.

codefromthecrypt:
I pulled the unit test changes into a separate PR as they were instrumental in figuring out what was going on in the e2e tests. https://github.com/tetratelabs/getenvoy/pull/151

This test is purely addressing the e2e, but other benefits such as less dependencies won't be realized until other tests stop using gingko/gomega. Sadly, just this part took so much time I don't have time to port all remaining tests. OTOH, it should be far easier with this and #151 in.

codefromthecrypt:
run phases fail on timeout locally and in CI for rust. I'll look into it, but I raised an issue as generally speaking everything rust related takes a huge amount of time. We should probably bump the matrix so that the language e2e runs is managed externally and defaults to tinygo https://github.com/tetratelabs/getenvoy/issues/152

codefromthecrypt:
kicked e2e tests due to #145 (current script we are using routinely times out installing docker on osx)

codefromthecrypt:
green build now. @mathetake PTAL

Screen Shot 2021-03-31 at 10 25 23 AM

.github/workflows/release.yaml
```diff
@@ -71,33 +71,11 @@ jobs:
       - name: "Push extension builder images"
         run: make builders.push BUILDERS_TAG=${{ env.RELEASE_VERSION }}
 
-  e2e_bin:
```
mathetake, line 74: nice. I was thinking that this is not needed but I was too lazy smile

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,162 @@
+// Copyright 2021 Tetrate
```
mathetake, line 1: I like these utility functions. nice.

test/e2e/getenvoy_extension_build_test.go
```diff
@@ -15,70 +15,48 @@
...
+// extension.Languages. "getenvoy extension init" is a prerequisite, so run first.
+//
+// Note: "getenvoy extension build" can be extremely slow due to implicit responsibilities such as downloading modules
+// or compilation. This uses Docker, so changes to the Dockerfile or contents like "commands.sh" effect performance.
+//
+// Note: Pay close attention to values of util.E2E_BUILTIN_TOOLCHAIN_CONTAINER_OPTIONS as these can change assumptions.
+// CI may override this to set HOME or CARGO_HOME (rust) used by "getenvoy" and effect its execution.
+func TestGetEnvoyExtensionBuild(t *testing.T) {
+	const extensionName = "getenvoy_extension_build"
+	requireEnvoyBinaryPath(t) // Ex. After running "make bin", E2E_GETENVOY_BINARY=$PWD/build/bin/darwin/amd64/getenvoy
```
mathetake, line 36: maybe we'd better move `requireEnvoyBinaryPath` to `TestMain` since it's enough to check at the startup time?

mathetake: so that we don't need to copy and paste this in every e2e :smile:

test/e2e/getenvoy_extension_init_test.go
```diff
@@ -15,94 +15,89 @@
...
+// "getenvoy extension init" should be fast and reliable because it doesn't use Docker.
+func TestGetEnvoyExtensionInit(t *testing.T) {
+	const extensionName = "getenvoy_extension_init"
+	requireEnvoyBinaryPath(t) // Ex. After running "make bin", E2E_GETENVOY_BINARY=$PWD/build/bin/darwin/amd64/getenvoy
+
+	type testTuple struct {
+		testName string
+		extension.Category
+		extension.Language
+		currentDirectory bool
```
mathetake, line 42: looks like `currentDirectory` is always false. Am I missing something?

mathetake: ah - sorry nvm

test/e2e/getenvoy_extension_build_test.go
codefromthecrypt: sgtm

test/e2e/main_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+// TestMain ensures state required for all tests, notably that util.E2E_GETENVOY_BINARY is set.
+//
+// Note: "getenvoy extension build" and commands that imply it, can be extremely slow due to implicit responsibilities
+// such as downloading modules or compilation. Commands like this use Docker, so changes to the Dockerfile or contents
+// like "commands.sh" will effect performance.
+//
+// Note: Pay close attention to values of util.E2E_BUILTIN_TOOLCHAIN_CONTAINER_OPTIONS as these can change assumptions.
+// CI may override this to set HOME or CARGO_HOME (rust) used by "getenvoy" and effect its execution.
+func TestMain(m *testing.M) {
```
mathetake, line 49: nice :relaxed:

test/e2e/util/command.go
```diff
@@ -61,30 +71,37 @@ func (b *cmdBuilder) Exec() (string, string, error) {
...
-			errs <- err
-		}
+		errc <- b.cmd.Wait()
 	}()
-	canceled := false
-	cancel := func() error {
-		if canceled {
-			return nil
+
+	// Ensure terminate is only called once. As this happens from a single thread, there's no need to lock.
```
mathetake, line 87: maybe we could use `sync.Once` rather than using closure capturing the state `terminated` :smile:

mathetake: ah - this would make the consistency of the returned function of these utilities. Fine as it is!

test/e2e/main_test.go
```diff
@@ -31,16 +32,30 @@ import (
...
+	// As this is an e2e test, we execute all tests with a binary compiled earlier.
+	//
+	// Ex. After running "make bin", E2E_GETENVOY_BINARY=$PWD/build/bin/darwin/amd64/getenvoy
+	_, err := e2e.Env.GetEnvoyBinary()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, `failed to start e2e tests: %v`, err)
+		os.Exit(1)
+	}
```
mathetake, line 57: I guess you've dropped `e2e.GetEnvoyBinaryPath = path` mistakenly

codefromthecrypt: indeed and I wish I saw your comment when you wrote it instead of now :) fixed!

mathetake:
I believe the value props @codefromthecrypt mentioned are now achieved, and e2e tests now become much simpler than before. Notably, 1) not using Ginkgo or any related stuff would definitely lower the onboarding cost for folks both from Go and other languages including me, and also 2) the better integration with IDEs from zero. LGTM to me now.

test/e2e/util/command.go
```diff
@@ -61,30 +71,37 @@ func (b *cmdBuilder) Exec() (string, string, error) {
...
+	terminated := false
+	terminate := func() {
+		if terminated {
+			return
+		}
```
mathetake, line 92: question: when does this function reenter? If the reentrance should not happen, shall we make this `require.False(t, terminated)`?

codefromthecrypt: we test the debug data which requires stopping the server first. So, the defer hook later invokes it again.

codefromthecrypt: I moved the pre-emption up, where it is easier to see why. PTAL

codefromthecrypt:
thanks for the review and pointers @mathetake, as well the support behind the scenese. this is definitely better now

codefromthecrypt:
re-kicked the osx job as it is failing for the usual reason:
```
Waiting for docker service to be in the running state. 	If Docker is not ready within 10 minutes, it's better to fail the current CI Job and let the comitter to re-start the job manually
```

mathetake:
@musaprg any comments?

musaprg:
Personally, +1 to this PR, though I'm not sure it would be the best way in a community. I prefer the standard way in Go and I think it makes current tests more simple and readable.

musaprg:
also, ginkgo/gomega is a little bit complicated for contributors who new to them, for me at first as well.

mathetake (approved):
Thanks @codefromthecrypt for this awesome work. Will merge after all the jobs pass

mathetake:
note that now we can run individual test in a click of button 
Screenshot from 2021-03-31 14-00-31
Screenshot from 2021-03-31 14-00-18

Issue #100, opened 2020-09-02
ci: move all CI steps from CircleCI to GitHub Actions


yskopets:

codefromthecrypt:
Agree with this as the differences between CI aren't helpful.

codefromthecrypt:
This is locked up due to some tests failing severely when executed in GH actions. Here's the snippet that ideally would have things work

```yaml
  check:
    name: "Run CI checks on Linux"
    runs-on: ubuntu-latest
    container: golang:1.16  # match go.mod version
    steps:
      - name: "Checkout"
        uses: actions/checkout@v2

      - name: "Install Go"
        uses: actions/setup-go@v2
        with:
          go-version: '1.16.2'

      - name: "Init on first use"
        run: make init

      - name: "Run checks"
        run: make check

  test:
    name: "Run unit tests on Linux"
    runs-on: ubuntu-latest
    steps:
      - name: "Checkout"
        uses: actions/checkout@v2

      - name: "Install Go"
        uses: actions/setup-go@v2
        with:
          go-version: '1.16.2'

      - name: "Init on first use"
        run: make init

      - name: "Prefetch latest envoy"
        run: go run cmd/getenvoy/main.go fetch standard:1.17.1

      - name: "Run tests"
        run: make test GO_TEST_EXTRA_OPTS="-timeout 60s -v"

      - name: "Generate coverage report"
        run: make coverage GO_COVERAGE_EXTRA_OPTS="-p 1"

      - name: "Upload coverage report"
        uses: actions/upload-artifact@v2
        with:
          name: coverage-report
          path: build/coverage
```

the check task works, test runs a few until eventually they blow up like so:
```
ok  	github.com/tetratelabs/getenvoy/pkg/binary/envoy/controlplane	21.264s
go build github.com/tetratelabs/getenvoy/pkg/cmd.test: open /tmp/go-build1173000906/b1040/importcfg: no such file or directory
# github.com/russross/blackfriday
vet: in github.com/russross/blackfriday, can't import facts for package "bytes": open $WORK/b038/vet.out: no such file or directory
# github.com/cpuguy83/go-md2man/md2man
vet: in github.com/cpuguy83/go-md2man/md2man, can't import facts for package "bytes": open $WORK/b038/vet.out: no such file or directory
# github.com/spf13/cobra/doc
vet: in github.com/spf13/cobra/doc, can't import facts for package "bytes": open $WORK/b038/vet.out: no such file or directory
[...]

PR #132, opened 2021-03-25
Moves lint and format to make, managed by bingo

### LINKED ISSUE #131: self-contain binaries needed for the build
Right now, I notice differences in workstation runs vs what's in CI. Binaries needed for lint and format are not managed in the `Makefile`, and I've often had different behavior between runs locally and what CI catches.

I'd like to see make manage the binaries used, besides go and make itself. This reduces the chance for errors. I would like to see less shell scripts needed for setup.

For example, grpc-middleware uses bingo for this https://github.com/grpc-ecosystem/go-grpc-middleware/tree/v2

https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/.bingo/Variables.mk#L44-L48

```
GOLANGCI_LINT := $(GOBIN)/golangci-lint-v1.27.0
$(GOLANGCI_LINT): $(BINGO_DIR)/golangci-lint.mod
	@# Install binary/ries using Go 1.14+ build command. This is using bwplotka/bingo-controlled, separate go module with pinned dependencies.
	@echo "(re)installing $(GOBIN)/golangci-lint-v1.27.0"
	@cd $(BINGO_DIR) && $(GO) build -mod=mod -modfile=golangci-lint.mod -o=$(GOBIN)/golangci-lint-v1.27.0 "github.com/golangci/golangci-lint/cmd/golangci-lint"
```

https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/Makefile#L87-L94

```
lint_module: $(FAILLINT) $(GOLANGCI_LINT) $(MISSPELL)
--snip--
	@cd $(DIR) && $(GOLANGCI_LINT) run
```

Bingo generates the makefile snippets which then help pin the binary and version, also managing implicitly building them.

Another way is to use docker to run the build (can be tricky especially with local module overrides and caching of deps)

codefromthecrypt:
This removes a few shell scripts and in the process updates linters to
more current versions. For example, we were using golangci from 2019. To
adjust, some invalid statements were removed.

This also adds goimport formatting, used by some of our other projects.

The main part of this change is using https://github.com/bwplotka/bingo
as the linters we used were go programs anyway. Bingo generates make
entries you can use to on-demang install tools without affecting your
go.mod and without shell scripts.

Fixes #131


.bingo/.gitignore
```diff
@@ -0,0 +1,12 @@
+
```
codefromthecrypt, line 1: NOTE: files in .bingo are generated, so you can ignore them

.circleci/config.yml
```diff
@@ -32,18 +32,16 @@ jobs:
...
+
   test:
     executor: builder
     resource_class: medium+
     steps:
       - checkout
       - run: make init
-      - run: ./ci/install-envoy
-      - run: ./ci/test
+      - run: go run cmd/getenvoy/main.go fetch standard:1.11.0
```
codefromthecrypt, line 43: this deserves a comment, if someone can remember why this was done.

.golangci.yml
```diff
@@ -93,67 +91,30 @@ linters-settings:
...
-      - equalFold
-      - flagDeref
-      - flagName
       - hexLiteral
-      - indexAlloc
       - initClause
       - methodExprCall
       - nilValReturn
       - octalLiteral
-      - offBy1
```
codefromthecrypt, line 127: removed all these lines to get rid of redundant warnings in output

pkg/binary/envoy/config.go
```diff
@@ -97,7 +97,7 @@ func (r *Runtime) SaveConfig(name, config string) (string, error) {
 		return "", fmt.Errorf("Unable to create directory %q: %v", configDir, err)
 	}
 	filename := name + ".yaml"
-	err := ioutil.WriteFile(filepath.Join(configDir, filename), []byte(config), 0644)
+	err := ioutil.WriteFile(filepath.Join(configDir, filename), []byte(config), 0600)
```
codefromthecrypt, line 100: new version of the sec linter is more strict. I went with the advice instead of ignoring it.

.circleci/config.yml
mathetake: would it be possible to change this to `standard:1.17.1` which is the current latest?

Makefile
```diff
@@ -175,3 +181,32 @@ $(foreach lang,$(BUILDERS_LANGS),$(eval $(call GEN_PULL_EXTENSION_BUILDER_IMAGE_
...
+# To avoid that, before running it, we collapse all imports in one block, then run the formatter.
+.PHONY: format
+format: $(GOIMPORTS) ## Format all Go code
+	@echo "--- format ---"
+	@$(LICENSER) apply -r "Tetrate"
+	@find . -type f -name '*.go' | xargs gofmt -s -w
+	@for f in `find . -name '*.go'`; do \
+	    awk '/^import \($$/,/^\)$$/{if($$0=="")next}{print}' $$f > /tmp/fmt; \
+	    mv /tmp/fmt $$f; \
+	    $(GOIMPORTS) -w -local github.com/tetrateio/getenvoy $$f; \
```
mathetake, line 206: I believe this should be tetratelabs :joy: 
```suggestion
	    $(GOIMPORTS) -w -local github.com/tetratelabs/getenvoy $$f; \
```

.circleci/config.yml
codefromthecrypt: yeah let's

Makefile
```diff
@@ -175,3 +181,32 @@ $(foreach lang,$(BUILDERS_LANGS),$(eval $(call GEN_PULL_EXTENSION_BUILDER_IMAGE_
...
+##@ Code quality and integrity
+
+LINT_OPTS ?= --timeout 5m
+.PHONY: lint
+# generate must be called while generated source is still used
+lint: generate $(GOLANGCI_LINT) $(SHFMT) $(LICENSER) .golangci.yml  ## Run the linters
+	@echo "--- lint ---"
+	@$(SHFMT) -d .
+	@$(LICENSER) verify -r .
+	@$(GOLANGCI_LINT) run $(LINT_OPTS) --config .golangci.yml
```
mathetake, line 194: I think we should also check if it's goimport formatted with something like `${goimport execute ...} && git diff --exit 1`

codefromthecrypt: ok I will add this to the "check" task in make similar to some other projects

codefromthecrypt: PTAL! thanks and good idea

mathetake: super! :+1:

codefromthecrypt:
@mathetake can you disable circleci? I finished the almost done migration here https://app.circleci.com/settings/project/github/tetratelabs/getenvoy

.github/workflows/commit.yaml
```diff
@@ -1,85 +1,123 @@
...
     branches:
-    - master
+      - master
   pull_request:
     branches:
-    - '**'
+      - '**'
 
 jobs:
+  check:
```
codefromthecrypt, line 16: these two were from circleci

codefromthecrypt:
some kooky stuff in tests like this. I'll get to the bottom of it..

```
vet: in github.com/containerd/containerd/remotes, can't import facts for package "context": open $WORK/b002/vet.out: no such file or directory
open /tmp/go-build2538038221/b1114/vet.cfg: no such file or directory
```

codefromthecrypt:
I think I have to rollback the circleci -> GH action thing until we can kill ginkgo #130 as normal tests are passing, but then the whole thing crashes. I was hoping I wouldn't have to fix circleci, but seems the case.

I'll summarize the crashing thing in #100

codefromthecrypt:
UPDATE: there's a panic in #133, once that's sorted I'll rebase this, eliminating the circleci thing, which I've documented in #100

yskopets:
@codefromthecrypt 

> I think I have to rollback the circleci -> GH action thing until we can kill ginkgo #130 as normal tests are passing

This statement is simply not true:
* here is a CI build (based on this PR) that runs all Go std tests and excludes Ginkgo-based ones - it is [failing](https://github.com/tetratelabs/getenvoy/runs/2192351046?check_suite_focus=true)
* here is a CI build (based on this PR) that runs all Ginkgo-based tests and non-problematic Go std tests - it is [successful](https://github.com/tetratelabs/getenvoy/runs/2192425605?check_suite_focus=true)

codefromthecrypt:
rebased over the go 1.16 PR and removed the circleci related changes

codefromthecrypt:
failure seems to repeat. will look into it..

```
/home/circleci/.getenvoy/builds/standard/1.11.0/linux_glibc/bin/envoy: text file busy
```

codefromthecrypt:
opened #136 about the strange test dep on a hard-coded envoy version

codefromthecrypt:
FYI "Run unit tests on Linux" is stale.. this is a known issue when you have added GH actions then remove it on the same PR. It still shows up in the output.

mathetake (approved):
thx!

codefromthecrypt:
thanks for helping @mathetake

Issue #140, opened 2021-03-26
ci: significant regression in run time of `e2e` tests on MacOS


yskopets:
**Describe the bug**

* prior to https://github.com/tetratelabs/getenvoy/pull/135, the run time of `e2e` tests on MacOS was about `~42 min`, e.g. [here](https://github.com/tetratelabs/getenvoy/runs/2189593019)
* after https://github.com/tetratelabs/getenvoy/pull/135, the run time of `e2e` tests on MacOS is `3+ hours` and counting, e.g. [here](https://github.com/tetratelabs/getenvoy/runs/2198918254?check_suite_focus=true) and [here](https://github.com/tetratelabs/getenvoy/runs/2199262800?check_suite_focus=true)

**To Reproduce**
1. Check out the run time of CI `Run e2e tests on MacOS` job since https://github.com/tetratelabs/getenvoy/pull/135

**Expected behavior**
* As a contributor, I would prefer faster feedback from `e2e` tests

**Screenshots**
* <img width="1122" alt="Screenshot 2021-03-26 at 08 42 53" src="https://user-images.githubusercontent.com/19395458/112598694-42b8e800-8e0f-11eb-881c-bdf6bc4e440c.png">
* <img width="1179" alt="Screenshot 2021-03-26 at 08 42 01" src="https://user-images.githubusercontent.com/19395458/112598628-2452ec80-8e0f-11eb-85f8-bbecb77cf009.png">

codefromthecrypt:
note before #135 <https://github.com/tetratelabs/getenvoy/pull/135> just
the installation part of docker for mac routinely hung on PRs, which is why
I looked at it in the first place.

>

codefromthecrypt:
specifically, it would hang on the output "Waiting for docker service to be in the running state"

yskopets:
* Is a CI build that takes `3,5+ hours` instead of `42min` better ?
* Why #135 was merged without waiting until CI build completes successfully ?

codefromthecrypt:
before this merged even docker install wouldn't finish.

https://github.com/tetratelabs/getenvoy/runs/2198242029?check_suite_focus=true

there are significant problems in that setup. I chatted with @mathetake and knowing that's the case, merged based on unix success. I think you also know there are significant issues in and outside CI with osx #88.

[...]

PR #141, opened 2021-03-26
Attempts a different docker for mac version

codefromthecrypt:
This is a wild guess at what could be behind #140


codefromthecrypt:
closing because switching back to the pinned version had no positive effect.

Issue #145, opened 2021-03-27
macOS+rust e2e tests in CI are pinned to an over two year old version of Docker


codefromthecrypt:
macOS+rust e2e tests in CI are pinned to an over two year old version of Docker, handled in `./ci/e2e/macos/install_docker.sh`. The rest of the matrix uses latest Docker per #159.

The impact is that we are blind to any issues that present only on recent versions of Docker + OS/x, such as #147. To mitigate this, people should manually run `make e2e` on a macbook on occasion.

The underlying problem is not that tests break, it is that they take too long to complete. The rust performance problem is being tracked in #152 and this issue should be revisited on any significant progress. However, this isn't a sure thing. For example, special casing in #158 resulted in half the time spent in macOS+rust e2e (on a modern MacBook Pro). However, GH actions run of those same tests failed on timeout after an hour.

There means there could be some possibilities that only change CI. For example, the descending performance could be related to memory or other environment details that differ between CI and a macbook pro. In other words, even if work on the root problem is slow progressing, there could be some changes to make things "fast enough" (40m?) to use normal Docker again.

lizan:
IIRC it is pinged to older version because newer Docker Desktop doesn't install on CI machines, if we can address that it would be great.

codefromthecrypt:
@lizan thx. to clarify the script I mention above does install, build images, run etc. Just we have some tests that are failing (in a bad way, just hanging). Hopefully the general cleanup work on tests can get us to a point where we can understand why and put the unpinned script back in (or some variant)

codefromthecrypt:
the `go test -timeout` flag kicked in to kill the tests at 45m. Since locally I had a problem resolved by #147, it is possible we are only down to runtime slowness.

I'll bump the timeout on that branch to see the baseline of an unpinned docker+macos.

![Screen Shot 2021-04-01 at 7 10 08 AM](https://user-images.githubusercontent.com/64215/113222320-4c33cb80-92b9-11eb-9f09-018d9e4ad803.png)

codefromthecrypt:
I found that a plugin I used in the past supports OS/x with the same approach as here. Since we are not in the business of hosting shell scripts, I changed my branch to use this instead https://github.com/marketplace/actions/setup-docker

I've hesitated on running against it until #156 as it takes a long time to run both extension languages sequentially. Running these isolated will allow us to better understand the extent of things such as cargo flags.

codefromthecrypt:
[...]

PR #135, opened 2021-03-26
Formats CI yaml and switches to a recent install_docker script

codefromthecrypt:
The former script to install docker fails and also has dead links
describing why it was written. This replaces the script with one updated
this month, and with history and license info about where it came from.

Signed-off-by: Adrian Cole <adrian@tetrate.io>

PR #146, opened 2021-03-27
Reverts to old macos install_docker.sh

### LINKED ISSUE #140: ci: significant regression in run time of `e2e` tests on MacOS

codefromthecrypt:
The old script that was failing to install and swapped out in #135 works again. This reverts it until we can diagnose what's behind the hanging in #145, or something else fixes it.

Fixes #140

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -20,6 +20,7 @@ jobs:
   bin:
     name: "Build `getenvoy` and `e2e` binaries for use in e2e tests"
     runs-on: ubuntu-latest
+    timeout-minutes: 15  # instead of 360 by default
     steps:
       - name: "Checkout"
         uses: actions/checkout@v2
@@ -46,6 +47,7 @@ jobs:
     needs:
       - bin
     runs-on: ubuntu-latest
+    timeout-minutes: 30  # instead of 360 by default
     steps:
       - name: "Checkout"
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
FYI this script is back to timing out again, but it did manage to finish at least twice in the last 24 hours. That this fails will be normal until the root cause is investigated and fixed as described in #145

yskopets (approved):

yskopets:
@codefromthecrypt Thanks for spotting that I missed updating timeouts in `.github/workflows/release.yaml` ! 👍

PR #147, opened 2021-03-28
Fixes bug where fully resolved image ref wasn't used

codefromthecrypt:
Recent docker tools require fully resolved image tags (ex :latest
explicitly). While we ensured the latest tag was appended, we didn't use
it. I noticed this pushing to my local registry on latest OS/x and Docker.


mathetake (approved):

mathetake:
Thank you!

Issue #137, opened 2021-03-26
Switch tests to use require not t.Errorf


codefromthecrypt:
`t.Errorf` doesn't fail the tests and leads to panics like below

```
testing.tRunner.func1.2(0x908360, 0xce3ee0)
	/usr/local/go/src/testing/testing.go:1144 +0x332
testing.tRunner.func1(0xc0001fad80)
	/usr/local/go/src/testing/testing.go:1147 +0x4b6
panic(0x908360, 0xce3ee0)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
github.com/tetratelabs/getenvoy/pkg/binary/envoy/debug.Test_debugging_outputs.func1(0xc0001fad80)
	/home/circleci/project/pkg/binary/envoy/debug/node_test.go:44 +0x3ea
testing.tRunner(0xc0001fad80, 0x9b8388)
	/usr/local/go/src/testing/testing.go:1194 +0xef
created by testing.(*T).Run
```

We shouldn't use this. Instead move to require assertions that fail immediately, thus simplifying code and preventing panics.

Ex.
```diff
-                       if err != nil {
-                               t.Errorf("error stating %v: %v", path, err)
-                       }
+                       require.NoError(t, err, "error stating %v", path)
```

codefromthecrypt:
this is still an issue

codefromthecrypt:
long since done now

PR #148, opened 2021-03-30
Normalizes Reference.Platform to lower-hyphen format

codefromthecrypt:
Before, we mixed format of `Reference.Platform` to sometimes enum and
other times lower-hyphen format. This is confusing and leads to subtle
bugs. For example, some code did simply a lowercase and forgot to
convert underscores to hyphens.

This change normalizes to always write lower-hypen.

This also fixes a couple tests which panic'ed due to #137, but were
otherwise unrelated.

### DIFF
--- a/pkg/binary/envoy/fetch.go
+++ b/pkg/binary/envoy/fetch.go
@@ -23,7 +23,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/mholt/archiver"
 	"github.com/schollz/progressbar/v2"
@@ -87,9 +86,7 @@ func (r *Runtime) BinaryStore() string {
 }
 
 func (r *Runtime) platformDirectory(key *manifest.Key) string {
-	platform := strings.ToLower(key.Platform)
-	platform = strings.ReplaceAll(platform, "-", "_")
-	return filepath.Join(r.BinaryStore(), key.Flavor, key.Version, platform)
+	return filepath.Join(r.BinaryStore(), key.Flavor, key.Version, key.Platform)
 }
 
 func fetchEnvoy(dst, src string) error {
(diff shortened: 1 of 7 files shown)


pkg/types/reference.go
```diff
@@ -39,7 +47,7 @@ func ParseReference(text string) (*Reference, error) {
 	if len(matches) != 4 {
 		return nil, errors.Errorf("%q is not a valid GetEnvoy reference. Expected format: <flavor>:<version>[/<platform>]", text)
 	}
-	return &Reference{strings.ToLower(matches[1]), strings.ToLower(matches[2]), strings.ToLower(matches[3])}, nil
```
codefromthecrypt, line 42: this was the only place where we wrote the value missing the hyphen conversion I noticed.

codefromthecrypt:
This failed due to rate-limiting. I'll rebase and push after #150

pkg/binary/envoy/fetch.go
```diff
@@ -87,9 +86,7 @@ func (r *Runtime) BinaryStore() string {
 }
 
 func (r *Runtime) platformDirectory(key *manifest.Key) string {
-	platform := strings.ToLower(key.Platform)
-	platform = strings.ReplaceAll(platform, "-", "_")
-	return filepath.Join(r.BinaryStore(), key.Flavor, key.Version, platform)
+	return filepath.Join(r.BinaryStore(), key.Flavor, key.Version, key.Platform)
```
mathetake, line 89: just want to confirm that this function is covered by `TestRuntime_Fetch` in test?

mathetake (approved):

pkg/binary/envoy/fetch.go
codefromthecrypt: it isn't needed anymore because we are no longer creating invalid platform keys. (tests of conversion ending up in the correct place are elsewhere in locate_test)

codefromthecrypt:
thanks for the look!

PR #151, opened 2021-03-30
Rewrites trickiest cmd tests to help clarify what's happening

codefromthecrypt:
Before, the cmd unit tests were difficult to understand. They are still
complicated and difficult, but far better documented now. Moreover, by
using standard testing libs, individual tests and even table cells can
be driven by the IDE, and failures point to a relevant location in a
table also.

See #130 for further rational about ginkgo/gomega vs testing/testify


pkg/cmd/extension/build/cmd_test.go
```diff
@@ -15,186 +15,162 @@
...
-	chdir := func(path string) string {
-		dir, err := filepath.Abs(path)
-		Expect(err).ToNot(HaveOccurred())
-
-		err = os.Chdir(dir)
-		Expect(err).ToNot(HaveOccurred())
-
-		return dir
+func TestGetEnvoyExtensionBuildValidateFlag(t *testing.T) {
+	type TestCase struct {
```
mathetake, line 32: guess we don't need to export the inline type :smile:

codefromthecrypt: good eye!

codefromthecrypt:
opened https://github.com/tetratelabs/getenvoy/issues/154 about the flakey test as there are about 4 that could break the same way

pkg/test/cmd/extension/command.go
```diff
@@ -0,0 +1,234 @@
+// Copyright 2021 Tetrate
```
mathetake, line 1: If this is only for tests, I suggest we suffix `_test.go` so it can not be compiled in the last binary

pkg/cmd/extension/build/cmd_test.go
```diff
@@ -15,186 +15,162 @@
...
-			c.SetArgs([]string{"extension", "build"})
-			err := cmdutil.Execute(c)
-			Expect(err).To(HaveOccurred())
+	for _, test := range tests {
+		test := test // pin! see https://github.com/kyoh86/scopelint for why
 
-			By("verifying command output")
-			Expect(stdout.String()).To(BeEmpty())
-			Expect(stderr.String()).To(Equal(fmt.Sprintf(`Error: there is no extension directory at or above: %s
+		t.Run(test.flag, func(t *testing.T) {
```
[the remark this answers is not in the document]
codefromthecrypt: sgtm

pkg/test/cmd/extension/command.go
codefromthecrypt: sgtm

mathetake: I realized that this package is contained in the `pkg/test` dir. so I think this is fine as it is!

mathetake, line 1: but I realized that this test utils don't have any `extension` specific stuff so we could move this to `pkg/test/cmd/command.go`. Even we have `NewRootCommand` which literally returns the root command of getenvoy which is not extension specific :smile:

pkg/cmd/extension/clean/cmd_test.go
```diff
@@ -15,186 +15,162 @@
...
-			c.SetArgs([]string{"extension", "clean"})
-			err := cmdutil.Execute(c)
-			Expect(err).To(HaveOccurred())
+	for _, test := range tests {
+		test := test // pin! see https://github.com/kyoh86/scopelint for why
 
-			By("verifying command output")
-			Expect(stdout.String()).To(BeEmpty())
-			Expect(stderr.String()).To(Equal(fmt.Sprintf(`Error: there is no extension directory at or above: %s
+		t.Run(test.flag, func(t *testing.T) {
```
mathetake, line 54: ditto

pkg/cmd/extension/push/cmd_test.go
```diff
@@ -15,122 +15,84 @@
...
+
+	// Verify the command failed with the expected error
+	expectedErr := "there is no extension directory at or above: " + dir
+	require.EqualError(t, err, expectedErr, `expected an error running [%v]`, cmd)
+	require.Empty(t, stdout.String(), `expected no stdout running [%v]`, cmd)
+	expectedStderr := fmt.Sprintf("Error: %s\n\nRun 'getenvoy extension push --help' for usage.\n", expectedErr)
+	require.Equal(t, expectedStderr, stderr.String(), `expected stderr running [%v]`, cmd)
+}
+
+// TestGetEnvoyExtensionPushWithExplicitFileOption shows
```
mathetake, line 78: looks like the comment is incomplete :smile:

pkg/cmd/extension/test/cmd_test.go
```diff
@@ -164,15 +164,15 @@ Run 'getenvoy extension test --help' for usage.
...
+				"--toolchain-container-options", `-e DOCKER_EXIT_CODE=3`,
 			})
 			err := cmdutil.Execute(c)
 			Expect(err).To(HaveOccurred())
 
 			By("verifying command output")
-			Expect(stdout.String()).To(Equal(fmt.Sprintf("%s/docker run -u 1001:1002 --rm -t -v %s:/source -w /source --init -e EXIT_CODE=3 build/image test\n", dockerDir, workspaceDir)))
+			Expect(stdout.String()).To(Equal(fmt.Sprintf("%s/docker run -u 1001:1002 --rm -t -v %s:/source -w /source --init -e DOCKER_EXIT_CODE=3 build/image test\n", dockerDir, workspaceDir)))
 			Expect(stderr.String()).To(Equal(fmt.Sprintf(`docker stderr
-Error: failed to unit test Envoy extension using "default" toolchain: failed to execute an external command "%s/docker run -u 1001:1002 --rm -t -v %s:/source -w /source --init -e EXIT_CODE=3 build/image test": exit status 3
```
mathetake, line 175: I would like you to convert `cmd_test.go` to non-ginkgo as well if you have enough energy :smile:

codefromthecrypt: sure.. makes sense to do the entire package and I missed some

pkg/cmd/extension/push/cmd_test.go
codefromthecrypt: someone died while writing

pkg/extension/workspace/example/runtime/getenvoy/runtime_test.go
```diff
@@ -17,166 +17,121 @@ package getenvoy_test
...
+envoy args: -c %s/envoy.tmpl.yaml
+`, envoyWd, ctx.Opts.Envoy.Path, envoyWd)
+	require.Equal(t, expectedStdout, stdout.String(), `expected stdout running [%v]`, ctx)
+
+	// Verify we didn't accidentally combine the stderr of envoy into stdout, or otherwise dropped it.
+	require.Equal(t, "envoy stderr\n", stderr.String(), `expected stderr running [%v]`, ctx)
+}
+
+func TestRuntimeRunFailsOnInvalidWorkspace(t *testing.T) {
+	invalidWorkspaceDir := extension.RequireAbsDir(t, "../configdir/testdata/workspace5")
```
mathetake, line 66: I would like `"../configdir/testdata/workspace5"` to be const like `workspace1` above :smile:

pkg/extension/workspace/example/runtime/getenvoy/runtime_test.go
```diff
@@ -17,166 +17,121 @@ package getenvoy_test
...
-					},
-					Envoy: runtime.EnvoyOpts{
-						Path: envoyPath(),
-					},
-				},
-				IO: ioutil.StdStreams{
-					Out: stdout,
-					Err: stderr,
+// relativeWorkspaceDir points to a usable pre-initialized workspace
+const relativeWorkspaceDir = "../configdir/testdata/workspace1"
```
mathetake, line 34: shall we have `getenvoy/testadata` rather than depending on `configdir`'s testadata?

mathetake:
Finally I've looked at every case and understood them. LGTM so far (I want to have cmd_test.go free of Ginkgo as in my comment).

pkg/extension/workspace/example/runtime/getenvoy/runtime_test.go
codefromthecrypt: advantage of re-use is less things to copy/pasta on update. disadvantage is the opposite. In normal tests I often use a generic fixture so that it doesn't look like taking data from another tests, rather centralizing the fixture...

In this case I don't mind copying the data, but later we can think about this problem in general terms.

codefromthecrypt:
actually there are still a lot of tests left in the package. To save you from a blinding amount of unrelated changes, I kept the code related to your direct comments and commit to migrating the other tests in a subsequent PR, sg @mathetake?

mathetake:
sg

codefromthecrypt:
re-kicked as the osx thing is busted

mathetake (approved):

codefromthecrypt:
appreciate your at least 4 hours on this @mathetake!

Issue #153, opened 2021-03-31
Delete ci/e2e/*/run_tests.sh scripts


codefromthecrypt:
Debugging issues like #152 are made more difficult due to scripts used to execute the e2e tests. It is not clear why the HOME directory is overridden only in OS/X, fo example, and this effects the GETENVOY_HOME used in tests.

`make e2e` should be invoked regardless of it it is CI or not. Any directories overridden for caching reasons should happen in the github workflow, not a shell script. This makes what's being overridden easier to see and integrate with github cache actions.

Once #130 is merged, we should clean up more of the shell scripts and integrate config more directly, as this will lead to better understanding and cache integrations.

Issue #152, opened 2021-03-30
Optimize the rust builder image


codefromthecrypt:
Currently, "envoy extension build" on rust matrices are extremely slow. This hurts the dev experience and ties up CI and/or requiring careful consideration of external caches.

Here's a run from a latest macbook pro
```
--- PASS: TestGetEnvoyExtensionBuild (684.48s)
    --- PASS: TestGetEnvoyExtensionBuild/category=envoy.filters.http,_language=rust (217.16s)
    --- PASS: TestGetEnvoyExtensionBuild/category=envoy.filters.http,_language=tinygo (11.11s)
    --- PASS: TestGetEnvoyExtensionBuild/category=envoy.filters.network,_language=rust (212.49s)
    --- PASS: TestGetEnvoyExtensionBuild/category=envoy.filters.network,_language=tinygo (11.55s)
    --- PASS: TestGetEnvoyExtensionBuild/category=envoy.access_loggers,_language=rust (221.18s)
    --- PASS: TestGetEnvoyExtensionBuild/category=envoy.access_loggers,_language=tinygo (11.00s)
```

While some of this is unavoidable, I think our builder image hasn't been optimized and there are possibly many things to look at here.

One that could be handy is to use the somewhat typical "first run" inside the Dockerfile. Running a baseline command inside docker can insure users of the image don't need to do a lot of excessive downloading. This could be a cargo command.

Again, I'm not sure what the best change should be, but some analysis should happen because the status quo bleeds dev time.

codefromthecrypt:
this issue was obfuscated by overrides done in shell scripts. I opened #153 to hopefully save the next person from a similar fate. Regardless, there's likely use looking to pre-seed the image itself, as caching like this is complex and hides the poor execution time end users will perceive. That is assuming the runs are faster after some sort of seed operation.

Another option could be to consider the normal rust sdk to see if it is better optimized, as there's likely some large effort maintaining a fork and also optimizing it https://github.com/proxy-wasm/proxy-wasm-rust-sdk

codefromthecrypt:
@lizan FWIW here's what's going on per latest run:

```bash
grep Finished 3_Run\ e2e\ tests\ \(macos-latest\,\ rust\).txt 
2021-04-21T23:15:33.2248320Z     Finished dev [unoptimized + debuginfo] target(s) in 1m 52s
2021-04-21T23:17:18.6423800Z     Finished dev [unoptimized + debuginfo] target(s) in 22.57s
[...]

Issue #144, opened 2021-03-26
ci: consider removing use of `chrono` library from Rust init templates since it's the slowest to compile in `e2e` tests on Mac


yskopets:
**Is your feature request related to a problem? Please describe.**

* `e2e` tests on `Mac` are much slower than on `Linux` (e.g., `40 min` vs `10 min`)
* tests for extensions in `Rust` take `3x` more time than tests for `TinyGo` (because `Rust` extension templates have more dependent libraries and it simply takes more time to compile them all)
* it seems that `chrono` library is the slowest to compile

**Describe the solution you'd like**

* consider removing use of `chrono` library in Rust init templates to speed up `e2e` tests on `Mac`
* this might cut `~5 min` out of `40 min`

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

PR #156, opened 2021-04-01
Pulls `E2E_EXTENSION_LANGUAGE` into an external matrix

codefromthecrypt:
Before, end-to-end test executions were very slow. Originally, there was only one extension language. When this changed, the e2e tests took longer than before, and there was no quick way to run all tests when only changing one language. This is because the test case managed the matrix of extension language.

### Key changes
`make e2e` defaults to run all extension languages sequentially, just like before. However, running tests from an IDE defaults to "tinygo" and can be overridden to run all tests via `E2E_EXTENSION_LANGUAGE=all`. CI uses this same variable to break up the matrix, allowing parallel runs without giving up readability.

#### This allows more precise caching and overrides per-language
Right now, we have scripts that could be deleted with better management of our language. For example, #153 centers on overrides for rust like so:

```
export E2E_TOOLCHAIN_CONTAINER_OPTIONS="${E2E_TOOLCHAIN_CONTAINER_OPTIONS} -v ${E2E_CACHE_DIR}:/tmp/cache/getenvoy -e CARGO_HOME=/tmp/cache/getenvoy/extension/rust-builder/cargo"
```

A change here can help move those parameters into the CI, and only for the matrix option effected by them.

#### Defaulting to tinygo provides early feedback
Tinygo end-to-end tests finish dramatically quicker, because the "extension build" command completes faster (usually 20x). By defaulting the language matrix in tests to "tinygo", developers of these can have a much faster workflow without setting ENV variables every time. They can also have better insight into whether a test fail is related to them or not.

Users of `make e2e` are not effected by the default as in make `E2E_EXTENSION_LANGUAGE=all` and runs all languages sequentially like before.

#### This helps understand language-based runtime problems
See #144 and #152 for things easier to understand after this change. Notably, the rust runtime is extremely slow, and it is hard to understand improvements when tinygo is also run in the same matrix. Moreover, we can change timeouts according to language.

#### This seems optimized for now, where `make e2e` only testing extensions
It is understood that currently end-to-end tests happen to only test extensions and that this optimizes for the current codebase, not a future one that may have other matrix concerns. However, even if there are future matrices, a solution of running e2e per command group could work and leave this approach in-tact.

Rather than optimize for a future that is incompatible with this approach, this addresses now (and a future that is compatible). We sorely need to be efficient troubleshooting e2e to ensure developer velocity today.


ci/e2e/linux/run_tests.sh
```diff
@@ -26,7 +26,7 @@ mkdir -p "${E2E_CACHE_DIR}"
 sudo chown -R $(id -u):$(id -g) "${E2E_CACHE_DIR}"
 
 # to speed up `getenvoy extension build|test`, re-use a single cache across all extensions created by e2e tests
-export E2E_BUILTIN_TOOLCHAIN_CONTAINER_OPTIONS="${E2E_BUILTIN_TOOLCHAIN_CONTAINER_OPTIONS} -v ${E2E_CACHE_DIR}:/tmp/cache/getenvoy -e CARGO_HOME=/tmp/cache/getenvoy/extension/rust-builder/cargo"
+export E2E_TOOLCHAIN_CONTAINER_OPTIONS="${E2E_TOOLCHAIN_CONTAINER_OPTIONS} -v ${E2E_CACHE_DIR}:/tmp/cache/getenvoy -e CARGO_HOME=/tmp/cache/getenvoy/extension/rust-builder/cargo"
```
codefromthecrypt, line 29: it is less confusing when there aren't extra words in variable names

DEVELOPER.md
```diff
@@ -44,7 +44,13 @@ and follow the wizard.
 
 ### How to run e2e Tests
 
-Run:
-```shell
-make e2e
-```
+End-to-end (e2e) tests rely on a `getenvoy` binary that defaults to what was built by `make bin`.
```
codefromthecrypt, line 47: past me would like this, so will future

codefromthecrypt:
Example of early-feedback in action. tinygo tests completed, even on MacOS, in 14m.

Screen Shot 2021-04-01 at 8 55 41 AM

DEVELOPER.md
```diff
@@ -44,7 +44,13 @@ and follow the wizard.
...
+
+In simplest case, execute `make e2e` to run all tests configured. You can also set the below ENV variables to effect the
```
mathetake, line 49: this makes me think that `make e2e` run all tests for "all languages". So maybe it would be helpful if we add `make e2e all` or similar for running like that?

ci/e2e/linux/run_tests.sh
mathetake: +1

DEVELOPER.md
codefromthecrypt: maybe e2e.all as that wouldn't effect args?

mathetake: sg

Makefile
```diff
@@ -105,7 +106,14 @@ test.ci: generate
...
 e2e: $(call GETENVOY_OUT_PATH,$(GOOS),$(GOARCH))
 	docker-compose up -d
-	go test $(E2E_OPTS) $(E2E_EXTRA_OPTS) $(E2E_PKG_LIST)
+	$(RUN_E2E)
+
+# Runs e2e for all extensions languages in pkg/extension/workspace/config/extension/helpers.go
+.PHONY: e2e.all
+e2e.all: $(call GETENVOY_OUT_PATH,$(GOOS),$(GOARCH))
+	docker-compose up -d
+	E2E_EXTENSION_LANGUAGE=rust $(RUN_E2E)
```
codefromthecrypt, line 115: note: this is intentionally not fancy, as calling something twice is easier to understand than for loops in make

test/e2e/util/data.go
```diff
@@ -1,44 +0,0 @@
...
-	extension.Language
-}
-
-func (t CategoryLanguageTuple) String() string {
-	return fmt.Sprintf("category=%s, language=%s", t.Category, t.Language)
-}
-
-// GetCategoryLanguageCombinations returns all combinations of a supported
-// extension category with a supported programming language.
-func GetCategoryLanguageCombinations() []CategoryLanguageTuple {
```
mathetake, line 36: I just thought that using env var here is enough for what we want to achieve here. Then we can have `E2E_EXTENSION_LANGUAGE=all` instead of `make e2e.all`. WDYT?

mathetake: something like
```go
if envVar == "all" || language == envVar { tuples = append(tuples, CategoryLanguageTuple{category, language}) }
```

codefromthecrypt: I think I will always override this to tinygo because it will take too long. Remember how it was annoying to have to set the go binary every time we run e2e in the IDE? This will put that back for the language..

What if we had make default that to all, but go default it to tinygo?

mathetake: right. Thanks for clarifying.

Makefile
```diff
@@ -105,7 +106,14 @@ test.ci: generate
...
+	E2E_EXTENSION_LANGUAGE=tinygo $(RUN_E2E)
```
mathetake, line 116: not sure if the test results would be cached from the previous run for Rust here..

mathetake (approved):
LGTM except my concern about the cache https://github.com/tetratelabs/getenvoy/pull/156/files#r605325802

Makefile
codefromthecrypt: I think the cache concern is addressed now, since all matrices materialize in go even if externally seeded?

codefromthecrypt: PS on my laptop `-test.failfast` seems to invalidate cache, this is set by default in make so that CI runs don't proceed forever.. we can remove that with a comment.

codefromthecrypt: but to answer your question, yeah it seems that go caches the name, not the matrix. So, the above (which no longer exists) wouldn't have cached properly!

codefromthecrypt: added notes about the cache behavior thanks, as I didn't know this! https://github.com/golang/go/issues/42544#issuecomment-726204341

test/e2e/main_test.go
```diff
@@ -41,29 +43,74 @@ var (
...
+
 	os.Exit(m.Run())
 }
 
+// getExtensionLanguage reads E2E_EXTENSION_LANGUAGE or defaults to extension.LanguageTinyGo because it builds an order
+// of magnitude faster extension.LanguageRust. All languages test when "E2E_EXTENSION_LANGUAGE=all".
+func getExtensionLanguages() ([]extension.Language, error) {
+	fromEnv := os.Getenv(E2E_EXTENSION_LANGUAGE)
+	if fromEnv == "all" {
+		return extension.Languages, nil
```
mathetake, line 103: nice

mathetake (approved):

codefromthecrypt:
thanks for the help @mathetake I like this after your review more than before it!

PR #157, opened 2021-04-01
Fixes typos in ci scripts until we can remove them

codefromthecrypt:
#153 removes these, but this tidies up meanwhile


mathetake (approved):

codefromthecrypt:
I also renamed the to-be-deleted ci scripts directory to match $GOOS (darwin, not macos) as this being different has bit me a couple times so far

PR #158, opened 2021-04-02
Propagates runtime.GOOS as docker env GETENVOY_GOOS to speed up rust

codefromthecrypt:
The current way we build extensions is to mount a volume containing a directory created by "getenvoy extension init" in a "docker run" command. This implies a "bind mount" which is slow on macOS and particularly slow in rust (cargo) builds. This topic has locked our project into an old docker script, and even then it is quite slow.

This change tries to work around the problem by sharing information about the docker host (eg the machine calling `getenvoy`). What it does is propagates runtime.GOOS as docker env GETENVOY_GOOS. The rust builder image looks at this and optimizes if it is macOS.

While there are other optimizations we should address in #152, this should make enough difference to allow us to upgrade docker.

### DIFF
--- a/images/extension-builders/entrypoint.sh
+++ b/images/extension-builders/entrypoint.sh
@@ -14,25 +14,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -e
+set -ue
 
 SCRIPT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}")"  && pwd)
-GETENVOY_WORKSPACE_DIR="${GETENVOY_WORKSPACE_DIR:-$PWD}"
 
-USAGE="usage: build [--output-file PATH]
+# Note: we are using option syntax for --output-file even if this was always required. This is for compatibility with
+# older versions of the getenvoy binary.
+USAGE="usage: build --output-file path/to/extension.wasm
    or: test
    or: clean
 
 examples:
-   # build Wasm extension (location of *.wasm file is undefined)
-   build
-
    # build Wasm extension and copy *.wasm file to a given location
    build --output-file target/extension.wasm
-
-options:
-   build:
-   --output-file PATH   Path relative to the workspace root to copy *.wasm file to
 "
 
 usage() {
--- a/images/extension-builders/rust/commands.sh
+++ b/images/extension-builders/rust/commands.sh
@@ -13,31 +13,75 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+set -ue
 
-# Ensure location of the build directory.
-export CARGO_TARGET_DIR="${CARGO_TARGET_DIR:-${GETENVOY_WORKSPACE_DIR}/target}"
+export CARGO_TARGET="wasm32-unknown-unknown"
 
-# Keep Cargo cache inside the build directory, unless a user explicitly
-# overrides CARGO_HOME.
-export CARGO_HOME="${CARGO_HOME:-${CARGO_TARGET_DIR}/.cache/getenvoy/extension/rust-builder/cargo}"
+# Ensure location of the build cache directory.
+# See https://doc.rust-lang.org/cargo/guide/build-cache.html
+export CARGO_TARGET_DIR=${CARGO_TARGET_DIR:-${PWD}/target}
+
+# Keep Cargo cache inside the build directory, unless a user explicitly overrides CARGO_HOME.
+# See https://doc.rust-lang.org/cargo/guide/cargo-home.html
+export CARGO_HOME=${CARGO_HOME:-${CARGO_TARGET_DIR}/.cache/getenvoy/extension/rust-builder/cargo}
+
+# This is used in macOS, where we build in temporary directories. We only cache minimum as copying is slow.
+# See https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-c
+copy_cargo_home_cache() {
+	source_dir=${1}
+	dest_dir=${2}
+
+	# See if there is anything to copy
+	dirs=""
+	for dir in bin registry/index registry/cache git/db; do
+		test -d "${source_dir}/$dir" && dirs="$dirs $dir"
+	done
+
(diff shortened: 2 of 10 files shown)


images/extension-builders/entrypoint.sh
```diff
@@ -14,25 +14,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -e
+set -ue
 
 SCRIPT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}")"  && pwd)
 GETENVOY_WORKSPACE_DIR="${GETENVOY_WORKSPACE_DIR:-$PWD}"
```
codefromthecrypt, line 20: this isn't overridden anywhere, it isn't used in tinygo, and I don't think it helps to alias PWD. Can we get rid of this?

codefromthecrypt: for example, since initial import the work dir (pwd) was fixed via this `"-w", "/source",` in toolchain.go

images/extension-builders/rust/commands.sh
```diff
@@ -53,15 +54,11 @@ help:  make sure Cargo workspace includes a library crate with name '${lib_name}
...
 
-	local destination_file="$1"
+	local destination_file="${GETENVOY_WORKSPACE_DIR}/$1"
 	if [[ -n "${destination_file}" ]]; then
 		log_message "     Copying *.wasm file to '${destination_file}'"
-
-		destination_file="${GETENVOY_WORKSPACE_DIR}/${destination_file}"
-		local tmp_file="${destination_file}.tmp"
-		mkdir -p "$(dirname "${tmp_file}")"
-		cp "${cargo_output_file}" "${tmp_file}"
```
codefromthecrypt, line 63: if two-step copy had a reason to exist, there are no comments explaining why. I'm removing this, but it can be added back with a comment why regardless of if I'm around or not.

images/extension-builders/rust/commands.sh
```diff
@@ -53,15 +54,11 @@ help:  make sure Cargo workspace includes a library crate with name '${lib_name}
 "
 	fi
 
-	local destination_file="$1"
+	local destination_file="${GETENVOY_WORKSPACE_DIR}/$1"
 	if [[ -n "${destination_file}" ]]; then
```
codefromthecrypt, line 58: this condition isn't needed.. I'll remove in a later commit on this PR

codefromthecrypt:
Last successful run before this on rust+macOS was 37m, now it is 28m. This is not good enough, but it might be good enough to get recent docker to succeed..

codefromthecrypt:
Even if future things optimize this further, I think we should merge it. Using `E2E_TOOLCHAIN_CONTAINER_OPTIONS="-v /tmp/cargohome:/tmp/cargohome -e CARGO_HOME=/tmp/cargohome"` on my laptop, the results are staggering for `E2E_EXTENSION_LANGUAGE=rust`

Before, the elapsed time of `make e2e` was 19m37s and now 10m49s

mathetake:
do we need GOOS? I think `uname` command is enough to infer whether the host is dawrin or not

mathetake:
disclaimer: I haven't looked into the detail yet

codefromthecrypt:
> do we need GOOS? I think uname command is enough to infer whether the host is dawrin or not
this is the interesting thing. we need to know the host executing docker, not the host OS of the builder image. In that case, GOOS will resolve correctly. The builder images are always linux

mathetake:
yeah right

mathetake:
nm

codefromthecrypt:
fyi this improves things, especially local (ex on my laptop). However, it doesn't improve enough to remove the >2year old version of OS/x (e2e take longer than 45m still). I left #159 pinned to old docker, even if using this branch to make that faster.

images/extension-builders/entrypoint.sh
```diff
@@ -79,6 +79,11 @@ command_build()  {
 		shift
 	done
 
+	# Since code import in #98, this was always required.
+	if [ "${output_file}" == "" ]; then
```
mathetake, line 83: fyi we would never reach this branch as long as we use the container through `getenvoy extension init` https://github.com/tetratelabs/getenvoy/blob/master/pkg/extension/workspace/toolchain/builtin/default_config.go#L69-L77

mathetake: *build

images/extension-builders/entrypoint.sh
mathetake: looks like we can get rid of.

mathetake:
I tried manually but I got the error..
```
/Users/mathetake/getenvoy/build/bin/darwin/amd64/getenvoy extension build
error: could not find `Cargo.toml` in `/source` or any parent directory
Error: failed to build Envoy extension using "default" toolchain: failed to execute an external command "/usr/local/bin/docker run -u 501:20 --rm -e GETENVOY_GOOS=darwin -t -v /Users/mathetake/mytest:/source -w /source --init getenvoy/extension-rust-builder:latest build --output-file target/getenvoy/extension.wasm": exit status 101
```

images/extension-builders/entrypoint.sh
codefromthecrypt: yeah I was wondering how much we should support workspaces init'ed in the past... if not an issue, we should delete this I agree. wdyt?

mathetake:
well, this maybe irrelevant to this PR, same happens to the rc1.. let me take a look.

mathetake:
verified that actually caching works expectedly:) waiting for the test finishes.

images/extension-builders/entrypoint.sh
mathetake: not an issue, so let's delete 😄

codefromthecrypt: sure

mathetake (approved):

codefromthecrypt:
here's before and after

ok  	github.com/tetratelabs/getenvoy/test/e2e	1344.519s

real	22m28.594s
user	0m27.534s
sys	0m39.393s



ok  	github.com/tetratelabs/getenvoy/test/e2e	634.428s

real	10m37.880s
user	0m23.826s
sys	0m35.207s

PR #159, opened 2021-04-02
Migrates to GH action matrix for E2E runs, preferring actions-setup-docker

### LINKED ISSUE #145: macOS+rust e2e tests in CI are pinned to an over two year old version of Docker

### LINKED ISSUE #153: Delete ci/e2e/*/run_tests.sh scripts

codefromthecrypt:
This moves special-casing to rust+macOS cell in a normal build matrix.
Issue #152 still needs investigation on how to get around the slowness.

Fixes #153


codefromthecrypt:
re-kicking build as the rust+macOS-latest matrix died

Screen Shot 2021-04-02 at 11 36 09 AM

codefromthecrypt:
rebased on #158

codefromthecrypt:
This shows the special-casing where >2year old docker is only used for the rust+macOS cell in the matrix

Screen Shot 2021-04-03 at 8 45 21 AM
Screen Shot 2021-04-03 at 8 45 14 AM

codefromthecrypt:
rebased back to master to decouple this from other various PRs

codefromthecrypt:
FYI: this all passes, the only reason it isn't showing that is because via the matrix, the action names drifted. So, it is expecting feedback from the old names. This is expected

mathetake:
realized that TinyGo also experiences the increase of e2e running time, but this is vs running on the latest macos docker instead of 2yo one. Thanks for the work, I'm going to merge this.

mathetake:
Workflow name `Run end-to-end (e2e) tests (macos-latest, rust, -v ${RUNNER_TEMP}:/tmp/cargohome -e CARGO_HOME=/t...` looks confusing, so I'm going to try `-v ${RUNNER_TEMP}:/tmp/cargohome -e CARGO_HOME=/t...` both for TinyGo and Rust so that it clears job names

mathetake:
changed the job name according to https://futurestud.io/tutorials/github-actions-customize-the-job-name

mathetake:
:raised_hands: 
Screenshot from 2021-04-05 09-48-35

mathetake (approved):

codefromthecrypt:
good idea on rename and thanks for sharing!

PR #161, opened 2021-04-04
Migrates cmd tests from ginkgo,gomega to testing,testify

codefromthecrypt:
This completes migration of ./pkg/cmd/... from ginkgo,gomega to
testing,testify per rationale noted in #130.

This also fixes a bug where we accidentally added ignore files into
the output directory of `getenvoy extension init` and backfills tests
to ensure that doesn't happen again.

### elaboration on .gitignore and `getenvoy extension init`
Before, we added `.gitignore` files on `getenvoy extension init`. These
assume the end-user would be checking the workspaces into git. However,
the workspaces were neither initialized for git, nor the `.gitignore`
files complete per-language. Moreover, certain directories are subject
to variables, such as `CARGO_TARGET_DIR`.

This removes the .gitignore file generation completely, and with it the
duplication involved. If an end user is using our project, able to do
`git init`, yet want us to manage `.gitignore` also, we can revisit upon
such feedback later.


test/e2e/main_test.go
```diff
@@ -250,8 +210,8 @@ func requireExtensionClean(t *testing.T, workDir string) {
 	err := os.Chdir(workDir)
 	require.NoError(t, err, `error changing to directory: %v`, workDir)
 
-	cmd := getEnvoy("extension clean")
```
codefromthecrypt, line 253: these cmd -> c changes were mainly to help avoid clashing with the package

pkg/test/morerequire/morerequire.go
```diff
@@ -0,0 +1,108 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package morerequire includes more require functions than "github.com/stretchr/testify/require"
+// Do not add dependencies on any main code as it will cause cycles.
+package morerequire
```
codefromthecrypt, line 17: the stuff here is the same as it was, this is just pulled out to help reduce chance of package cycles

pkg/extension/init/init.go
```diff
@@ -102,10 +103,13 @@ func (s *scaffolder) walk(sourceDirName, destinationDirName string) (errs error)
 
 func (s *scaffolder) visit(sourceDirName, destinationDirName string, sourceFileInfo os.FileInfo) (errs error) {
 	baseOutputFileName := sourceFileInfo.Name()
-	// We rename go.mod to go.mod_ to workaround https://github.com/golang/go/issues/45197
-	if baseOutputFileName == "go.mod_" {
-		baseOutputFileName = "go.mod"
+	switch baseOutputFileName {
+	case ".gitignore", ".licenserignore":
```
mathetake, line 107: I think `.gitignore` should be there, so that users will not git add the generated cache/binary into VC if they git inited on the extension directory. For example, TinyGo generates build caches and binary under `build/` dir :https://github.com/tetratelabs/getenvoy/blob/master/data/extension/init/templates/tinygo/envoy.filters.http/default/.gitignore

mathetake: this is not only for us, but also for users

test/e2e/main_test.go
mathetake: +1

pkg/extension/init/init.go
codefromthecrypt: I didn't notice any end user requests for this, and the ignore files aren't correct either (ex missing lock file at least on the rust project). I'd prefer we remove .gitignore and wait for a user to complain. If they complain, we can add them back, but we should if doing so consider if we shouldn't be `git init`'ing the workspace or otherwise making it git relevant too, sg?

mathetake: sounds good! Thanks.

codefromthecrypt: updated the description comment to this effect!

pkg/cmd/extension/example/cmd_add_test.go
```diff
@@ -15,243 +15,176 @@
...
+	expectedStderr := fmt.Sprintf("Error: %s\n\nRun 'getenvoy extension examples add --help' for usage.\n", expectedErr)
+	require.Equal(t, expectedStderr, stderr.String(), `unexpected stderr running [%v]`, c)
+}
+
+func TestGetEnvoyExtensionExamplesAdd(t *testing.T) {
+	type testCase struct {
+		name                   string
+		templateWorkspace      string
+		flags                  []string
+		flagValues             []string
```
mathetake, line 103: can these be grouped together as `args []string` like others? 😄

codefromthecrypt: sure I'll make a sweep on this. I meant to convert them all..

pkg/cmd/root_test.go
```diff
@@ -15,220 +15,187 @@
...
+			expected: "http://manifest/arg",
+		},
+	}
+
+	for _, test := range tests {
+		test := test // pin! see https://github.com/kyoh86/scopelint for why
+
+		t.Run(test.name, func(t *testing.T) {
+			tearDown := test.setup()
+			defer tearDown()
```
mathetake, line 189: Shall we check if setup func is not nil, then we could remove `emptySetup`?
```suggestion
		if test.setup != nil {
			tearDown := test.setup()
			defer tearDown()
		}
```

pkg/cmd/extension/init/cmd_test.go
```diff
@@ -15,191 +15,201 @@
...
-				Expect(descriptor.Runtime.Envoy.Version).To(Equal("standard:1.17.0"))
-
-				By("verifying command output")
-				Expect(stdout.String()).To(BeEmpty())
-				Expect(stderr.String()).NotTo(BeEmpty())
-			},
-			func() []TableEntry {
-				entries := []TableEntry{}
-				for _, category := range []string{"envoy.filters.http", "envoy.filters.network", "envoy.access_loggers"} {
-					for _, language := range []string{"rust"} {
```
mathetake, line 194: oops, I didn't know that this is only tested for Rust, thanks

pkg/cmd/extension/init/params_test.go
```diff
@@ -16,89 +16,81 @@ package init
...
-			}()
+	cwd, err := os.Getwd()
+	require.NoError(t, err, "error getting current working directory")
 
-			By("verifying path under a file")
-			invalidPath := filepath.Join(tmpFile.Name(), "new_dir")
-			err = newParams().OutputDir.Validator(invalidPath)
-			Expect(err).To(MatchError(fmt.Sprintf(`stat %s: not a directory`, invalidPath)))
-		})
+	file := os.Args[0]
```
mathetake, line 38: neat!

mathetake (approved):
super!

codefromthecrypt:
thanks! will address the feedback shortly

pkg/cmd/root_test.go
codefromthecrypt: good! I didn't commit this as it drifted your intent, but thanks!

codefromthecrypt:
Thanks for another epic review @mathetake

PR #163, opened 2021-04-06
Migrates exec tests to testing/testify and fixes exit condition

codefromthecrypt:
Before, we used an incorrect exit condition to determine if a process
still exists. Now, we look up the PID.

This converts test code to testing/testify per rationale in #130.
This upgrades testify to support error conditions.


pkg/util/exec/exec_test.go
```diff
@@ -19,138 +19,148 @@ package exec
...
+		close(errCh)
+	}()
+
+	// Wait for shell script to start
+	require.Eventually(t, func() bool {
+		return strings.Contains(stdout.String(), "running")
+	}, 1*time.Second, 10*time.Millisecond)
+
+	// Send a fake signal to our signal handler which invokes the stop channel
+	fakeSignal := syscall.Signal(0)
```
codefromthecrypt, line 147: this was super confusing to me, and finally realized that our signals would never get to the process, hence changing the signal and adding a comment

pkg/util/exec/exec_test.go
```diff
@@ -19,138 +19,148 @@ package exec
...
+	stopCh <- fakeSignal
+	close(stopCh)
+
+	// Wait for the err channel which means the process shutdown.
+	for err := range errCh {
+		require.Equal(t, commonerrors.NewShutdownError(fakeSignal), err)
+	}
+
+	// Verify the program received a SIGTERM because the process was still alive upon exec.terminate()
+	require.Equal(t, "SIGTERM caught!\n", stderr.String(), `invalid stderr running [%v]`, cmd)
```
codefromthecrypt, line 157: this is important to know because otherwise we don't know why we expect a kill to eventually occur.

pkg/util/exec/testdata/test_trap_sigterm.sh
```diff
@@ -14,8 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-printf >&1 '%s' "running"
+echo running
```
codefromthecrypt, line 17: not sure why this was more complicated than echo before.. changed it because it feels less distracting.

pkg/util/exec/testdata/test_trap_sigterm.sh
```diff
@@ -14,8 +14,9 @@
...
 
-trap 'echo "SIGTERM caught!"' TERM
+trap 'echo >&2 "SIGINT caught!"' INT
```
codefromthecrypt, line 19:
```suggestion
```

codefromthecrypt (commented):
notes

pkg/util/exec/exec.go
```diff
@@ -117,25 +118,17 @@ func Run(cmd *exec.Cmd, streams ioutil.StdStreams) error {
 }
 
 func terminate(cmd *exec.Cmd, errCh <-chan error) {
```
mathetake, line 120: I guess we don't need to pass errCh to this function anymore.

pkg/util/exec/exec.go
```diff
@@ -151,3 +144,13 @@ func terminate(cmd *exec.Cmd, errCh <-chan error) {
 		log.Warnf("external process didn't exit gracefully within %s: %q", killTimeout, cmd)
 	}
```
mathetake, line 145: remove this switch and simply have 
```
<-time.After(killTimeout)
if isProcessRunning(cmd.Process.Pid) {
	if e := cmd.Process.Kill(); e != nil {
		log.Warnf("failed to send SIGKILL to the external process %q: %v", cmd, e)
	}
}
_ = cmd.Process.Release()
```

then we could remove the defer above.

mathetake: per https://github.com/tetratelabs/getenvoy/pull/163/files#r607519680

pkg/util/exec/exec_test.go
mathetake: yeah it took me a few minutes to understand this..

mathetake (approved):
LGTM except the comment above

pkg/util/exec/exec.go
codefromthecrypt: if we do that we need to override killTimeout otherwise the tests will hang. NP will do

codefromthecrypt: done

pkg/util/exec/exec.go
codefromthecrypt: yeah thanks for the tips!

codefromthecrypt:
I might revert the part that reverts the stop signal as it is making the circleci flakey..

Issue #164, opened 2021-04-06
Add git to the TinyGo builder image


nacx:
When building TinyGo extensions, it is common that Go module dependencies are downloaded using `git`, but the binary is not available in the builder image. It would be great to have it there so that extensions that use custom dependencies can be nicely built.

By default, Go modules will attempt to use HTTP to fetch dependencies, but that will help when downloading modules that are not public, as it will prompt for credentials and will fail on that interactive process. It would be nice, in addition, to preconfigure Git in the builder image to use SSH by default with something like:
```bash
git config --global url.ssh://git@github.com/.insteadOf https://github.com/
```
This would allow users to build custom extensions even if they live in private git repositories, by mounting the local `.ssh` folders in the build image. This, however, cannot be easily done due to how [the current user is being resolved](https://github.com/tetratelabs/getenvoy/blob/master/pkg/extension/workspace/toolchain/builtin/toolchain.go#L93) in the build process. Some local tests showed that the build was being run with this command:
```bash
$ getenvoy extension build --toolchain-container-options '-v /Users/nacx/.ssh:/root/.ssh'
(...)
/usr/local/bin/docker run -u 501:20 --rm -t -v /opt/tetrate/ngac-envoy-wasm:/source -w /source --init -v /Users/nacx/.ssh:/root/.ssh getenvoy/extension-tinygo-builder:0.2.1-rc2 build --output-file build/extension.wasm
```
Note that the UID 501 is the UID of my user in the host, but does not exist as a user in the container image.

In order to allow properly preconfiguring Git and allowing to mount stuff for the builder user, the base image should probably create a custom user (say `builder`) and configure it to be the default user in that container. This way we can have a predictable UID/GID for the container user and allow end-users to better customize the build process.

codefromthecrypt:
we should probably pre-fetch common dependencies and make sure that works also, as the `getenvoy extension` commands slow by re-downloading github.com/stretchr/testify@v1.6.1 and github.com/tetratelabs/proxy-wasm-go-sdk@v0.1.1 on each run

codefromthecrypt:
obviated by #200

PR #166, opened 2021-04-06
Execute chmod for TinyGo gocache on trap

mathetake:
Signed-off-by: Takeshi Yoneda <takeshi@tetrate.io>

Previously, when `go {test,build}` fails, chmod would never be called. That would cause the nasty permission denied error when test/build failures occur during e2e as well as users try to remove an extension dir after failure.


codefromthecrypt (approved):
What if the build directory was baked into the image? We should consider more in addressing #164, but this is good for now. thanks!

codefromthecrypt:
merging because we know that (macos-latest, rust) fails as it is still pinned to a flakey install script discussed in #145

Issue #113, opened 2021-02-04
Bintray is going away


dkrutsko:
According to [this blog post](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter), Bintray is being retired May 1st. I believe this project will be affected as some of your artifacts, like RPM, are being stored there.

codefromthecrypt:
could this be handled by whatever is hosting https://getenvoy.io?

However redirects and forwarding works, we could push the manifest.json on release to a gh-pages branch, which could be cached or served one way or another from a getenvoy.io path?

codefromthecrypt:
@lizan thoughts on this?

DerekStrickland:
FWIW, bintray is undergoing a scheduled brownout right now, and thus the getenvoy CLI is failing.

codefromthecrypt:
#170 will fix this on next release.

There is a hidden option that folks can use in a pinch meanwhile, but disclaimer this flag is internal and so don't rely on it being around: `getenvoy --manifest https://dl.getenvoy.io/public/raw/files/manifest.json fetch standard:1.17.1`

oktocat:
it's a bit premature to close it, there's still work that needs to be done with CI and packaging.

codefromthecrypt:
ok can you link issues to that so we know when to close it? (possibly with check boxes). As I understand it, the other steps won't effect this repository, but if any do, let's make that visible somehow.

oktocat:
right, other issues will affect different repos like this one: https://github.com/tetratelabs/getenvoy-package/pull/105

codefromthecrypt:
The current getenvoy CLI doesn't reference anything in bintray anymore, and envoy system packages will eventually move upstream. Folks who want to know about system packages can follow tetratelabs/getenvoy-package#110

The current CLI release was built with a simpler JSON format. The latest version is [v0.3.0](https://github.com/tetratelabs/getenvoy/releases/tag/v0.3.0) though as it is a big release probably a v0.3.1 will happen soon with feedback.

The canonical version list is here https://raw.githubusercontent.com/tetratelabs/getenvoy/master/site/envoy-versions.json
but directly referenced via redirect in the default of `ENVOY_VERSIONS_URL=https://getenvoy.io/envoy-versions.json`

This file will have latest patch of non EOL releases defined by upstream, and anything older is exception basis https://github.com/envoyproxy/envoy/blob/main/RELEASES.md#release-schedule

[...]

PR #170, opened 2021-04-09
Migrate from bintray to cloudsmith

arkodg:
Jfrog is sunsetting bintray on May 1st, 2021
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

Use Cloudsmith (https://cloudsmith.io/) instead to host binaries

Signed-off-by: Arko Dasgupta <arko@tetrate.io>


pkg/cmd/root_test.go
```diff
@@ -149,9 +149,9 @@ func TestGetEnvoyManifest(t *testing.T) {
 
 	tests := []testCase{ // we don't test default as that depends on the runtime env
 		{
-			name:     "default is https://tetrate.bintray.com/getenvoy/manifest.json",
+			name:     "default is https://dl.cloudsmith.io/public/tetrate/getenvoy/raw/files/manifest.json",
```
codefromthecrypt, line 152: curious.. wouldn't this be a good time to use DNS we control instead of having to wonder about a similar future of another SAAS? If this is just manifest json, even GH pages branch might be ok...

arkodg: yah good point, @oktocat  will be setting up so we can use `dl.getenvoy.io` here, just surfacing all the changes needed for the migration for now and stashing these changes in draft mode

codefromthecrypt:
fyi here is the tracking issue this would resolve #113

codefromthecrypt:
how's this going?

arkodg:
👋  updated the PR with the new URLs
will probably need to rerun the CI and merge once the packages have synced from bintray to cloudsmith

codefromthecrypt:
Thanks for the update. By re-run CI, you mean the tests here? or a release or..

arkodg:
👍  was referring to this PR workflow

oktocat (approved):

codefromthecrypt (approved):
great. I'm sure folks will be happy to have a world with less 503

PR #171, opened 2021-04-10
Fixes docker.io normal case rate-limiting

### LINKED ISSUE #149: Adjust for docker.io rate limits
**Describe the bug**
Builds can fail due to docker.io rate limits like so:
```
Unable to find image 'busybox:latest' locally
docker: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.
See 'docker run --help'.
```

**To Reproduce**
do a bunch of builds, then look at the logs

**Expected behavior**
builds run

**Additional context**
This is caused by using base images hosted by docker.io we can use a different registry for base images to avoid this.

codefromthecrypt:
Each pull request implies many docker.io requests, this has led to
builds failing on account of rate limiting. This avoids that by using
the GitHub Container Registry (ghcr.io) instead of Docker Hub
(docker.io). ghcr.io is not rate-limited. Specifically, we use internal
copies of images used as parents or in testing. Since these aren't what
end users use anyway, this doesn't impact them.

At the end of the day, ghcr.io/tetratelabs/getenvoy-internal has tags
corresponding to our needs:

```yaml
          - source: busybox:1.32.1  # test image: ci/e2e/darwin/install_docker.sh
            target_tag: busybox
          - source: registry:2  # test image: docker-compose.yml
            target_tag: registry
          - source: rust:1.51.0  # parent image: images/extension-builders/rust/Dockerfile
            target_tag: rust
          - source: tinygo/tinygo:0.17.0  # parent image: images/extension-builders/tinygo/Dockerfile
            target_tag: tinygo
```

Updating our base layers or otherwise implies a change to
`.github/workflows/internal-images.yml` and re-triggering the workflow
or waiting until the next day for automatic publishing to occur.

Another change in this PR is removing the docker dependency from unit
tests via a mock. Even if we have safe images, we still shouldn't
subject unit tests to flakiness, extra time, or network dependencies
unless there is no other way.

Fixes #149

### DIFF
--- a/.github/workflows/internal-images.yml
+++ b/.github/workflows/internal-images.yml
@@ -1,0 +1,46 @@
+# yamllint --format github .github/workflows/internal-images.yml
+---
+name: internal-images
+
+# Refresh the tags once a day
+on:
+  schedule:
+    - cron: "23 3 * * *"
+  workflow_dispatch:  # Allows manual refresh
+
+# This copies images from docker.io to ghcr.io/tetratelabs/getenvoy-internal:$tag
+# Using these in tests and as a parent (FROM) avoids docker.io rate-limits particularly on pull requests.
+jobs:
+  copy-images:
+    strategy:
+      matrix:
+        # Be precise in tag versions to improve reproducibility
+        include:
+          - source: busybox:1.32.1  # test image: ci/e2e/darwin/install_docker.sh
+            target_tag: busybox
+          - source: registry:2  # test image: docker-compose.yml
+            target_tag: registry
+          - source: rust:1.51.0  # parent image: images/extension-builders/rust/Dockerfile
+            target_tag: rust
+          - source: tinygo/tinygo:0.17.0  # parent image: images/extension-builders/tinygo/Dockerfile
+            target_tag: tinygo
+    runs-on: ubuntu-latest
+    steps:
+      # Same as doing this locally: echo "${GHCR_TOKEN}" | docker login ghcr.io -u "${GHCR_TOKEN}" --password-stdin
+      - name: Login into GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          # GHCR_TOKEN=<hex token value>
+          #   - pushes Docker images to ghcr.io
+          #   - create via https://github.com/settings/tokens
+          #   - assign via https://github.com/organizations/tetratelabs/settings/secrets/actions
+          #   - needs repo:status, public_repo, write:packages, delete:packages
+          password: ${{ secrets.GHCR_TOKEN }}
+
+      - name: Pull and push
+        run: |  # This will only push a single architecture, which is fine as we currently only support amd64
+          docker pull ${{ matrix.source }}
+          docker tag ${{ matrix.source }} ghcr.io/tetratelabs/getenvoy-internal:${{ matrix.target_tag }}
+          docker push ghcr.io/tetratelabs/getenvoy-internal:${{ matrix.target_tag }}
--- a/pkg/cmd/extension/push/cmd_test.go
+++ b/pkg/cmd/extension/push/cmd_test.go
@@ -40,7 +40,7 @@ const defaultTag = "latest"
 
 // TestGetEnvoyExtensionPush shows current directory is usable, provided it is a valid workspace.
 func TestGetEnvoyExtensionPush(t *testing.T) {
-	mock := mockRegistryServer()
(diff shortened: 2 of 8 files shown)


.github/workflows/internal-images.yml
```diff
@@ -0,0 +1,46 @@
...
+# Refresh the tags once a day
+on:
+  schedule:
+    - cron: "23 3 * * *"
+  workflow_dispatch:  # Allows manual refresh
+
+# This copies images from docker.io to ghcr.io/tetratelabs/getenvoy-internal:$tag
+# Using these in tests and as a parent (FROM) avoids docker.io rate-limits particularly on pull requests.
+jobs:
+  copy-images:
```
codefromthecrypt, line 14: mirroring approach borrows heavily from https://github.com/open-telemetry/opentelemetry-java/blob/main/.github/workflows/docker-test-containers.yml

.github/workflows/internal-images.yml
```diff
@@ -0,0 +1,46 @@
...
+        # Be precise in tag versions to improve reproducibility
+        include:
+          - source: busybox:1.32.1  # test image: ci/e2e/darwin/install_docker.sh
+            target_tag: busybox
+          - source: registry:2  # test image: docker-compose.yml
+            target_tag: registry
+          - source: rust:1.51.0  # parent image: images/extension-builders/rust/Dockerfile
+            target_tag: rust
+          - source: tinygo/tinygo:0.17.0  # parent image: images/extension-builders/tinygo/Dockerfile
+            target_tag: tinygo
```
mathetake, line 26: I would prefer to include the version info in the docker hub. What if the container updated while building the release containers? That would change the assumption that the container built at release job would pass all the unittest.. WDYT?

```suggestion
        include:
          - source: busybox:1.32.1  # test image: ci/e2e/darwin/install_docker.sh
            target_tag: busybox-1.32.1
          - source: registry:2  # test image: docker-compose.yml
            target_tag: registry-2
          - source: rust:1.51.0  # parent image: images/extension-builders/rust/Dockerfile
            target_tag: rust-1.15.0
          - source: tinygo/tinygo:0.17.0  # parent image: images/extension-builders/tinygo/Dockerfile
            target_tag: tinygo-0.17.0
```

pkg/cmd/extension/push/cmd_test.go
```diff
@@ -97,3 +116,42 @@ Pushed %s:latest
...
+		case "PUT":
+			if r.URL.Path == "/upload" {
+				statusCode = 200 // Pretend we accepted the blob
+				_ = r.ParseForm()
+				w.Header().Add("Docker-Content-Digest", r.Form.Get("digest"))
+			} else if r.URL.Path == "/v2/getenvoy/sample/manifests/latest" {
+				statusCode = 200 // Pretend we accepted the manifest
+				if raw, err := ioutil.ReadAll(r.Body); err == nil {
+					w.Header().Add("Docker-Content-Digest", "sha256:"+hash(raw))
+				}
```
mathetake, line 145: nit looks like you forgot to handle `err != nil` 😄

mathetake (approved):

.github/workflows/internal-images.yml
codefromthecrypt: So, this topic effects the builder images as the test ones don't matter.

Changes to this file apply on the cron, which is 3:23 am UTC. So, if someone updated, it wouldn't happen immediately. The risk would be is the release is going on and passes 3:23am UTC where this file was updated. I think this is a low risk because releases don't occur that frequently.

If we solved like this, we would introduce a double-maintenance problem of fixing the image version here, waiting for it to be published, then changing in the corresponding files. Almost certainly, we would end up with drift because someone will forget to update twice.

Meanwhile, we currently only push "latest" tags. What this means is even if we prevent simultaneous updates, the next time the builder image is pushed, any version will get that whether they like it or not.

We can eventually pin releases to a docker tag, but while we don't I think the approach here doesn't make as much sense as it seems to at first glance, so leaving it alone is probably best. wdyt?

pkg/cmd/extension/push/cmd_test.go
codefromthecrypt: yeah I actually considered it and got lazy. I think I will make error nil a 500. good point!

codefromthecrypt:
thanks for the help @mathetake

codefromthecrypt:
PS this worked https://github.com/tetratelabs/getenvoy/actions/runs/739756509

Issue #169, opened 2021-04-09
Clarify version support policy


codefromthecrypt:
Right now, versions supported here are opaque, and usually coincide with the first time something was touched. This isn't good long term as the burden of pinning old libraries is too high.

For example, https://github.com/tetratelabs/getenvoy/pull/168 only brings us up to Istio 1.6 due to lack of understanding of the policy. We could also argue that the feature may not even be worth the struggle as it took hours just to get this much working, and libraries change dramatically again from 1.6 to 1.7, meanwhile this thrash steals a lot of attention from Envoy itself.

To help the project last long term, we should make a policy about Envoy and Istio versions supported. While Istio may pin us to older Envoy, I suspect we shouldn't support versions of Istio that the project themselves no longer support.

What would make sense here? I would like to put it on the README and change the project accordingly.

codefromthecrypt:
I think istio 1.7 will end up a minimum as we can't afford the maintenance churn on 1.6. 

Particularly, the `--bootstrap istio` option to `getenvoy run` is the only thing that requires istio, and it is days of effort and still not working right. This is marked experimental, so I think it should be valid to update this whenever we like (or remove it)

@lizan Can we remove `--bootstrap istio` or set its minimum version to something at least 1.7? Note: this may also pin us to a floor envoy version, though I suspect you will know more than me.

lizan:
Yes. Istio 1.6 is already EOL so we can remove support of that.

codefromthecrypt:
#174 moves to Envoy 1.15 and Istio 1.7, but it seems like Istio 1.7 is also EOL! https://istio.io/latest/about/supported-releases

Should we move one more?

codefromthecrypt:
following the rabbit down the hole, if we went to istio 1.18, this would mean envoy 1.16.. https://github.com/istio/envoy/commit/5b0c5f7b21f84b6ab86e5f416bdaf6bb0fbc2a32

codefromthecrypt:
spiked this https://github.com/tetratelabs/getenvoy/pull/177

If we're ok with it, we can say our policy is to support what isn't EOL, with an exception that if there's a conflict between istio and envoy's floor version, we take the lower floor. SG?

lizan:
sounds good to me

codefromthecrypt:
murphy's law.. 1.16 doesn't work. so we can try the floor of both when that means envoy 1.17 :P https://github.com/tetratelabs/getenvoy/pull/177

[...]

PR #174, opened 2021-04-12
Updates to Envoy 1.15 and Istio 1.7

codefromthecrypt:
This migrates to floor versions implied in Istio 1.7

See #169


codefromthecrypt:
inlined the api from getenvoy-package per https://github.com/tetratelabs/getenvoy-package/pull/103

this allowed us to use latest protobuf libs

pkg/extension/workspace/config/helpers.go
```diff
@@ -15,12 +15,14 @@
 package config
 
 import (
-	"github.com/ghodss/yaml"
```
codefromthecrypt, line 18: dead lib

pkg/extension/example/init/registry/types.go
```diff
@@ -15,16 +15,13 @@
...
+	"io/fs"
 	"io/ioutil"
-	"net/http"
-	"os"
 	"path"
 	"path/filepath"
 	"strings"
 
-	"github.com/pkg/errors"
-	"github.com/shurcooL/httpfs/vfsutil"
```
codefromthecrypt, line 26: dead lib

pkg/extension/workspace/example/envoy/util/load.go
```diff
@@ -30,32 +32,41 @@ import (
...
 // On the bright side, unmarshalling will never fail due to use of unknown types.
 type fakeAnyResolver struct{}
 
-func (r fakeAnyResolver) Resolve(typeURL string) (proto.Message, error) {
+func (r fakeAnyResolver) FindMessageByName(message protoreflect.FullName) (protoreflect.MessageType, error) {
+	return r.FindMessageByURL(string(message))
+}
+
+func (r fakeAnyResolver) FindMessageByURL(url string) (protoreflect.MessageType, error) {
 	// NOTE: as of github.com/golang/protobuf/jsonpb@v1.3.x, the only way to achieve
```
codefromthecrypt, line 58: I left this comment alone intentionally even if it refers to an older version. I don't know if there's a way around this in the new version, but I also couldn't find one looking in an hour. This approach still works.

pkg/extension/workspace/example/envoy/util/load_test.go
```diff
@@ -68,28 +65,11 @@ func TestNewFakeAnyResolver(t *testing.T) {
...
 
 		t.Run(test.name, func(t *testing.T) {
-			resolver := newFakeAnyResolver()
+			resolver := fakeAnyResolver{}
 
-			actual, err := resolver.Resolve(test.typeURL)
+			actual, err := resolver.FindMessageByURL(test.typeURL)
 			require.NoError(t, err)
-
-			props := proto.GetProperties(reflect.ValueOf(actual).Elem().Type())
```
codefromthecrypt, line 76: this was library specific. the key here is that it doesn't err. other tests show that "any" values are consumable

pkg/manifest/print.go
```diff
@@ -54,19 +55,25 @@ func fetch(url string) (*api.Manifest, error) {
...
-		return nil, fmt.Errorf("error unmarshalling manifest: %v", err)
+	if err := protojson.Unmarshal(body, &result); err != nil {
+		return nil, fmt.Errorf("error unmarshalling manifest: %w", err)
 	}
 	return &result, nil
 }
 
 func deterministicFlavors(flavors map[string]*api.Flavor) []*api.Flavor {
-	flavorList := []*api.Flavor{}
+	flavorList := make([]*api.Flavor, 0, len(flavors))
```
codefromthecrypt, line 76: dodge lint

pkg/extension/workspace/example/runtime/configdir/configdir_test.go
```diff
@@ -32,50 +32,34 @@ import (
...
+			expectValidBootstrap: false,
 		},
 		{
-			name:         "envoy.tmpl.yaml: invalid paths to `lds` and `cds` files",
-			workspaceDir: "testdata/workspace4",
-			isEnvoyTemplate: func(name string) bool {
-				return name == "envoy.tmpl.yaml"
-			},
-			requireBootstrap: expectValidBootstrap,
+			name:                 "envoy.tmpl.yaml: invalid paths to `lds` and `cds` files",
```
codefromthecrypt, line 55: we accidentally had also included lds and cds files in the config directory despite the paths being incorrect. These were templated and made yaml parsing fail. The simpler out was to not include these files in the test about invalid paths to them!

api/manifest.proto
```diff
@@ -0,0 +1,72 @@
+// Copyright 2021 Tetrate
```
codefromthecrypt, line 1: I didn't import the source history of this file as there weren't many revisions and the last was nearly 2 years ago https://github.com/tetratelabs/getenvoy-package/commits/master/api/manifest.proto

pkg/extension/workspace/example/runtime/configdir/testdata/workspace3/expected/getenvoy_extension_run/envoy.tmpl.yaml
```diff
@@ -18,7 +18,7 @@ static_resources:
               typed_config:
                 "@type": type.googleapis.com/envoy.extensions.filters.network.wasm.v3.Wasm
                 config:
-                  configuration: {"@type":"type.googleapis.com/google.protobuf.StringValue","value":"{\"key\":\"value\"}\n"}
+                  configuration: {"@type":"type.googleapis.com/google.protobuf.StringValue", "value":"{\"key\":\"value\"}\n"}
```
codefromthecrypt, line 21: this normalizes to space after as marshaling seems to default this way now

codefromthecrypt:
cmd test is wrestling about space-after in json encoding.. locally even with the exact same golang version, make test is fine..

pkg/binary/envoy/config.go
```diff
@@ -54,9 +51,8 @@ func ParseMode(s string) Mode {
 func NewConfig(options ...func(*Config)) *Config {
 	cfg := &Config{
 		AdminPort:      15000,
-		StatNameLength: 189,
```
mathetake, line 57: I see `StatNameLength` is never used 😄 Thanks!

pkg/binary/envoy/debug/testdata/admin.yaml
```diff
@@ -22,7 +22,7 @@ static_resources:
         - filters:
             - name: envoy.http_connection_manager
               typed_config:
-                "@type": type.googleapis.com/envoy.config.filter.network.http_connection_manager.v2.HttpConnectionManager
+                "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
```
mathetake, line 25: +1

pkg/cmd/extension/run/cmd.go
```diff
@@ -66,14 +67,14 @@ func (opts *runOpts) validateExtension() error {
 	// pre-built *.wasm file
 	if opts.Extension.WasmFile != "" {
 		if err := osutil.IsRegularFile(opts.Extension.WasmFile); err != nil {
-			return errors.Wrapf(err, "unable to find a pre-built *.wasm file at %q", opts.Extension.WasmFile)
+			return fmt.Errorf("unable to find a pre-built *.wasm file at %q: %w", opts.Extension.WasmFile, err)
```
mathetake, line 70: +1

pkg/extension/workspace/config/helpers.go
```diff
@@ -15,12 +15,14 @@
...
+	"sigs.k8s.io/yaml"
```
mathetake, line 18: would it be possible to just use https://pkg.go.dev/gopkg.in/yaml.v3 on which k8s one is built? The smaller dep, the better 😄?

pkg/extension/workspace/example/envoy/template/expand_test.go
```diff
@@ -68,12 +68,12 @@ func TestExpand(t *testing.T) {
...
 			input:    `{{ .GetEnvoy.Extension.Config }}`,
-			expected: `{"@type":"type.googleapis.com/google.protobuf.StringValue","value":"{\"key\":\"value\"}"}`,
+			expected: `{"@type":"type.googleapis.com/google.protobuf.StringValue", "value":"{\"key\":\"value\"}"}`,
 		},
 		{
 			name:     "access to proto message",
-			input:    `{{ (.GetEnvoy.Extension.Code).Message.GetLocal.GetFilename }}`,
-			expected: `/path/to/extension.wasm`,
+			input:    `"{{ (.GetEnvoy.Extension.Code).Message.GetLocal.GetFilename }}"`,
+			expected: `"/path/to/extension.wasm"`,
```
mathetake, line 76: what do we need this change?

mathetake: why

pkg/extension/workspace/example/runtime/configdir/testdata/workspace4/.getenvoy/extension/examples/default/cds.yaml
```diff
@@ -1,17 +0,0 @@
-version_info: v0
```
mathetake, line 1: I guess we could also delete `pkg/extension/workspace/example/runtime/configdir/testdata/workspace4/expected/getenvoy_extension_run/lds.tmpl.yaml` and crd.yaml?

pkg/extension/workspace/config/helpers.go
codefromthecrypt: I tried, but we rely on behavior, and we also need "sigs.k8s.io/yaml" elsewhere for coercing yaml to json to struct. It is a bigger bite than it looks, though I agree desirable.

pkg/extension/workspace/example/envoy/template/expand_test.go
codefromthecrypt: makes testing easier as we can use json assertions on all the cells

codefromthecrypt: before, this cell was the only one that wasn't valid json itow

pkg/extension/workspace/example/runtime/configdir/testdata/workspace4/.getenvoy/extension/examples/default/cds.yaml
codefromthecrypt: will try and delete if possible. sounds good!

pkg/extension/workspace/config/helpers.go
codefromthecrypt: https://github.com/tetratelabs/getenvoy/issues/176

pkg/extension/workspace/config/helpers.go
```diff
@@ -20,9 +20,9 @@ import (
 
 var (
 	// Marshal converts an object into its YAML representation.
-	Marshal func(interface{}) ([]byte, error) = yaml.Marshal
```
codefromthecrypt, line 23: editor said this is redundant, and who am I to argue? ;)

pkg/extension/workspace/config/helpers.go
mathetake: OK, thanks for clarifying!

pkg/extension/workspace/config/helpers.go
```diff
@@ -15,12 +15,14 @@
...
+	Marshal = yaml.Marshal
 	// Unmarshal converts YAML into an object.
-	Unmarshal func([]byte, interface{}) error = yaml.Unmarshal
+	Unmarshal = func(bytes []byte, i interface{}) error {
+		return yaml.Unmarshal(bytes, i) // no opts
+	}
```
mathetake, line 27: should be simplified..
```suggestion
	Unmarshal = yaml.Unmarshal
```

codefromthecrypt: sorry this was something from the last review also. the comment // no opts is there because we can't shorten it because yaml.Unmarshal has `, opts ...JSONOpt` param...

somehow I lost this change in merging the prior which is why you are seeing it again. sorry about that

mathetake: well `make build` and `make test` pass even if just aliasing `Unmarshal = yaml.Unmarshal`.. so just curious what do you mean by "we can't"..?

pkg/extension/workspace/example/envoy/template/expand_test.go
mathetake: thanks, make sense!

pkg/extension/workspace/config/helpers.go
codefromthecrypt: so it has to do with what the point is of this wrapper.

if you alias like this, it won't compile: `Unmarshal = func(bytes []byte, i interface{}) yaml.Unmarshal`

if you alias without bounding out the options, it will, but then what is the point as you've now leaked the type of the option. If that's the case, why not use yaml.Unmarshal directly

codefromthecrypt: iotw I think if we are blindly aliasing, it is fine by me but we should just delete helpers.go in that case.. make sense?

codefromthecrypt: I'll go with deleting helpers.go as that will also help narrow the dependency on sigs.k8s.io/yaml at the same time

I don't think helpers.go deserves the attention it has gotten :D

mathetake: > if you alias without bounding out the options,

I see. Now I understand. I would just prefer deleting aliasing here..

mathetake: too few place where these aliases are used.. I think it's just complicating the things.
```
$ git grep -E "config.Unmarshal" | grep -v test
pkg/extension/workspace/model/example.go:       err := config.Unmarshal(file.Content, descriptor)
pkg/extension/workspace/model/workspace.go:     err = config.Unmarshal(data, descriptor)
pkg/extension/workspace/toolchain/builtin/factory.go:                   if err := config.Unmarshal(args.Toolchain.Config.Content, cfg); err != nil {
pkg/extension/workspace/toolchain/lookup.go:    err = config.Unmarshal(file.Content, meta)
mathetake@mathetake ~/getenvoy remotes/origin/envoy1.15-istio1.7*
$ git grep -E "config.Marshal" | grep -v test
// nothing!
```

codefromthecrypt: OK! and now only one file depends on sigs.k8s.io/yaml

mathetake (approved):

mathetake:
Thanks for all the work here! 🚀

codefromthecrypt:
thanks for another epic review.. we're gettin there!

PR #168, opened 2021-04-07
Upgrades Envoy 1.11 -> 1.12 and Istio 1.2 -> 1.6

codefromthecrypt:
This upgrades Envoy from 1.11 -> 1.12 and Istio from 1.2 -> 1.6 and fixes the MacOS build.

Since Envoy 1.12 [fixes /dev/stdout on OS/x](https://github.com/envoyproxy/envoy/pull/7998), `EnableEnvoyLogCollection` can now work on macOS. Upgrading further will be in another pull request.

This also fixes functionality related to lsof and debug. Before, we didn't use timeouts, which could result in hangs. Now, we wait up to 3 seconds to capture this information.

fixes #88

### General notes

This also fixes race conditions by reading back the admin port in tests via the "--admin-address-path" flag as implemented in `envoy.EnableAdminAddressDetection`

This intentionally stops at Istio 1.6 as versions beyond this imply configuration changes to our test data.

This should also fix #88 (macOS support)

### On go:embed for the istio envoy bootstrap template

Before, our unit tests would download a 30MB+ targz and copy it to verify that a json file from envoy was still the same contents. This replaces that code with `go:embed` so we don't need to test it (because there's no risk of code formatters or anything else interfering with it.

This removes a large amount of the unit test runtime and with it a test that isn't good to execute `-race` against!
```
$ go test ./pkg/binary/envoy/controlplane -run TestIstioBootStrapTemplateEqualsReleaseJson
ok  	github.com/tetratelabs/getenvoy/pkg/binary/envoy/controlplane	7.127s
```

PR #178, opened 2021-04-15
Removes Istio bootstrap

codefromthecrypt:
The maintenance and side-effects of Istio bootstrap integration have
proven to be the bulk of effort in maintaining this project, and the
bulk of its size. This removes the feature to free us to focus on Envoy,
which better aligns with the primary charter of the project.

Before:
```bash
$ du -k build/bin/*/amd64/getenvoy
77924	build/bin/darwin/amd64/getenvoy
66692	build/bin/linux/amd64/getenvoy
```

Now:
```bash
$ du -k build/bin/*/amd64/getenvoy
20768   build/bin/darwin/amd64/getenvoy
18428   build/bin/linux/amd64/getenvoy
```


pkg/cmd/extension/push/cmd.go
```diff
@@ -18,7 +18,7 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/docker/distribution/reference"
+	"github.com/containerd/containerd/reference/docker"
```
codefromthecrypt, line 21: we don't need to import the docker distribution to verify a reference. NOTE: we still import it indirectly due to oras, but I can manage that in a follow-up PR

pkg/extension/workspace/example/envoy/template/defaults.go
```diff
@@ -41,7 +41,7 @@ var (
 					SocketAddress: &envoycore.SocketAddress{
 						Address: "127.0.0.1",
 						PortSpecifier: &envoycore.SocketAddress_PortValue{
-							PortValue: 9901,
+							PortValue: 9901, // TODO: why default admin port to 9901?
```
codefromthecrypt, line 44: no idea on this one..

pkg/pii/pii.go
```diff
@@ -121,12 +120,6 @@ func (f Filter) Process(logs []string) []string {
 	return out
 }
 
-// Process process logs with the default filter
-// an empty array of logs and an error instance is returned in the event of an error
-func Process(logs []string) ([]string, error) {
```
codefromthecrypt, line 126: dead code

codefromthecrypt (commented):
notes

mathetake:
Others: the rational is not only the maintenance burden, but also the fact that we have only supported 1.2 (https://github.com/tetratelabs/getenvoy/pull/168) in the released getenvoy version.. which itself I think proves that this feature is no longer in use by the end user.

codefromthecrypt:
unit test went from 10m to 3m, but I think most of this is that the linters consume less resources due to the simpler dependency tree.

codefromthecrypt:
yeah it is mostly the check phase and likely less bulk to download which is why PRs can check so much faster now.

Screen Shot 2021-04-15 at 11 20 17 AM
Screen Shot 2021-04-15 at 11 20 49 AM

pkg/extension/workspace/example/envoy/template/defaults.go
mathetake: Envoy folks usually uses `9901` for admin, but it's just a tradition I think.

mathetake (approved):
LGTM, but I want @lizan 's view before merging 🙂

codefromthecrypt:
rebased over #179 to bring down another 2MiB

```bash
du -k build/bin/*/amd64/getenvoy
20760	build/bin/darwin/amd64/getenvoy
18420	build/bin/linux/amd64/getenvoy
```

codefromthecrypt:
updated the description with the size difference based on this vs latest master (after rebasing)

codefromthecrypt:
@liamawhite I don't think you've been pinged directly on this, but I think you have been the only user of this feature. I searched github google etc and haven't seen any reference to this not made by an employee. We also know that the last release of getenvoy doesn't support any recent release of istio and no-one complained. Finally, it was marked experimental from the beginning. I hope you agree that for the health of the project it is time to delete this. If not please comment why and help pitch in on the maintenance impact which I could enumerate further on, but choosing to save keystrokes.

codefromthecrypt:
rebased for the 4th time. hopefully we can get an answer soon

codefromthecrypt:
Going to merge this so that we can conserve time. If there are any post-merge discussions about it, we can have them any time between now and our next release, which I don't know the schedule of.

PR #179, opened 2021-04-15
Removes dependency on docker distribution

codefromthecrypt:

### DIFF
--- a/pkg/cmd/extension/push/cmd_test.go
+++ b/pkg/cmd/extension/push/cmd_test.go
@@ -51,7 +51,7 @@ func TestGetEnvoyExtensionPush(t *testing.T) {
 
 	// Run "getenvoy extension push localhost:5000/getenvoy/sample"
 	c, stdout, stderr := cmd.NewRootCommand()
-	c.SetArgs([]string{"extension", "push", localRegistryWasmImageRef})
+	c.SetArgs([]string{"extension", "push", localRegistryWasmImageRef, "--use-http", "true"})
 	err := cmdutil.Execute(c)
 
 	// A fully qualified image ref includes the tag
@@ -106,7 +106,7 @@ func TestGetEnvoyExtensionPushWithExplicitFileOption(t *testing.T) {
 
 	// Run "getenvoy extension push localhost:5000/getenvoy/sample --extension-file testdata/workspace/extension.wasm"
 	c, stdout, stderr := cmd.NewRootCommand()
-	c.SetArgs([]string{"extension", "push", localRegistryWasmImageRef, "--extension-file", wasm})
+	c.SetArgs([]string{"extension", "push", localRegistryWasmImageRef, "--extension-file", wasm, "--use-http", "true"})
 	err := cmdutil.Execute(c)
 
 	// Verify the pushed a latest tag to the correct registry
--- a/pkg/extension/wasmimage/wasmimage.go
+++ b/pkg/extension/wasmimage/wasmimage.go
@@ -92,13 +92,8 @@ func isWasmBinary(path string) bool {
 func registryHosts(insecure, plainHTTP bool) docker.RegistryHosts {
 	client := http.DefaultClient
 
-	if insecure {
-		client.Transport = &http.Transport{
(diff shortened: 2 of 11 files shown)


codefromthecrypt:
this takes another almost 2MB off the binary size after #178

pkg/extension/wasmimage/wasmimage.go
```diff
@@ -85,3 +88,30 @@ func isWasmBinary(path string) bool {
...
+			},
+		}
+	}
+
+	return func(host string) ([]docker.RegistryHost, error) {
+		config := docker.RegistryHost{
+			Client:       client,
+			Host:         host,
+			Scheme:       "https",
+			Path:         "/v2",
```
mathetake, line 109: what's this `/v2` path..?

codefromthecrypt: good question.. I'll put a comment https://docs.docker.com/registry/spec/api/#overview

codefromthecrypt:
looks like we didn't indicate `--use-http true` before as the default went to accept plain text. From now on, if someone is using a registry that isn't https, they need to pass `--use-http true` to `getenvoy extension push`

pkg/cmd/extension/push/cmd_test.go
```diff
@@ -106,7 +106,7 @@ func TestGetEnvoyExtensionPushWithExplicitFileOption(t *testing.T) {
 
 	// Run "getenvoy extension push localhost:5000/getenvoy/sample --extension-file testdata/workspace/extension.wasm"
 	c, stdout, stderr := cmd.NewRootCommand()
-	c.SetArgs([]string{"extension", "push", localRegistryWasmImageRef, "--extension-file", wasm})
+	c.SetArgs([]string{"extension", "push", localRegistryWasmImageRef, "--extension-file", wasm, "--use-http", "true"})
```
codefromthecrypt, line 109: otherwise, it will end up as https://localhost:5000/v2/...

pkg/extension/wasmimage/wasmimage.go
```diff
@@ -85,3 +88,25 @@ func isWasmBinary(path string) bool {
...
+	if insecure && !plainHTTP {
+		client.Transport = &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}} // nolint:gosec
+	}
+
+	return func(host string) ([]docker.RegistryHost, error) {
+		config := docker.RegistryHost{
+			Client:       client,
+			Host:         host,
+			Scheme:       "https",
+			Path:         "/v2", // For Docker registry:2 https://docs.docker.com/registry/spec/api/#overview
```
mathetake, line 104: what if we push against other registries..? This should work with ECR and GCR as well (but unfortunately we don't have test infra right now). Maybe we could just remove `/v2`..?

mathetake: anyway I have to refactor the logic this packaging stuff since the OCI spec has changed since the PR for this feature was merged.. so we could leave this as is now.

codefromthecrypt: this isn't a change btw, before you just didn't see it adding /v2

codefromthecrypt: for example, the mock test I recently added? it had /v2 in there because the old client sent it also. If we delete it, not only will it not work, but the mock test will show it won't.

codefromthecrypt: ps most of this is following down the rabbit hole https://github.com/deislabs/oras/blob/a64b0ac4faf26c58016fbd7ed1a9eb9d049e54ee/cmd/oras/resolver.go because we support old flags you would end up noticing the default from docker is hard-coding the path to /v2

mathetake: I see, that makes sense! Thanks for clarifying! 💯

mathetake (approved):
Thanks a lot!

codefromthecrypt:
thanks for asking the hard questions @mathetake!

Issue #172, opened 2021-04-11
Refactor out osutil.SetupSignalHandler


codefromthecrypt:
`osutil.SetupSignalHandler` results in confusing code and flakey tests. Recently, there are more tools available, including [NotifyContext](https://golang.org/pkg/os/signal/#NotifyContext).

We should be able to reduce tech debt by deleting as many utilities here as possible. Using standard libraries will also help with cross-platform support. I suggest we delete `osutil.SetupSignalHandler` and handle the ctrl-c logging with `NotifyContext`. If sufficiently simplified, we could consider deleting the flakey test `TestRunShutdownError` and the shell scripts it uses.

Here are a couple examples of `NotifyContext` in use:
https://github.com/int128/kubectl-external-forward/blob/55f2b2b4623b9d20129078be0dd86f470a698ecf/pkg/externalforwarder/external_forwarder.go
https://github.com/nekoshita/golang-signal-notify-context-example

PR #180, opened 2021-04-16
Cleans signal handling and reduces complexity of "getenvoy run"

codefromthecrypt:
Before, we had complex signal handling in both "getenvoy run" and also
normal command invocation. This solves the first part by reducing the
amount of gears used to run the process, and also moving to
NotifyContext for signal interception.

See #172 for more information on NotifyContext, though this change
doesn't remove the osutil functions, yet. That will be a different PR
as it is a different branch of code.

### DIFF
--- a/pkg/binary/envoy/debug/log.go
+++ b/pkg/binary/envoy/debug/log.go
@@ -19,7 +19,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"time"
 
 	"github.com/tetratelabs/log"
 
@@ -29,8 +28,10 @@ import (
 
 // EnableEnvoyLogCollection is a preset option that registers collection of Envoy access logs and stderr
 func EnableEnvoyLogCollection(r *envoy.Runtime) {
-	if err := os.MkdirAll(filepath.Join(r.DebugStore(), "logs"), os.ModePerm); err != nil {
-		log.Errorf("unable to create directory to write logs to, no logs will be captured: %v", err)
+	logsDir := filepath.Join(r.DebugStore(), "logs")
+	err := os.MkdirAll(logsDir, os.ModePerm)
+	if err != nil {
+		log.Errorf("unable to create logs directory %q, so no logs will be captured: %v", logsDir, err)
 		return
 	}
 	r.RegisterPreStart(captureStdout)
@@ -42,13 +43,15 @@ func captureStdout(r binary.Runner) error {
 	if err != nil {
 		return err
 	}
+	r.RegisterPostTermination(func(runner binary.Runner) error {
+		return f.Close()
+	})
 	r.SetStdout(func(w io.Writer) io.Writer {
 		if w == nil {
 			return f
 		}
 		return io.MultiWriter(w, f)
 	})
-	go capture(r, f)
 	return nil
 }
 
@@ -57,13 +60,15 @@ func captureStderr(r binary.Runner) error {
 	if err != nil {
 		return err
 	}
+	r.RegisterPostTermination(func(runner binary.Runner) error {
+		return f.Close()
+	})
 	r.SetStderr(func(w io.Writer) io.Writer {
 		if w == nil {
 			return f
 		}
 		return io.MultiWriter(w, f)
 	})
-	go capture(r, f)
 	return nil
 }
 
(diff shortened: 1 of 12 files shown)


pkg/binary/envoy/run.go
```diff
@@ -46,26 +47,31 @@ func (r *Runtime) RunPath(path string, args []string) error {
...
 
-	ctx, cancel := context.WithCancel(context.Background())
-	r.ctx = ctx
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer cancel()
+	r.fakeInterrupt = cancel
 
-	// #nosec -> users can run whatever binary they like!
-	r.cmd = exec.Command(path, args...)
+	// We can't use CommandContext even if that seems correct here. The reason is that we need to invoke preTerminate
```
codefromthecrypt, line 54: I would have the whole morning back if this comment was here before I arrived!

pkg/binary/envoy/runtime.go
```diff
@@ -171,7 +164,11 @@ func (r *Runtime) envoyReady() (int, error) {
 	}
 }
 
-// SendSignal sends a signal to the parent process
-func (r *Runtime) SendSignal(s os.Signal) {
```
codefromthecrypt, line 175: this was cumulative days to understand, because there are so many signal processing code and it wasn't clear what was for testing and what was needed as barely any comments.

I hope the comment says why it exists at all, until we stop it from existing by moving this type of test to e2e which controls the lifecycle of the process and can kill it directly.

pkg/binary/envoy/termination.go
```diff
@@ -26,25 +26,27 @@ func (r *Runtime) handleTermination() {
...
-			log.Infof("Envoy process (PID=%d) exited successfully", r.cmd.Process.Pid)
+			log.Infof("Envoy process (PID=%d) exited successfully", p.Pid)
 			return
 		}
-		log.Infof("Envoy process (PID=%d) terminated prematurely", r.cmd.Process.Pid)
+		log.Infof("Envoy process (PID=%d) terminated prematurely", p.Pid)
 		return
 	}
 
+	defer func() { // Ensure the SIGINT forwards to Envoy even if a pre-termination hook panics
```
codefromthecrypt, line 39: before we could and did have processes dangling for this reason

codefromthecrypt:
looks like I'll be doing the log file close refactor now after all as it fails in linux otherwise :D

pkg/binary/envoy/debug/log.go
```diff
@@ -75,22 +80,3 @@ func createLogFile(path string) (*os.File, error) {
...
-	if err := file.Close(); err != nil {
-		log.Errorf("error closing access log file: %v", err)
-	}
-	r.RegisterDone()
-}
-
-// wait blocks until the child process reaches the state passed
-// Note: It does not guarantee that it is in the specified state just that it has reached it
-func wait(r binary.Runner, state int) {
-	for r.Status() < state {
```
codefromthecrypt, line 91: the key change of this PR is moving from indirect inspection of process state, to direct. Anything like archiving logs should happen after the process terminates. It is easier to understand what's going on by looking where the process was started.

codefromthecrypt:
once this is in, we can move most of the functions that await run completion to e2e, possibly in parallel depending on if #178 is allowed to merge or not (as it is the only thing that requires static ports)

pkg/binary/envoy/run.go
```diff
@@ -46,32 +45,50 @@ func (r *Runtime) RunPath(path string, args []string) error {
...
+	if cmd.ProcessState != nil {
+		log.Infof("Envoy process (PID=%d) terminated prematurely", cmd.Process.Pid)
+		return r.handlePostTermination()
 	}
-	return os.RemoveAll(r.DebugStore())
+
+	r.handleTermination()
+
+	// Block until the process is complete. This ensures file descriptors are closed.
+	<-waitCtx.Done()
```
codefromthecrypt, line 89: this is what now enforces clean stop prior to invoking any post termination hooks (currently only log and archive hooks)

pkg/binary/envoy/termination.go
```diff
@@ -15,40 +15,59 @@
...
+
+func interrupt(p *os.Process) {
+	log.Infof("Sending Envoy process (PID=%d) SIGINT", p.Pid)
+	_ = p.Signal(syscall.SIGINT)
+}
+
+func (r *Runtime) handlePostTermination() error {
+	for _, f := range r.postTermination {
+		if err := f(r); err != nil {
+			log.Error(err.Error())
```
jcchavezs, line 52: I recommend give context to this log, otherwise it is hard to know what was the application trying to do when this error happened.

```suggestion
			log.Errorf("failed to handle post termination: %v",  err)
```

pkg/binary/envoytest/util.go
```diff
@@ -115,24 +114,25 @@ func RequireRunTerminate(t *testing.T, r binary.Runner, options RunKillOptions)
...
-		// envoy.waitForTerminationSignals() registers SIGINT and SIGTERM
-		r.SendSignal(syscall.SIGTERM)
+	deferredInterrupt := func() {
+		r.FakeInterrupt()
 	}
 
 	defer func() {
-		if deferredTerminate != nil {
-			deferredTerminate()
+		if deferredInterrupt != nil {
```
jcchavezs, line 122: I don't see how `deferredInterrupt` can be `nil`. Maybe you just want to call it on defer directly? Also, unless `deferredInterrupt` gets mutated in the future put the if wrapping the defer to save some CPU :p

codefromthecrypt: because it drifted! great eye

musaprg (approved):
Looks good! :+1:

codefromthecrypt:
thanks for the review, @jcchavezs and @musaprg!

PR #184, opened 2021-04-19
Removes chrono from rust extension examples

### LINKED ISSUE #144: ci: consider removing use of `chrono` library from Rust init templates since it's the slowest to compile in `e2e` tests on Mac

codefromthecrypt:
This removes chrono from the logging examples. Using this library
contributes to compilation time and is also unrealistic as logging
libraries typically perform their own timestamps. We used the chrono
lib to convert the envoy time into a timestamp, but it is better for the
build to not use that api.

This also removes unused build aliases and renames .cargo/config to
.cargo/config.toml per recommendations:
https://doc.rust-lang.org/cargo/reference/config.html

Fixes #144

### DIFF
--- a/data/extension/init/templates/rust/envoy.access_loggers/default/src/logger.rs
+++ b/data/extension/init/templates/rust/envoy.access_loggers/default/src/logger.rs
@@ -1,9 +1,7 @@
 use std::convert::TryFrom;
 
-use chrono::{offset::Local, DateTime};
-
 use envoy::extension::{access_logger, AccessLogger, ConfigStatus, Result};
-use envoy::host::{log, ByteString, Clock, Stats};
+use envoy::host::{log, ByteString, Stats};
 
 use super::config::SampleAccessLoggerConfig;
 use super::stats::SampleAccessLoggerStats;
@@ -12,13 +10,11 @@ use super::stats::SampleAccessLoggerStats;
 pub struct SampleAccessLogger<'a> {
     config: SampleAccessLoggerConfig,
     stats: SampleAccessLoggerStats,
-    // This example shows how to use Time API provided by Envoy host.
-    clock: &'a dyn Clock,
 }
 
 impl<'a> SampleAccessLogger<'a> {
     /// Creates a new instance of Sample Access Logger.
-    pub fn new(clock: &'a dyn Clock, stats: &dyn Stats) -> Result<Self> {
+    pub fn new(stats: &dyn Stats) -> Result<Self> {
         let stats = SampleAccessLoggerStats::new(
             stats.counter("examples.access_logger.log_entries_total")?,
         );
@@ -26,14 +22,13 @@ impl<'a> SampleAccessLogger<'a> {
         Ok(SampleAccessLogger {
             config: SampleAccessLoggerConfig::default(),
             stats,
-            clock,
         })
     }
 
     /// Creates a new instance of Sample Access Logger
     /// bound to the actual Envoy ABI.
     pub fn default() -> Result<Self> {
-        Self::new(Clock::default(), Stats::default())
+        Self::new(Stats::default())
     }
 }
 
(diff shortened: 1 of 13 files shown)


data/extension/init/templates/rust/envoy.access_loggers/default/src/logger.rs
```diff
@@ -12,28 +10,25 @@ use super::stats::SampleAccessLoggerStats;
 pub struct SampleAccessLogger<'a> {
     config: SampleAccessLoggerConfig,
     stats: SampleAccessLoggerStats,
-    // This example shows how to use Time API provided by Envoy host.
```
codefromthecrypt, line 15: this lied as it only showed how to initialize it. the rust linter found this was never used

data/extension/init/templates/rust/envoy.access_loggers/default/src/logger.rs
```diff
@@ -67,11 +62,8 @@ impl<'a> AccessLogger for SampleAccessLogger<'a> {
         // Update stats
         self.stats.log_entries_total().inc()?;
 
-        let now: DateTime<Local> = self.clock.now()?.into();
```
codefromthecrypt, line 70: this just increased download and compilation time. normal logging config will show the time.

codefromthecrypt: @mathetake what's your take on this.. should we consume the value of `self.clock.now()` in a cheaper way? Personally, I think we aren't required to use this in a strange way just to use it, and it is better off not using it at all.

mathetake: this shows the usage of `Clock` API in envoy-sdk, so maybe we'd better do that in a cheaper way (i.e. not intoing Datatime<Local>).

codefromthecrypt: fair enough!

data/extension/init/templates/rust/envoy.access_loggers/default/src/logger.rs
```diff
@@ -67,11 +65,9 @@ impl<'a> AccessLogger for SampleAccessLogger<'a> {
...
-
         log::info!(
-            "logging at {} with config: {:?}",
-            now.format("%+"),
+            "logging at {:?} with config: {:?}",
+            self.clock.now(),
```
codefromthecrypt, line 70: this still uses the api, just doesn't bother converting it to a more readable date

codefromthecrypt:
I don't anticipate this being a slam dunk for build performance, but it helps a little bit. We still need to look deeply about why the rust build is so slow

data/extension/init/templates/rust/envoy.access_loggers/default/src/logger.rs
mathetake, line 70:
```suggestion
            self.clock.now()?,
```

This would make the output a little bit less ugly from `Ok(SystemTime(1618809891.154186s))` to `SystemTime(1618809891.154186s)`

codefromthecrypt: thanks.. I'll apply to the other places also.

codefromthecrypt:

PR #188, opened 2021-04-21
Templatizes current version of envoy

codefromthecrypt:
This allows us to cache the download of envoy when it doesn't change.


pkg/cmd/run.go
```diff
@@ -41,29 +42,25 @@ var (
 // NewRunCmd create a command responsible for starting an Envoy process
 func NewRunCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "run <reference|filepath> [flags] [-- <envoy-args>]",
```
codefromthecrypt, line 44: this didn't work so removed it

pkg/reference.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2019 Tetrate
```
codefromthecrypt, line 1:
```suggestion
// Copyright 2021 Tetrate
```

.github/workflows/commit.yaml
```diff
@@ -121,10 +122,11 @@ jobs:
         uses: actions/cache@v2
         with:
           path: |
+            ~/.getenvoy/builds
```
codefromthecrypt, line 125: I need to look into multiple restore keys, if possible because these versions have nothing to do with eachother

codefromthecrypt: documented

codefromthecrypt:
cache works

Screen Shot 2021-04-21 at 1 11 18 PM

pkg/binary/envoytest/util.go
```diff
@@ -27,20 +27,17 @@ import (
...
-// This currently latest, but we may support a range at some point.
-var reference = "standard:1.17.1"
-
 var once sync.Once
 var errorFetchingEnvoy error
 
 // FetchEnvoyAndRun retrieves the Envoy indicated by reference only once. This is intended to be used with TestMain.
-// In CI, you can execute this to obviate latency during test runs: "go run cmd/getenvoy/main.go fetch standard:1.17.1"
+// Execute this to obviate latency during test runs: go run cmd/getenvoy/main.go fetch $(cat pkg/reference.txt)
 func FetchEnvoyAndRun(m *testing.M) {
```
codefromthecrypt, line 41: a later change will move tests that use this into the e2e thing and make this go away

codefromthecrypt:
looks like total 500MB cached now (dependencies and envoy binary)

mathetake:
so how much this makes tests faster..?

codefromthecrypt:
8s down to 2s, though there's probably some increase in time unpacking the cache.

Faster isn't the only thing here, it is also more reliable and less clutter log output. Ex no 503 on bintray unless the version changes as the cache is local. Even bigger than that is less copy/paste as that did drift in the past

Before:
```
Wed, 21 Apr 2021 22:53:36 GMT
Run go run cmd/getenvoy/main.go fetch standard:1.17.1 

—snip—

Wed, 21 Apr 2021 22:53:44 GMT go test ./pkg/...
```

After:
```
Wed, 21 Apr 2021 14:02:28 GMT
Run go run cmd/getenvoy/main.go fetch $(cat pkg/reference.txt) 

Wed, 21 Apr 2021 14:02:30 GMT standard:1.17.1/linux-glibc is already downloaded 

Wed, 21 Apr 2021 14:02:30 GMT go test ./pkg/...
```

codefromthecrypt:
IOTW, test execution wouldn't be faster unless we had a drift bug somewhere because we pre-seeded the cache before and after.

mathetake (approved):
Thanks!

codefromthecrypt:
Thanks again for asking why @mathetake

Issue #189, opened 2021-04-21
Remove support for initializing and building extensions in a language besides tinygo


codefromthecrypt:
This is a placeholder issue which will be elaborated over the next couple days with justification to remove rust and solely support tinygo for new extensions.

"getenvoy extension push --extension-file" supports any wasm, which could have been built in rust or another language.

It is understood that rust served a purpose in developing the wasm extension feature. While we are grateful for that experience, we cannot pretend this choice hasn't come at a heavy price that costs us every day. Future revisions to this issue will enumerate the costs as well as any technical gaps we have and how they would be resolved.

lizan:
The purpose of the getenvoy extension toolkit is to have a standard way to develop wasm extensions for Envoy, for many languages. While Rust costs much we should never remove it.

codefromthecrypt:
Will flesh out the description later, but first phase is I made a tag and tagged every change where rust was the cause or dominated the effort. There are still some issues yet to be solved and they are labeled also. https://github.com/tetratelabs/getenvoy/issues?q=label%3Arust+

codefromthecrypt:
Also it is understood that https://github.com/tetratelabs/envoy-wasm-rust-sdk and its dependency on a personal fork as described in #183... that the only consumer of these repos are this project. It is understood that they were over a year in the making and this can be an emotional thing to consider stopping work like that. If people feel strongly about that, they could still work on them even if this project doesn't depend on it directly anymore.

codefromthecrypt:
It became clear to me folks may have misunderstood the end state.

* this project no longer depends on non-upstream rust sdk in any way
* this project stops embedding rust builds (via extension init, run, test)
* instead we introduce a pre-canned example in rust using the normal sdk and prove we can run it

@lizan before I rewrite description etc, do you feel any differently with how I've restated the change?

codefromthecrypt:
obviated by #200

PR #190, opened 2021-04-21
Removes custom docker+osx matrix for rust

### LINKED ISSUE #145: macOS+rust e2e tests in CI are pinned to an over two year old version of Docker

codefromthecrypt:
This removes the custom script for rust at expense of longer runtimes,
last checked at 1hr for the e2e.

Fixes #145


codefromthecrypt:
finally pass. regardless of #189, we can consider merging this at the cost of time as it shows we work with a relevant version of docker on osx.

Screen Shot 2021-04-21 at 5 31 09 PM

mathetake (approved):

PR #191, opened 2021-04-21
run builders mount in delegated consistency

lizan:
Defaults the source mount consistency to delegated, this speeds up builds a lot especially on Docker for Mac.

Signed-off-by: Lizan Zhou <lizan@tetrate.io>

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -103,7 +103,7 @@ jobs:
           - extension-language: rust
             # Attempt to avoid re-downloading almost 200MB of deps per test that implies "getenvoy extension build"
             # This uses the RUNNER_TEMP variable directly to avoid having to mkdir a sub-path.
-            toolchain-container-options: "-v ${RUNNER_TEMP}:/tmp/cargohome -e CARGO_HOME=/tmp/cargohome"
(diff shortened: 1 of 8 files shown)


codefromthecrypt:
really appreciate the attempt here.

I was looking into this, especially why it isn't documented anymore and finally found why https://github.com/docker/for-mac/issues/5402 If this is for mac we probably don't want to add this even if it might have helped in an older version. You'll notice the run time didn't change. I suppose they allowed docker to not crash on the old args for convenience..

lizan:
Yeah though it appears to be faster on my local mac, while I don't use them frequently. The other approach could be mount cache from docker volume. I'll give it another shot.

codefromthecrypt:
sure. I want you to not waste the same time I did, so here are some bread crumbs.
* indeed it is fast on local mac for reasons including memory. I don't have the command to monitor the memory usage during a rust build, but I suspect it is high and possibly contributing
* We reduced time significantly by volume copying here https://github.com/tetratelabs/getenvoy/blob/master/images/extension-builders/rust/commands.sh#L28-L30
* the download cache between runs is here https://github.com/tetratelabs/getenvoy/blob/master/.github/workflows/commit.yaml#L109

good luck!

codefromthecrypt:
obviated by #200

PR #195, opened 2021-04-22
Removes `getenvoy extension run --envoy-path` and cleans tech debt

codefromthecrypt:
`--envoy-path` was mainly used for tests and is removed, but the bulk
of this change is tech debt cleanup.

It is understood that a large amount of tech debt cleanup is here. The
bulk of ideas here were reviewed and @mathetake accepted review burden
before and during this work. We agreed that someone has to clean this
up and it doesn't get easier later.
# No more overriding envoy path
We introduced options to allow overriding the location of the envoy binary, effectively subverting the idea that getenvoy results in a known working binary. This feature was earlier removed from `getenvoy run` where we formerly sneakily allowed the envoy version to be file path! This completes the job by also removing the `--envoy-path` from `getenvoy extension run`.

As described below, we have other ways to control unit tests now, and we shouldn't bleed that into the CLI arguments unless end-users that don't work here request it.

# Tech debt
Below describes tech debt cleanup, things internally that don't effect users directly

## Less jargon is better
Before we had a myriad of different directories and names for similar directories.

* `DebugStore` -> `WorkingDir`
* `TmplDir` -> `WorkingDir`
* `DebugDir` -> `WorkingDir/..` (until we top level the concept and it should be archive dir!)
* `RootDir` -> `HomeDir`
* `WorkspaceDir` -> `ExtensionDir`
* `GetEnvoyBinaryPath` -> `GetEnvoyPath`
* `/source` volume mount in docker -> `ExtensionDir`
* `OutputDir` -> `ExtensionDir`
* `NoPrompt` -> `NoWizard`
* `StylesEnabled` -> `NoColors`

## Under abstraction hurts badly
We had several shell scripts that did similar things as well as several test fixtures that also did similar things. Once something repeats a few times, it is now time to abstract!

* `RequireCaptureScript` consolidates the following into :::, using more portable shell syntax.
  * `pkg/binary/envoy/testdata/sleep.sh`
  * `pkg/extension/workspace/toolchain/builtin/testdata/toolchain/docker`
  * `pkg/util/exec/testdata/test_stdio.sh`
  * `pkg/util/exec/testdata/test_trap_sigterm.sh`

As a result, there are few shell scripts and can be manually formatted on change. This also removes the need for `shfmt` from `.bingo` and `Makefile`.

## Over abstraction hurts badly
We had many occurrences of interfaces directly coupled to implementation. This assumes a lot including what another implementation might feel like, but more importantly, it creates navigation problems and also type casting everywhere (ex `r.(*Runtime)`. Moreover, it always hurts test coverage. We should stop abstracting until there are 3 implementations of something.

* `binary.Runner` had only one implementation (Runtime) so `pkg/extension/workspace/example/runtime` is gone
* `printer interface` -> `io.Writer`
* `dockerutil.ImageName` -> `fmt.Sprintf`
* `names.Add` -> `append`
* `GetCurrentUser` -> `user.Current`
* `ErrorHandlers` -> explicit code that does the same
* `RunContext` gone
*  `runtime.go` (`RuntimeOption` etc) gone as no longer needed
* `FetchRunner` is not needed to be coupled to `Runtime` and gone
* `ComposableRunE` etc gone as not needed

## No envoy in unit tests is a happy place
Running envoy in unit tests makes things like coverage tools difficult. It makes it near impossible to run offline, slows tests, and complicates CI

* no need for the large amount of "status" code used only by debug tests
* no more fetching envoy in GH action for the unit tests
* less checked in envoy config in testate
* "debug" unit tests run significantly faster than before
* "debug" tests can be improved in the future covering far more corner cases than easy to orchestrate with a real envoy process.

## Less output clutter is better
Before, we spammed the user with inconsequential knowledge, such as the PID that is already dead. Each line we choose to mark info or higher should be impactful. Even though we override stout/err in most unit tests, our command code defaulted to real stdout/stderr. This is worse than nil for unit tests especially as people shouldn't be bothered with lines of output unless they matter.

* logging of pid when it already died
* redundant logging that includes the function name that's already logged
* Run doesn't default unset to os.Stdout and os.Stderr anymore

## Implicits and mutable global variables usually hurt
Formerly, we had a number of implicit lookup code, which made navigation even more difficult and harder to control. This also hurt test coverage. While it can be tedious to retrofit code to be configurable, mutable global variables are worse. The extra setup/teardown to capture and revert values id distracting. It also doesn't play well in parallel tests.

* `home.go` -> `GlobalOpts.HomeDir`
* `manifest.SetURL()` -> `GlobalOpts.ManifestURL`
* `GetCurrentWorkspace()` -> `GetWorkspaceAt(o.ExtensionDir)`
* `FindWorkspaceDir` is no longer needed
* `OverrideGetCurrentUser` -> `builtin.DefaultDockerUser`
* `globals.NoPrompt` -> `GlobalOpts.NoWizard`
* `globals.NoColors` -> `GlobalOpts.NoColors`
* PWD overrides for extension commands -> `GlobalOpts.ExtensionDir`
* PATH overrides for docker -> `GlobalOpts.DockerPath`
* `envoy.tmpl.yaml` -> `envoy.yaml`


codefromthecrypt:
pinged a few for review, but first one wins a 🐤

pkg/binary/envoy/runtime.go
```diff
@@ -57,39 +58,33 @@ func (o RuntimeOptions) AndAll(opts RuntimeOptions) RuntimeOptions {
...
-type fetcher struct {
-	store string
-}
-
 // Runtime manages an Envoy lifecycle including fetching (if necessary) and running
 type Runtime struct {
 	binary.FetchRunner
-	fetcher
 
-	RootDir  string
```
codefromthecrypt, line 90: renamed a few variants to homeDir as there's no need to confuse people

pkg/binary/envoy/runtime.go
```diff
@@ -57,39 +58,33 @@ func (o RuntimeOptions) AndAll(opts RuntimeOptions) RuntimeOptions {
...
-	}
-
 	for _, option := range options {
 		option(runtime)
 	}
 	return runtime, nil
 }
 
-type fetcher struct {
-	store string
```
codefromthecrypt, line 82: this type didn't really decouple anything, but it did add complexity

pkg/binary/envoy/runtime.go
```diff
@@ -57,39 +58,33 @@ func (o RuntimeOptions) AndAll(opts RuntimeOptions) RuntimeOptions {
...
+	if debugDir == "" {
+		debugDir = filepath.Join(homeDir, "debug")
+	}
+
 	runtime := &Runtime{
-		RootDir:         local,
-		fetcher:         fetcher{local},
-		TmplDir:         filepath.Join(local, "templates"),
+		homeDir:         homeDir,
+		debugStore:      filepath.Join(debugDir, strconv.FormatInt(time.Now().UnixNano(), 10)),
```
codefromthecrypt, line 69: exact same pattern as before, just based over a variable. We can't easily undo the unixNano thing, and there's not a big need to change that as long as we can change the basedir

pkg/cmd/extension/run/cmd.go
```diff
@@ -209,7 +210,7 @@ Run Envoy extension in the example setup.`,
 			// run example
 			opts.Run.Workspace = workspace
 			opts.Run.Example.Example = example
-			err = examples.RunExample(&runtime.RunContext{
+			err = examples.RunExample(global.HomeDir, global.DebugDir, &runtime.RunContext{
```
codefromthecrypt, line 213: it is strange, but another file initializes global.HomeDir, so I repeated the pattern here.

pkg/cmd/root.go
```diff
@@ -65,7 +65,7 @@ bootstrap generation and automated collection of access logs, Envoy state and ma
 			if rootOpts.HomeDir == "" {
 				return errors.New("GetEnvoy home directory cannot be empty")
 			}
-			common.HomeDir = rootOpts.HomeDir
+			global.HomeDir = rootOpts.HomeDir
```
codefromthecrypt, line 68: renamed common to global to avoid name clash and also it is a global variable

pkg/cmd/run.go
```diff
@@ -68,45 +67,19 @@ getenvoy run postgres:nightly --templateArg endpoints=127.0.0.1:5432,192.168.0.1
...
 	cmd.Flags().StringToStringVar(&templateArgs, "templateArg", map[string]string{},
 		"arguments passed to a config template for substitution")
+	cmd.PersistentFlags().StringVar(&global.DebugDir, "debug-dir", "",
+		`Store debug data in an alternate directory. Defaults to envoy_home/debug/timestamp`)
 	return cmd
 }
-
-// Function creates config file based on template args passed by a user.
-// The return value is Envoy command line option which must be passed to Envoy.
-func processTemplateArgs(flavor string, templateArgs map[string]string, runtime *envoy.Runtime) (string, error) {
```
codefromthecrypt, line 88: all this was unnecessary as you can pass a yaml literal. deleting this increases our test coverage.

pkg/flavors/flavors.go
```diff
@@ -50,7 +50,7 @@ func GetFlavor(flavor string) (FlavorConfigTemplate, error) {
 }
 
 // CreateConfig function checks flavor specific parameters, get flavor's template and
-// create a config.
+// create a config yaml used as the `--config-yaml` argument to Envoy.
```
codefromthecrypt, line 53: new comment says why we don't need any more things except the literal. ps this is only used for postgres I think

test/e2e/getenvoy_extension_build_test.go
```diff
@@ -34,19 +34,17 @@ func TestGetEnvoyExtensionBuild(t *testing.T) {
 		test := test // pin! see https://github.com/kyoh86/scopelint for why
 
 		t.Run(test.String(), func(t *testing.T) {
+			// TODO: uses Docker, but may be possible to run parallel
```
codefromthecrypt, line 37: TODO, but not on this PR as we have more work to do. For example all the debug tests are e2e and should move here first

test/e2e/getenvoy_extension_build_test.go
```diff
@@ -34,19 +34,17 @@ func TestGetEnvoyExtensionBuild(t *testing.T) {
...
 			workDir, removeWorkDir := RequireNewTempDir(t)
 			defer removeWorkDir()
 
-			_, revertChDir := RequireChDir(t, workDir)
```
codefromthecrypt, line 40: chdir is not safe in parallel, so that's why I had to redo variables everywhere to be explicit.

test/e2e/getenvoy_run_test.go
```diff
@@ -63,35 +65,6 @@ func TestGetEnvoyRun(t *testing.T) {
...
 
-// backupDebugDir backs up ${GETENVOY_HOME}/debug in case the test hasn't overridden it and the developer has existing
-// data there. The function returned reverts this directory.
-//
-// Typically, this will run in the default ~/.getenvoy directory, as a means to avoid re-downloads of files such as
-// .getenvoy/builds/standard/1.17.1/darwin/bin/envoy (~100MB)
-//
-// While CI usually overrides the `HOME` variable with E2E_TOOLCHAIN_CONTAINER_OPTIONS, a developer may be
-// running this on their laptop. To avoid clobbering their old debug data, backup the
-func backupDebugDir(t *testing.T) (string, func()) {
```
codefromthecrypt, line 74: not parallel safe either

codefromthecrypt (commented):
notes

codefromthecrypt:
going to revise this a bit as the root options are confusing and I feel like fixing that now

codefromthecrypt:
@mathetake Finally completed 6 days of effort on this. I appreciate the review is hard, but I know we share stake in the future of this project, 2k less lines is worth it I hope.

codefromthecrypt:
I spent a couple hours writing the description. The only thing worse than a huge change is a one line description. Hope this helps and sets a good example.

codefromthecrypt:
I may have tightened up a timeout too much here, will look at it tomorrow, but it doesn't effect the change in general

```
[2021-04-28 09:42:10.803][11536][info][main] [external/envoy/source/server/server.cc:731] starting main dispatch loop
2021-04-28T09:42:10.946448Z	info	discovered admin address: 127.0.0.1:39289
    getenvoy_run_test.go:79: 
        	Error Trace:	getenvoy_run_test.go:79
        	            				getenvoy_run_test.go:55
        	Error:      	Received unexpected error:
        	            	reached timeout 1m0s
        	Test:       	TestGetEnvoyRun
        	Messages:   	error parsing admin address from stdout of [: /home/runner/work/getenvoy/getenvoy/build/bin/linux/amd64/getenvoy run standard:1.17.1 -- --config-yaml admin: {access_log_path: '/dev/stdout', address: {socket_address: {address: '127.0.0.1', port_value: 0}}}]
2021-04-28T09:43:10.754283Z	info	GetEnvoy process (PID=11531) received SIGINT
```

go.mod
```diff
@@ -17,7 +17,6 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/opencontainers/image-spec v1.0.1
 	github.com/otiai10/copy v1.5.1
-	github.com/pkg/errors v0.9.1
```
mathetake, line 20: 🎉

cmd/getenvoy/main.go
```diff
@@ -17,13 +17,12 @@ package main
...
 
+	"github.com/tetratelabs/getenvoy/pkg/binary/envoy/globals"
 	"github.com/tetratelabs/getenvoy/pkg/cmd"
-	cmdutil "github.com/tetratelabs/getenvoy/pkg/util/cmd"
 )
 
 func main() {
-	root := cmd.NewRoot()
-	if err := cmdutil.Execute(root); err != nil {
+	if cmd.Execute(cmd.NewRoot(&globals.GlobalOpts{})) != nil {
```
mathetake, line 25: better print the error returned by cmd.Execute?

codefromthecrypt: it is probably not easy to tell but `cmd.Execute(cmd)` does this. We have some special case logging we do which is different `cmd.Execute()` one day I hope we can use a different cli lib

codefromthecrypt: I think maybe I can rename the import as the word clash is confusing

pkg/cmd/extension/test/cmd_test.go
```diff
@@ -141,38 +128,39 @@ func TestGetEnvoyExtensionBuildWithDockerOptions(t *testing.T) {
...
+		o.ExtensionDir, o.DockerPath, expectedDockerArgs)
+	require.Equal(t, expectedStdout, stdout.String(), `expected stdout running [%v]`, c)
 
 	// We also expect "docker stderr" in the output for the same reason.
 	expectedStderr := fmt.Sprintf("docker stderr\nError: %s\n\nRun 'getenvoy extension test --help' for usage.\n", expectedErr)
 	require.Equal(t, expectedStderr, stderr.String(), `expected stderr running [%v]`, c)
 }
+
+// requireGlobalOpts returns the options needed to run a docker-based extension command and a cleanup function.
+func requireGlobalOpts(t *testing.T) (*globals.GlobalOpts, func()) {
```
mathetake, line 158: same here

pkg/cmd/fetch.go
```diff
@@ -16,46 +16,37 @@ package cmd
...
 Retrieves the referenced Envoy binary from GetEnvoy. The reference can be a full or partial reference.
 A complete list of available builds can be retrieved using` + "`getenvoy list`" + `.`,
-		Example: `# Fetch using a partial manifest reference to retrieve a build suitable for your operating system.
-getenvoy fetch standard:1.11.1
+		Example: fmt.Sprintf(`# Fetch using a partial manifest reference to retrieve a build suitable for your operating system.
+getenvoy fetch %[1]s
 		
 # Fetch using a full manifest reference to retrieve a specific build for Linux. 
-getenvoy fetch standard:1.11.1/linux-glibc`,
+getenvoy fetch s%[1]s/linux-glibc`, reference.Latest),
```
mathetake, line 40: nit: why can't we simply use `%s` instead of `s%[1]s`?

pkg/cmd/fetch.go
```diff
@@ -16,46 +16,37 @@ package cmd
...
 	return &cobra.Command{
 		Use:   "fetch <reference>",
 		Short: "Retrieve Envoy binaries from GetEnvoy.",
 		Long: `
 Retrieves the referenced Envoy binary from GetEnvoy. The reference can be a full or partial reference.
 A complete list of available builds can be retrieved using` + "`getenvoy list`" + `.`,
-		Example: `# Fetch using a partial manifest reference to retrieve a build suitable for your operating system.
-getenvoy fetch standard:1.11.1
+		Example: fmt.Sprintf(`# Fetch using a partial manifest reference to retrieve a build suitable for your operating system.
+getenvoy fetch %[1]s
```
mathetake, line 37: same here

pkg/cmd/run/cmd.go
```diff
@@ -0,0 +1,117 @@
+// Copyright 2019 Tetrate
```
mathetake, line 1: seems inconsistent to me because other top commands are in `pkg/cmd` directory, and  only this `run` cmd has dedicated directory `pkg/cmd/run`.

pkg/cmd/run/cmd.go
```diff
@@ -0,0 +1,117 @@
...
+		},
+	}
+	cmd.Flags().StringToStringVar(&templateArgs, "templateArg", map[string]string{},
+		"arguments passed to a config template for substitution")
+	return cmd
+}
+
+// InitializeRunOpts allows us to handle defaulting when tests override run parameters.
+// Notably, this ensures globals.RunOpts EnvoyPath and WorkingDir exist.
+func InitializeRunOpts(o *globals.GlobalOpts, r string) error {
```
mathetake, line 88: could you give more descriptive name to `r string`? hard to figure out what it is about..

pkg/cmd/run/cmd.go
```diff
@@ -0,0 +1,117 @@
...
+		// When the directory is implicitly generated, we should create it to avoid late errors.
+		if err := os.MkdirAll(workingDir, 0750); err != nil {
+			return fmt.Errorf("unable to create working directory %q, so we cannot run envoy: %w", workingDir, err)
+		}
+		runOpts.WorkingDir = workingDir
+	}
+	return nil
+}
+
+// Run enables debug and runs Envoy with the IO from the cobra.Command
```
mathetake, line 111: i think it's better to have comment about the fact  that this is used for `extension run`.

pkg/extension/workspace/fs/interfaces.go
```diff
@@ -20,8 +20,8 @@ const (
 	extensionMetaDir = ".getenvoy/extension"
 )
 
-// WorkspaceDir represents a directory with an extension created by getenvoy toolkit.
-type WorkspaceDir interface {
+// ExtensionDir represents a directory with an extension created by getenvoy toolkit.
+type ExtensionDir interface {
```
mathetake, line 24: the same name appears in GlobalOpts which makes me confused though not sure what is better 😓

pkg/extension/workspace/fs/interfaces.go
```diff
@@ -20,8 +20,8 @@ const (
...
 	// GetRootDir returns path to the root dir of an extension.
 	GetRootDir() string
```
mathetake, line 26: maybe we could delete this interface or merge this to GlobalOpts?

pkg/extension/workspace/example/runtime/interfaces.go
```diff
@@ -16,7 +16,6 @@ package runtime
```
mathetake, line 16: the contents here are not "intefaces" anymore 😄

pkg/extension/workspace/toolchain/builtin/default_config_test.go
```diff
@@ -36,7 +36,7 @@ func TestDefaultConfigForValidate(t *testing.T) {
 	for i, lang := range extension.Languages {
 		expected, err := os.ReadFile(fmt.Sprintf("testdata/default_config/%s.toolchain.yaml", lang))
 		if err != nil {
-			panic(fmt.Errorf("missing default config for language %s: %w", lang, err))
+			t.Fatalf("missing default config for language %s: %v", lang, err)
```
mathetake, line 39: can we user `require` instead of `t.Fatalf` to be more consistent?

pkg/extension/workspace/toolchain/builtin/example_config.go
```diff
@@ -101,7 +100,7 @@ func renderExampleConfigTemplate(toolchain *builtinconfig.ToolchainConfig) []byt
 	err = tmpl.Execute(&out, toolchain)
 	if err != nil {
 		// must be caught by unit tests
-		panic(errors.Wrap(err, "failed to render example configuration template"))
+		panic(fmt.Errorf("failed to render example configuration template: %w", err))
```
mathetake, line 103: same here

pkg/extension/workspace/toolchain/registry/types.go
```diff
@@ -14,15 +14,15 @@
```
mathetake, line 14: wanna kill this directory since it contains another global states outside of `globals` pkg.

pkg/cmd/extension/init/cmd_test.go
```diff
@@ -100,58 +102,50 @@ func TestGetEnvoyExtensionInitValidateFlag(t *testing.T) {
...
 		}
 	}
 
 	for _, test := range tests {
 		test := test // pin! see https://github.com/kyoh86/scopelint for why
 
 		t.Run(test.name, func(t *testing.T) {
-			outputDir, revertOutputDir := RequireNewTempDir(t)
-			defer revertOutputDir()
+			extensionDir, revertExtensionDir := RequireNewTempDir(t)
```
mathetake, line 128: `revert` make me think that this would revert something e.g. going back to the some directory, etc. So maybe it's more appropriate to use `tearDownDir` or similar because it's just deleting the tempdir, right?

mathetake: same goes for others

pkg/cmd/extension/init/params.go
```diff
@@ -42,10 +42,10 @@ func (p *param) IsValid() bool {
...
 // params represents all parameters to be filled in by a user.
 type params struct {
-	Category  param
-	Language  param
-	OutputDir param
-	Name      param
+	Category     param
+	Language     param
+	ExtensionDir param
+	Name         param
```
mathetake, line 48: no need to export these fields 😄

mathetake: same goes for `param` struct above 😄

mathetake (commented):
first shot 🔫

pkg/cmd/fetch.go
```diff
@@ -16,46 +16,37 @@ package cmd
...
 		
 # Fetch using a full manifest reference to retrieve a specific build for Linux. 
-getenvoy fetch standard:1.11.1/linux-glibc`,
+getenvoy fetch s%[1]s/linux-glibc`, reference.Latest),
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) == 0 {
 				return errors.New("missing binary parameter")
```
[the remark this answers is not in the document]
codefromthecrypt: yeah actually sometimes we say reference and version interchangably. let's fix this part and maybe a later PR as we find more fuzz

pkg/cmd/fetch.go
mathetake: ah - nvm.

pkg/binary/envoy/fetch_test.go
```diff
@@ -29,22 +27,22 @@ import (
 
 	reference "github.com/tetratelabs/getenvoy/pkg"
 	"github.com/tetratelabs/getenvoy/pkg/manifest"
+	"github.com/tetratelabs/getenvoy/pkg/test/morerequire"
 )
```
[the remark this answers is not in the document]
codefromthecrypt: will carve out a basic test for it. generally it does too much, but agree there's not much reason to kick the can

pkg/cmd/run/cmd.go
codefromthecrypt: mainly I was trying to keep the PR from getting more massive. there was a circular dependency this resolved. (you'll notice the extension commands are in different dirs, but also there are almost no tests for the base commands either). Ack this is inconsistent, but I think we can do this after the fact as it will add a lot of code to backfill the missing tests. trust me to follow-up?

pkg/cmd/root.go
```diff
@@ -15,97 +15,109 @@
...
 
-// enableLoggingConfig checks whether logging should be configurable.
-//
-// At the moment, logging configuration is disabled by default to avoid abundance of options.
-//
-// TODO(yskopets): consider introducing simplified configuration options.
-func enableLoggingConfig() bool {
-	if enable, err := strconv.ParseBool(os.Getenv("EXPERIMENTAL_GETENVOY_LOGGING_CONFIG")); err == nil {
-		return enable
+func initializeGlobalOpts(globalOpts *globals.GlobalOpts) (string, string) {
```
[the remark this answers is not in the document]
codefromthecrypt: ok! as long as it doesnt' end up requiring cobra types sgtm

test/e2e/util/command.go
```diff
@@ -31,8 +31,8 @@ import (
 )
 
 var (
-	// GetEnvoyBinaryPath holds a path to a 'getenvoy' binary under test.
-	GetEnvoyBinaryPath = "getenvoy"
+	// GetEnvoyPatharyPath holds a path to a 'getenvoy' binary under test.
+	GetEnvoyPatharyPath = "getenvoy"
```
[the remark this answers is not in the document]
codefromthecrypt: it is find/replace :D

pkg/binary/envoytest/util.go
```diff
@@ -16,130 +16,79 @@ package envoytest
...
+	fakeEnvoy, removeFakeEnvoy := morerequire.RequireCaptureScript(t, "envoy")
+	defer removeFakeEnvoy()
 
-// FetchEnvoyAndRun retrieves the Envoy indicated by reference only once. This is intended to be used with TestMain.
-// Execute this to obviate latency during test runs: go run cmd/getenvoy/main.go fetch $(cat pkg/reference.txt)
-func FetchEnvoyAndRun(m *testing.M) {
-	once.Do(func() {
-		errorFetchingEnvoy = fetchEnvoy()
-	})
+	o := &globals.RunOpts{EnvoyPath: fakeEnvoy, WorkingDir: filepath.Join(debugDir, "1")}
```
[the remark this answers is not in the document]
codefromthecrypt: fake timestamp, our existing framework expects the run directory to be a subpath (any subpath) of the debug dir. I think that assumption is busted but also changing the "debug" concept in this PR would be massive, so I just faked the timestamp. I'll make a local called fakeTimestamp as that hopefully will make less distraction until we can cleanup the underlying problem.

pkg/cmd/extension/clean/cmd_test.go
```diff
@@ -141,38 +129,40 @@ func TestGetEnvoyExtensionCleanWithDockerOptions(t *testing.T) {
...
+		o.ExtensionDir, o.DockerPath, expectedDockerArgs)
+	require.Equal(t, expectedStdout, stdout.String(), `expected stdout running [%v]`, c)
 
 	// We also expect "docker stderr" in the output for the same reason.
 	expectedStderr := fmt.Sprintf("docker stderr\nError: %s\n\nRun 'getenvoy extension clean --help' for usage.\n", expectedErr)
 	require.Equal(t, expectedStderr, stderr.String(), `expected stderr running [%v]`, c)
 }
+
+// requireGlobalOpts returns the options needed to run a docker-based extension command and a cleanup function.
+func requireGlobalOpts(t *testing.T) (*globals.GlobalOpts, func()) {
```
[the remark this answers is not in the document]
codefromthecrypt: good idea!

pkg/binary/envoy/globals/globals.go
```diff
@@ -0,0 +1,71 @@
+// Copyright 2021 Tetrate
```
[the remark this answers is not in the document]
codefromthecrypt: ok

pkg/cmd/fetch.go
codefromthecrypt: same value is used twice. hard to see because there's an extra newline

pkg/cmd/run/cmd.go
codefromthecrypt: it is to unwind an import cycle needed to share code since getenvoy run and getenvoy extension run share logic, since we mentioned earlier, I'll leave it to that comment.

pkg/extension/workspace/fs/interfaces.go
codefromthecrypt: if you look carefully, you'll notice a lot of code that is still something to clean up, but practically speaking it is hard to do it all in one PR as even this was 6 full days of work. For now, at least keeping the same name for the same thing helps as future cleanups are easier to identify. Make sense?

pkg/extension/workspace/fs/interfaces.go
codefromthecrypt: maybe but not now, as this whole directory tree is complicated I hope it can disappear, but it was too much to comprehend at the time and logically at least something we can identify as redundant vs before.

pkg/extension/workspace/example/runtime/interfaces.go
codefromthecrypt: haha renamed to types for now. hopefully a later PR we can delete even more of the intermediate types

pkg/extension/workspace/toolchain/builtin/default_config_test.go
codefromthecrypt: yep!

pkg/extension/workspace/toolchain/builtin/example_config.go
codefromthecrypt: can't on this one as it isn't a test. the presence of panics everywhere is more than what we should do in this PR. I cleaned up the ones that were silly (ex the function already returns an error but we chose to panic instead). We can do a later PR to hopefully get rid of all the panics.

pkg/extension/workspace/fs/interfaces.go
mathetake: sg

pkg/extension/workspace/toolchain/registry/types.go
codefromthecrypt: ok I will delete it in this PR

pkg/cmd/root.go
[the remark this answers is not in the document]
codefromthecrypt: this is too coupled to the cobra flags. I would like to let the tension build here for removing cobra

pkg/cmd/run/cmd.go
codefromthecrypt: definitely

pkg/extension/workspace/toolchain/registry/types.go
codefromthecrypt: can't as the change is too huge, but we can in a later one

pkg/cmd/extension/init/cmd_test.go
codefromthecrypt: will go with removeExtensionDir as it is less jargon than what I had before

pkg/cmd/extension/init/params.go
codefromthecrypt: normally I would, but some of these are used in templates and don't want to clean that up right now if there's a reference miss. I'll add a TODO

codefromthecrypt:
Thanks for the review @mathetake

codefromthecrypt:
raised https://github.com/tetratelabs/getenvoy/issues/197 to de-abstract docker because I started to try that and it became a beast to add here

codefromthecrypt:
green! One thing I did to help with this is to use the admin-address file directly instead of parsing stdout (a buffered stream)

we flaked before not because of timeout but more likely due to buffering, as the process certainly didn't take a minute to start.

mathetake (approved):
Awesome work! I've tested this patch on my local build and it works as expected 😄

codefromthecrypt:
Thanks for the couple days reviewing this @mathetake (also the preceding 6 days of counseling leading to it)

codefromthecrypt:
#232 continues this

Issue #109, opened 2020-11-30
Can we have ARM64 binary support please?


lhaig:
#Environment

## Hardware 
TuringPi Cluster with Raspberry PI 3+ compute modules.
## Software 
Ubuntu 18.04 and 20.04
## Error

curl -L https://getenvoy.io/cli | sudo bash -s -- -b /usr/local/bin
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    93  100    93    0     0    123      0 --:--:-- --:--:-- --:--:--   123
100  9995  100  9995    0     0  11383      0 --:--:-- --:--:-- --:--:-- 11383
tetratelabs/getenvoy crit platform linux/arm64 is not supported.  Make sure this script is up-to-date and file request at https://github.com/tetratelabs/getenvoy/issues/new

## Requested feature

Please provide arm64 support for the getenvoy application.

cetanu:
Linking #84 #82 

Pretty keen for this too

cetanu:
If anyone knows how to run the pipeline for this project in order to produce an arm64 binary, or knows the barriers to doing so, I would be interested to hear them so that I can look into making it happen.

cetanu:
Actually I was able to check out this repo and build it for arm64, but this is the repo for the `getenvoy` utility itself.

I think this is the more appropriate repo: https://github.com/tetratelabs/getenvoy-package

lhaig:
Am happy to assist with creating a multi ARCH build script if this is what is needed. I use Consul in my demo's and it uses Envoy and I would like to have the ability to install and use it there if possible.

codefromthecrypt:
one thing we'd need to do with this is understand the impact for testing. We are currently using GitHub Actions for testing via e2e, and I'm not sure if it supports linux/arm64 yet (In other projects I had to use travis to test this for real and when can't use emulation which isn't great)

musaprg:
@codefromthecrypt Currently GitHub Actions(GHA) doesn't support linux/arm64 for now. One way is to prepare ARM instances on AWS as GHA self-hosted runner, and now I'm working on it.

codefromthecrypt:
[...]

PR #198, opened 2021-04-29
Add the target architecture field to the manifest.proto

musaprg:
related: https://github.com/tetratelabs/getenvoy-package/pull/109

Add the target architecture of envoy builds to the manifest.proto. These changes are related to getenvoy arm support tetratelabs/getenvoy#109.


api/manifest.proto
```diff
@@ -67,6 +67,16 @@ message Build {
       LINUX_GLIBC = 3;
    }
 
+   // Envoy currently supports [aarch64/arm64, s390x, mips64, ppc, x86_64] as build targets.
+   // See the envoyproxy/envoy repository for more details.
+   // https://github.com/envoyproxy/envoy/blob/main/bazel/BUILD
+   enum Architecture {
+        UNKNOWN_ARCH = 0;
+        X86_64 = 1;
+        ARM64 = 2;
```
musaprg, line 76: According to [this letter](https://lists.linaro.org/pipermail/linaro-toolchain/2014-April/004244.html), `aarch64` seems to be appropriate, but Go calls it as `arm64`. I'm not sure which is better.

```
$ go tool dist list | grep linux                                                                                                                 
linux/386
linux/amd64
linux/arm
linux/arm64
linux/mips
linux/mips64
linux/mips64le
linux/mipsle
linux/ppc64
linux/ppc64le
linux/riscv64
linux/s390x
```

codefromthecrypt:
We should merge this soon as it will break behavior and we've already update the manifest URL to a different source. Meaning merging this won't break old clients, because they are still on bintray. However, if we released our URL update already in master, then later added arch, we would break clients due to ambiguity on which URL to choose.

api/manifest.proto
codefromthecrypt: @musaprg since no one is responding maybe you can find someone who doesn't work here but works on envoy or maybe @mathetake can help. Main thing is arch is relative, and I suspect we want to optimize for the name envoy would use when doing builds, not necessarily what GOARCH says. It is helpful though it they are the same or it doesn't matter.

codefromthecrypt:
if we release before merging this, we should probably mention in the release notes that a subsequent release may change the file system layout used to download binaries (to allow fetching multiple archs of the same platform). I think that's fine as this is <1.0 anyway

api/manifest.proto
codefromthecrypt: for arm64, I think we should settle on what envoy refers in the docs: arm64
They never say aarch64, yet, so we should be safe

https://www.envoyproxy.io/docs/envoy/latest/search.html?q=arm64&check_keywords=yes&area=default#

counter data point: I noticed the buf project, just now changed their tooling to match linux (aarch64). Any choice we make will have a counter point, but I think knowing who to follow is best. We use upstream conventions first!

https://github.com/bufbuild/buf/releases/tag/v0.42.1
> Change the architecture suffix of the Linux ARM release assets from arm64 to aarch64 to match the output of uname -m on Linux.

codefromthecrypt: sheesh problem though, envoy uses the terms depending on OS and also sometimes changes the config name to have a suffix of CPU and other times not.

This is a problem for us when it is missing though we can guess amd64 (aka x86_64) in that case. The other thing we can do is simply accept either. For example, even if they choose aarch when refering to linux and arm when refering to darwin, it isn't the case they choose both for one OS. We can poke for either!

Meanwhile as this problem should bubble up into docker anyway we can also look at how they build their stuff.

https://github.com/envoyproxy/envoy/blob/main/bazel/BUILD

```
selects.config_setting_group(
    name = "apple",
    match_any = [
        ":darwin",
        ":darwin_x86_64",
        ":ios_arm64",
        ":ios_arm64e",
        ":ios_armv7",
        ":ios_armv7s",
        ":ios_i386",
        ":ios_x86_64",
    ],
)

selects.config_setting_group(
    name = "linux",
    match_any = [
        ":linux_aarch64",
        ":linux_mips64",
        ":linux_ppc",
        ":linux_s390x",
        ":linux_x86_64",
    ],
)
```

codefromthecrypt:
obviated by the new json format

PR #199, opened 2021-04-29
Fix not to handle error when cleaning under `api/`

musaprg:
Signed-off-by: Kotaro Inoue <k.musaino@gmail.com>


codefromthecrypt (approved):

codefromthecrypt:
thanks for fixing a broken window!

PR #200, opened 2021-05-06
Removes all extension code

codefromthecrypt:
A myriad of maintenance issues and conversations led to the summary
outcome that the "extension" subsystem is not a right fit. Rather than
continue weeks of refactoring and otherwise quality corrections, this
removes all extension (WASM) features.

The time otherwise spent on the >14k lines of code and configuration
can go towards improving getenvoy. Improving getenvoy will be far
easier when it isn't subject to hour turn-arounds on rust builds or
the complexity of embedding Docker.

Instead, we will suggest developers use the underlying SDKs and their
corresponding developer workflows. Over time, we may tighten integration
here with WASM in less coupled ways, for example integrating existing
E2E tests from TinyGo is far simpler than embedding builds, and easier
on CI as well.

Meanwhile, the supported way to run any extension (Lua or WASM), is to
pass a configuration file to the `getenvoy run` command.


codefromthecrypt:
note: I've inlined the e2e tests in the same container as unit tests now that they aren't as heavy anymore. This is cheaper in all ways than what we did before.

codefromthecrypt:
Also, as expected the binary is dramatically smaller again, down a third

```bash
du -k build/bin/*/*
13544	build/bin/darwin/amd64
12552	build/bin/linux/amd64
```

codefromthecrypt:
FYI, after this, there are a total of 9335 lines left in the project

mathetake:
maybe https://github.com/tetratelabs/getenvoy/blob/master/.dockerignore can also be removed? looks like it is for extension-builders only.

codefromthecrypt:
> maybe https://github.com/tetratelabs/getenvoy/blob/master/.dockerignore can also be removed? looks like it is for extension-builders only.

actually I think that ignore file is for the getenvoy docker image (/Dockerfile), though I'm not fully aware of who uses it...

codefromthecrypt:
It is ironic that the docker for mac thing is only now not installing :D

codefromthecrypt:
for E2E tests on mac the only thing that needs docker is the docker registry in docker-compose.yml  (docker image "registry:2" source https://github.com/distribution/distribution)

this is needed for the push tests as are intentionally not using a mocked registry for this test.. I'll see if there's a github action or otherwise solution for a temporary OCI registry as it just needs to work for one test. cc @arkodg if you know

pkg/cmd/extension/push.go
```diff
@@ -0,0 +1,117 @@
...
+	opts := &cmdOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "push /path/to/extension.wasm <image-reference>",
+		Short: "Push the WASM extension to the OCI-compliant registry.",
+		Long: `
+Push the WASM extension to the OCI-compliant registry. This command requires to login the target container registry with docker CLI`,
+		Example: `
+  # Push WASM extension to the local docker registry.
+  getenvoy extension push /path/to/extension.wasm localhost:5000/test/image-name:tag`,
```
codefromthecrypt, line 80: wonder if the order $path $ref vs $ref $path

Screen Shot 2021-05-07 at 8 34 20 AM

arkodg:
took a peek at your GHA file, you didnt make any changes to the `docker-practice/actions-setup-docker@v1` yet it doesnt start 😞 

maybe add some more knobs like
```
           docker_channel: stable
            docker_version: "20.10"
 ```
 https://github.com/docker-practice/actions-setup-docker/blob/master/.github/workflows/ci.yaml#L33
 
 unaware of a temp registry

File: pkg/cmd/extension/push.go (hunk 1 above)
Comment (codefromthecrypt) [reply pkg/cmd/extension/push.go:80]: @mathetake does embedding oras still make sense? push/pull is basically a wrapper over oras which has more options likely needed for pushing to other registries..

Comment (codefromthecrypt) [reply pkg/cmd/extension/push.go:80]: maybe in a later PR we can decide to keep or axe this. In an example, we can show oras alternatives to push/pull for wasm https://github.com/deislabs/orasOpensearch

**Comment by codefromthecrypt** [comment]
@mathetake I think we should remove the oras code as it decouples us completely. Ex this works using https://github.com/tetratelabs/proxy-wasm-go-sdk


```bash
$ make build.example name=access_logger
$ oras push localhost:5000/access_logger:dev \
  --manifest-config /dev/null:application/vnd.module.wasm.config.v1+json \
  ./examples/access_logger/main.go.wasm:application/vnd.module.wasm.content.layer.v1+wasm

# later you can pull like this..
$ oras pull --verbose localhost:5000/access_logger:dev -a
```

codefromthecrypt:
based on chat with @peterj and @mathetake I think we should remove push also. If users ask for it, we can rebuild a more relevant subcommand "getenvoy wasm push|pull" later, being mindful that repackaging oras can be tricky if anything is authenticated as we can (as we do now) easily block config helpers that oras already contains, like reading normal docker config.

mathetake:
also note that we even might not need oras at all since the Wasm OCI spec has changed from the one in this repo (using custom layer type) to be compatible with any registry by using existing well-known media type: https://docs.google.com/document/d/1Uf-3RPFRyNl6ExYUeZERlB9Y8kyxZFaB6NbDZyguXwI/edit

so for example https://github.com/google/go-containerregistry can be used. But please keep in mind that the spec might change in the future again. Stay tuned 😄

codefromthecrypt:
thanks for the tips @mathetake all the more reason to not have this feature now :D

codefromthecrypt:
opened this issue https://github.com/tetratelabs/getenvoy/issues/201

the next change to this PR will remove the push code and in doing so remove the need to install docker on osx

codefromthecrypt:
lines of code config and docs down to under 6k, binary down also

```
du -sk build/bin/darwin/amd64/getenvoy 
12728	build/bin/darwin/amd64/getenvoy
```

DEVELOPER.md
```diff
@@ -30,29 +21,16 @@ make test
...
-
-The below ENV variables to effect the e2e execution. These are defined in [main_test.go](test/e2e/main_test.go).
-Environment Variable              | Description
---------------------------------- | ------------------------------------------------------------------------------------
-`E2E_GETENVOY_BINARY`             | Overrides `getenvoy` binary. Defaults to `$PWD/build/bin/$GOOS/$GOARCH/getenvoy`
-`E2E_TOOLCHAIN_CONTAINER_OPTIONS` | Overrides `--toolchain-container-options` in Docker commands. Defaults to "".
-`E2E_EXTENSION_LANGUAGE`          | Overrides `--language` in `getenvoy extension` commands. Defaults to "all".
+Run:
+```shell
+make coverage
```
mathetake, line 35: looks like `make coverage` appears twice  😄

codefromthecrypt: hehe whoops

go.mod
```diff
@@ -7,25 +7,15 @@ require (
...
-	github.com/deislabs/oras v0.11.1
-	github.com/envoyproxy/go-control-plane v0.9.9-0.20201210154907-fd9021fe5dad
-	github.com/manifoldco/promptui v0.8.0
-	github.com/mattn/go-isatty v0.0.12
 	github.com/mattn/go-shellwords v1.0.11
 	github.com/mholt/archiver/v3 v3.5.0
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/opencontainers/image-spec v1.0.1
-	github.com/otiai10/copy v1.5.1
+	github.com/pkg/errors v0.9.1 // indirect
```
mathetake, line 13: wondering why pkg/errors comes up again?

codefromthecrypt: this is our logging library and we can remove it in another PR as it isn't needed to use an abstraction when you can't even configure it.

pkg/cmd/run_test.go
```diff
@@ -0,0 +1,143 @@
...
+			// Verify the command failed with the expected error
+			require.EqualError(t, err, test.expectedErr, `expected an error running [%v]`, c)
+			require.Empty(t, stdout.String(), `expected no stdout running [%v]`, c)
+			expectedStderr := fmt.Sprintf("Error: %s\n\nRun 'getenvoy run --help' for usage.\n", test.expectedErr)
+			require.Equal(t, expectedStderr, stderr.String(), `expected stderr running [%v]`, c)
+		})
+	}
+}
+
+func TestGetEnvoyRun(t *testing.T) {
```
mathetake, line 66: just curious are these following tests new ones in this PR? (`TestGetEnvoyRun` and `TestGetEnvoyRunFailWithUnknownVersion`)

codefromthecrypt: yeah I noticed that "extension run" had a couple better tests than our normal "getenvoy run" so moved them to reclaim the effort we spent writing them

mathetake:
this is irrelevant to this PR but looks like `TestSetupSignalHandlerIgnoresWhenContextCanceledBeforeSignal` is flaky somehow https://github.com/tetratelabs/getenvoy/runs/2524038393

mathetake (approved):

codefromthecrypt:
agree @mathetake https://github.com/tetratelabs/getenvoy/issues/172 mentions the root problem

codefromthecrypt:
thanks for the review @mathetake and the patience. I know you spent last 2 months refactoring all of this with me, but this is a better call than keeping it around just because we spent time on it. 🙇

PR #203, opened 2021-05-07
Removes non-standard flavors

codefromthecrypt:
As we prepare for upstream builds, non-standard flavors become a
liability. This removes them, but tolerates when they are present. In
practice, we only ever published one-off flavors for postgres and wasm.


codefromthecrypt:
as expected this knocks another 120k off the binary

Dockerfile
```diff
@@ -17,8 +17,9 @@ FROM gcr.io/distroless/cc
 ARG getenvoy_binary=getenvoy
 COPY ${getenvoy_binary} /
 
+# TODO: delete this when we build in default version
 ARG reference
```
codefromthecrypt, line 21: this is effectively a build arg. We won't need this anymore when we choose latest version. I'll definitely do this and soon

api/manifest.proto
```diff
@@ -20,33 +20,14 @@ option go_package = "github.com/tetratelabs/getenvoy/api;api";
 message Manifest {
    string manifest_version = 1;
 
-   // Key is the flavor name
+   // Flavor is always "standard" for historical reasons
```
codefromthecrypt, line 23: Adding a note rather than break the proto.

Later, we can refactor the proto. I think the json source for this proto and where it is hosted may always be in our repo frankly, so we should be able to break compat once we figure out the stable shape that works with upstream hosted builds. In any case we can keep compat for now cc @musaprg

pkg/cmd/run.go
```diff
@@ -46,35 +43,13 @@ Envoy state and machine state into the ` + "`~/.getenvoy/debug`" + ` directory.`
 		Example: fmt.Sprintf(`# Run using a manifest reference.
 getenvoy run %[1]s -- --config-path ./bootstrap.yaml
 
-# Run using a filepath.
-getenvoy run /usr/local/bin/envoy -- --config-path ./bootstrap.yaml
```
codefromthecrypt, line 50: old fuzz should have been deleted earlier

pkg/manifest/locate.go
```diff
@@ -28,21 +27,13 @@ import (
...
 )
 
-const (
-	referenceEnv = "ENVOY_REFERENCE"
-)
-
 // NewKey creates a manifest key based on the reference it is given
 func NewKey(reference string) (*Key, error) {
-	// This enables us to parameterize Docker images
-	if reference == "@" {
```
codefromthecrypt, line 38: deleted this to increase test coverage and in any case it was super weird

pkg/manifest/locate.go
```diff
@@ -75,23 +66,31 @@ func FetchManifest(manifestURL string) (*api.Manifest, error) {
 		return nil, fmt.Errorf("error reading %s: %w", manifestURL, err)
 	}
 
+	// Ignore deprecated fields relating to flavor
+	options := protojson.UnmarshalOptions{DiscardUnknown: true}
```
codefromthecrypt, line 70: this is how we don't crash on unknown

pkg/manifest/locate_test.go
```diff
@@ -75,6 +70,46 @@ func TestFetchManifest(t *testing.T) {
 	goodManifestBytes, err := protojson.Marshal(goodManifest)
 	require.NoError(t, err)
 
+	// this includes fields we removed like filterProfile and name
+	manifestBytesWithExtraFields := []byte(`{
```
codefromthecrypt, line 74: this is the test ensuring we don't crash on unknown

pkg/types/types.go
```diff
@@ -36,21 +35,24 @@ func PlatformFromEnum(s string) string {
...
-	if len(matches) != 4 {
-		return nil, fmt.Errorf("%q is not a valid GetEnvoy reference. Expected format: <flavor>:<version>[/<platform>]", text)
+	if len(matches) == 4 { // only allow historical naming convention when the "flavor" is "standard"
+		flavor := strings.ToLower(matches[1])
+		if flavor == "" || flavor == "standard:" {
+			return &Reference{strings.ToLower(matches[2]), PlatformFromEnum(matches[3])}, nil
+		}
 	}
-	return &Reference{strings.ToLower(matches[1]), strings.ToLower(matches[2]), PlatformFromEnum(matches[3])}, nil
+	return nil, fmt.Errorf("%q is not a valid GetEnvoy reference. Expected format: <version>[/<platform>]", text)
```
codefromthecrypt, line 50: not splitting hairs here.. if you either use a non-standard old-school flavor or mess up the syntax, we err. To be compat with old references, tolerating "standard:1.17.1" style syntax even if we don't care about the standard part

test/e2e/getenvoy_test.go
```diff
@@ -30,6 +30,16 @@ func TestGetEnvoyVersion(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestGetEnvoyList(t *testing.T) {
+	t.Parallel()
+
+	stdout, stderr, err := getEnvoy("list").Exec()
```
codefromthecrypt, line 36: never tested this before. This helps make sure we don't goof json parsing in real life data

codefromthecrypt (commented):
impl notes

mathetake:
one thing I want to mention is that even in upstream build there would be two flavors: standard and FIPS compliant. https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/ssl#fips-140-2 

I'm ok with this change to proto but want to make sure we could support both standard and FIPS in the future

codefromthecrypt:
@mathetake sounds good. for now we aren't building fips anyway and we can re-introduce it when the need arises, Ideally from some site saying they want it. I don't see a big problem in re-introducing anything including this, but we would have a hard time speculating how, when, and if upstream would build fips, so temporarily de-cluttering seems the choice of the day. sg?

mathetake:
sg, thanks!

lizan (changes requested):
While we plan upstream the builds, the manifest belongs to CLI and it will be us to publish it. So how those become liability? The flavors also reflects the desire from https://github.com/envoyproxy/envoy/issues/7582.

codefromthecrypt:
@lizan the only use of non-standard flavors were one-off builds made by us, correct? Also, as mentioned before, the structure in way flavors would be handled upstream, to the degree they are, would need to be decided. That issue you mentioned was from 2019 and so far no decision.. We can publish an update later, but this feature was premature and caused UX fail such as everyone having to put "standard:1.17.2" in their version names even if there was effectively no alternative. If we re-introduce flavors later (when they are relevant and published), we can update the UI

codefromthecrypt:
as sometimes discussions get theoretical, please look at the json we use now and you'll see what I mean. https://dl.getenvoy.io/public/raw/files/manifest.json

lizan:
> you can also see in getenvoy list. There are two ad-hoc "flavors" which even when named nightly weren't as you can see they are pinned to old envoy

> Even if we control the manifest, if there were alternate flavors, they would need to be published by someone. That's the relevance of this going upstream. If even we didn't do flavors, we shouldn't kite upstream doing them. If the point is for us to stop doing builds, we would only reference things upstream. Having a speculative construct that adds cognitive and code complexity is a liability.

"wasm" and "postgres" served its purpose exactly, we need the freedom of serving those flavors to let users to try out new features, for example the wasm flavor exists before they merged in upstream, same as postgres. While today they both merged into upstream so we only have "standard", but it doesn't mean we won't have flavors anymore. We don't want to update protos every time we need to do that?

codefromthecrypt:
> "wasm" and "postgres" served its purpose exactly, we need the freedom of serving those flavors to let users to try out new features, for example the wasm flavor exists before they merged in upstream, same as postgres. While today they both merged into upstream so we only have "standard", but it doesn't mean we won't have flavors anymore. We don't want to update protos every time we need to do that?

These experiments were made by Tetrate staff and resulted in builds also managed by Tetrate staff. There's no evidence end users were involved and if there were, we would have seen complaints from abandoned builds.

There is no evidence of ad-hoc usage from end users and they couldn't anyway unless they override the json endpoint, because they would also need to supply the download URL and that's in the same endpoint. If they can override the endpoint, then can make up a version scheme like 1.17.2-experiment and add it to their custom list, or make an entirely different endpoint like https://site/fips/manifest.json

codefromthecrypt:
To keep progress moving, I'll first fix the major UX version while retaining the "spare parts" in the manifest proto.

Basically, this:
* stop requiring "standard:" when indicating versions, assume standard when missing.
* prevent crashing when json has more fields than the proto defined (this also prevents crashing when we add arch later)

I'll follow-up later around the manifest in general and we can refer back to this at that point for context.

codefromthecrypt:
@mathetake PTAL when you get a chance

pkg/manifest/platform_linux.go
```diff
@@ -14,6 +14,9 @@
 
 package manifest
 
-func platform() string {
+// CurrentPlatform is the value of Reference.Platform for this GOOS.
+// Specifically, when "getenvoy" runs and detects GOOS == "linux", we choose the "linux-glibc" Envoy platform.
```
codefromthecrypt, line 18: this broke my brain. I'm not sure the history of this choice, but in any case it will probably chance when upstream does. For now, this comment can hopefully save the next person.

pkg/manifest/print_test.go
```diff
@@ -28,11 +28,11 @@ func TestPrint(t *testing.T) {
...
 
 	require.NoError(t, err)
-	require.Equal(t, `REFERENCE                                FLAVOR                VERSION
-standard:nightly/linux-glibc             standard              nightly
-standard:1.17.1/linux-glibc              standard              1.17.1
-standard:1.17.1/darwin                   standard              1.17.1
-standard-fips1402:1.10.0/linux-glibc     standard-fips1402     1.10.0
+	require.Equal(t, `REFERENCE                                VERSION
+nightly/linux-glibc                      nightly
+1.17.1/linux-glibc                       1.17.1
```
codefromthecrypt, line 33: note: we now suggest simpler references in list output

pkg/test/manifest/server.go
```diff
@@ -30,16 +30,16 @@ import (
...
 	"github.com/tetratelabs/getenvoy/api"
-	"github.com/tetratelabs/getenvoy/pkg/types"
+	"github.com/tetratelabs/getenvoy/pkg/manifest"
 )
 
 const archiveFormat = ".tar.gz"
 
 // RequireManifestTestServer serves "/manifest.json", which contains download links a compressed archive of artifactDir
-func RequireManifestTestServer(t *testing.T, manifest *api.Manifest) *httptest.Server {
+func RequireManifestTestServer(t *testing.T, m *api.Manifest) *httptest.Server {
```
codefromthecrypt, line 39: dodges import clash

mathetake:
> prevents crashing on manifest updates 

Do you mean "with this patch the cli would never crash even when we remove flavors in manifest"?

pkg/cmd/run.go
```diff
@@ -46,35 +43,13 @@ Envoy state and machine state into the ` + "`~/.getenvoy/debug`" + ` directory.`
...
-
 # List available Envoy flags.
 getenvoy run %[1]s -- --help
-
-# Run with Postgres specific configuration bootstrapped
-getenvoy run postgres:nightly --templateArg endpoints=127.0.0.1:5432,192.168.0.101:5432 --templateArg inport=5555
 `, defaultreference.Latest),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := InitializeRunOpts(o, args[0]); err != nil {
```
mathetake, line 50: guess this panics when users execute `getenvoy run` (without args).

pkg/manifest/locate.go
mathetake: +1

pkg/manifest/reference.go
```diff
@@ -36,19 +36,39 @@ func PlatformFromEnum(s string) string {
...
 	}
-	return &Reference{strings.ToLower(matches[1]), strings.ToLower(matches[2]), PlatformFromEnum(matches[3])}, nil
+	flavor := strings.ToLower(matches[1])
+	if flavor == "" {
+		flavor = "standard"
+	} else {
+		flavor = flavor[:len(flavor)-1]
+	}
+	version := strings.ToLower(matches[2])
+	platform := PlatformFromEnum(matches[3])
```
mathetake, line 55: I see that `PlatformFromEnum` should be unexported 😄

pkg/manifest/reference.go
```diff
@@ -36,19 +36,39 @@ func PlatformFromEnum(s string) string {
...
+		platform = CurrentPlatform()
+	}
+	return &Reference{flavor, version, platform}, nil
 }
 
+// String returns a plain version optionally platform qualified. Ex "1.17.2" or "1.17.2/darwin"
+// Reference.Flavor is prefixed with a colon when non-standard. Ex "experimental:1.18.2"
 func (r *Reference) String() string {
+	if r.Flavor == "standard" { // don't clutter output in normal case
+		if r.Platform == "" {
```
mathetake, line 67: when does platform becomes empty? I think that *Reference is only created via ParseReference which ensures Platform being not empty..? (I guess this is only the case of unit tests).

Dockerfile
codefromthecrypt: next PR will kill that image

pkg/cmd/run.go
codefromthecrypt: yeah I want to swap out the cli lib, but I'll fix this here and now meanwhile.

pkg/manifest/reference.go
```diff
@@ -36,19 +36,39 @@ func PlatformFromEnum(s string) string {
...
+		return nil, fmt.Errorf("%q is not a valid GetEnvoy reference. Expected format: [<flavor>:]<version>[/<platform>]", text)
 	}
-	return &Reference{strings.ToLower(matches[1]), strings.ToLower(matches[2]), PlatformFromEnum(matches[3])}, nil
+	flavor := strings.ToLower(matches[1])
+	if flavor == "" {
+		flavor = "standard"
+	} else {
+		flavor = flavor[:len(flavor)-1]
+	}
+	version := strings.ToLower(matches[2])
```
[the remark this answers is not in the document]
codefromthecrypt: oh wait.. yeah we don't enforce it is a number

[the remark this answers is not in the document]
codefromthecrypt: making a comment

pkg/manifest/reference.go
codefromthecrypt: yep especially as we now have one package we don't need this exported

pkg/manifest/reference.go
codefromthecrypt: correct I'll remove it

mathetake (approved):

codefromthecrypt:
#232 continues this

PR #209, opened 2021-05-11
Stops admin fetching on first error

codefromthecrypt:
Before, we added a dependency on multi-error to allow multiple problems
to accumulate before before returning control to the user. There's
little reason to allow partial failure across admin URLS, especially
when a partial failure still quits with an error status.

This removes the dependency and works normally instead. This also fixes
syntax where formerly we used defer in a for loop.


pkg/binary/envoy/debug/admin.go
```diff
@@ -54,28 +52,36 @@ func (e *envoyAdminDataCollection) retrieveAdminAPIData() error {
...
+	if err != nil {
+		return fmt.Errorf("could not read %v: %w", url, err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("received %v from %v", resp.StatusCode, url)
+	}
+	defer resp.Body.Close() //nolint
+
+	if _, err := io.Copy(f, resp.Body); err != nil {
+		return fmt.Errorf("could write response body of %v: %w", url, err)
```
codefromthecrypt, line 84:
```suggestion
		return fmt.Errorf("could not write response body of %v: %w", url, err)
```

mathetake (approved):

Issue #173, opened 2021-04-11
Incorrect use of Pdeathsig for killing child on linux.


codefromthecrypt:
**Describe the bug**
Right now, on Linux, we set `SysProcAttr.Pdeathsig = SIGTERM` with a comment that this will kill the child when the parent dies. This is incorrect usage to achieve the results. The correct way to use Pdeathsig would be to launch a goroutine, lock it to the OS thread, start the child in that goroutine, and don't exit the goroutine until the child exits. 

While revisiting this, we should re-check any assumptions for example that SIGTERM is the appropriate signal to catch. 

Personally, I think we should prefer behaviour mentioned in the relevant go issue on conflict as many signal related code have had problems here.

**Additional context**
See https://github.com/golang/go/issues/27505#issuecomment-713706104 for an example
See https://github.com/rootless-containers/rootlesskit/issues/65 for a discussion leading to this

dio:
We haven't done the suggested solution specifically, since the experiment in https://gist.github.com/dio/85d51d0c199b7598cefa56cb95521537 shows: that it only matters for many children processes. However, we have added tests for making sure when killing `func-e` on Linux, we propagate SIGKILL to `envoy`. See: https://github.com/tetratelabs/func-e/pull/372.

codefromthecrypt:
works for me. thanks for the research and if something changes in the future (like somehow we start a million envoys), we can reference this!

PR #211, opened 2021-05-11
Removes utility code formerly used by Docker

### LINKED ISSUE #172: Refactor out osutil.SetupSignalHandler

codefromthecrypt:
We had a lot of redundant code that was deleted with the extension
subcommand. However, code analysis failed to notice other parts that are
only referenced in their own tests. This completes removal of the extra
code made for the extension experiment.

Fixes #172 #173


pkg/cmd/root.go
```diff
@@ -106,15 +105,10 @@ func handleFlagOverrides(o *globals.GlobalOpts, homeDirFlag, manifestURLFlag str
 func Execute(cmd *cobra.Command) error {
 	actualCmd, err := cmd.ExecuteC()
 	if actualCmd != nil && err != nil { // both are always true on error
-		var serr exec.ShutdownError
```
codefromthecrypt, line 109: only dead code used this type

pkg/cmd/run.go
```diff
@@ -95,11 +94,8 @@ func InitializeRunOpts(o *globals.GlobalOpts, reference string) error {
 // This is exposed for re-use in "getenvoy extension run"
 func Run(o *globals.GlobalOpts, cmd *cobra.Command, args []string) error {
 	r := envoy.NewRuntime(&o.RunOpts)
-	r.IO = ioutil.StdStreams{
-		In:  cmd.InOrStdin(),
```
codefromthecrypt, line 99: in was never used

mathetake (approved):

PR #212, opened 2021-05-12
Refactors logging to not bother users on expected errors

codefromthecrypt:
Before, we had debug logs that were not possible to enable, as well logs
that bothered users with situations they could not action (such as vague
details about child process signals.). This stops logging on aspects the
user cannot control. It tightens logging so that if there's something we
don't expect, the user will see it.

For example, we know certain OS utilities are unimplemented, so we skip
writing files or fields when that happens. However, we don't bother
users. If there is something unexpected, it is logged at most once per
debug utility. We do not log anything anymore about a possibility of
someone doing "kill -9" and leaving the child envoy around. This is pure
distraction for most people.


codefromthecrypt:
Merging this increases statement coverage from 82.3% to 84.2%

pkg/binary/envoy/debug/admin.go
```diff
@@ -36,10 +36,11 @@ var adminAPIPaths = map[string]string{
 	"runtime":           "runtime.json",
 }
 
-// EnableEnvoyAdminDataCollection is a preset option that registers collection of Envoy Admin API information
-func EnableEnvoyAdminDataCollection(r *envoy.Runtime) {
+// enableEnvoyAdminDataCollection is a preset option that registers collection of Envoy Admin API information
+func enableEnvoyAdminDataCollection(r *envoy.Runtime) error {
```
codefromthecrypt, line 40: unexported this and similar as they didn't need to be exported

pkg/binary/envoy/debug/lsof.go
```diff
@@ -63,15 +61,7 @@ type openFilesDataCollection struct {
 }
 
 // retrieveOpenFilesData writes statistics of open files associated with envoy instance(s) to a json file
-// Errors from platform-specific libraries log to debug instead of raising an error or logging in an error category.
-// This avoids filling logs for unresolvable reasons.
 func (o *openFilesDataCollection) retrieveOpenFilesData() error {
-	f, err := os.Create(filepath.Join(o.lsofDir, "lsof.json"))
```
codefromthecrypt, line 69: re-organized to write file at the bottom, as doing so avoids writing an empty file

pkg/binary/envoy/debug/node.go
```diff
@@ -65,34 +63,25 @@ func (n *nodeCollection) ps() error {
...
-	w := tabwriter.NewWriter(out, 0, 8, 5, ' ', 0)
-	fmt.Fprintln(w, "PID\tUSERNAME\tSTATUS\tRSS\tVSZ\tMINFLT\tMAJFLT\tPCPU\tPMEM\tARGS")
-	for _, p := range processes {
-		proc := safeProc(ctx, p)
-		if proc == empty { // Ignore, but continue. The process could have died between the process listing and now.
-			continue
+	parsed, err := parseProcessTable(ctx, processes)
+	if len(parsed) == 0 {
+		if err != nil {
+			return fmt.Errorf("unable to parse any processes: %w", err)
```
codefromthecrypt, line 79: only looking at the error when we parsed nothing helps us allow partial process table listing (ex columns unsupported to be blank)

pkg/binary/envoy/debug/node.go
```diff
@@ -158,38 +163,49 @@ func safeProc(ctx context.Context, p *process.Process) *proc {
...
-	out, err := json.Marshal(is)
-	if err != nil {
-		return fmt.Errorf("unable to convert to json representation: %v", err)
+	if len(result) == 0 {
+		return nil // don't write a file on an unsupported platform
 	}
-	fmt.Fprintln(f, string(out))
-
-	return nil
+	return writeJSON(result, filepath.Join(n.nodeDir, "network_interface.json"))
```
codefromthecrypt, line 208: it increased coverage to extract this and use in the 3 places we write json

pkg/binary/envoy/fetch.go
```diff
@@ -59,7 +58,6 @@ func FetchIfNeeded(o *globals.GlobalOpts, reference string) (string, error) {
 			return "", fmt.Errorf("unable to create directory %q: %w", platformDirectory, err)
 		}
 
-		log.Debugf("fetching %v from %v", ref, binaryLocation)
```
codefromthecrypt, line 62: these debug statements weren't that useful as they couldn't be toggled anyway.

pkg/binary/envoy/procattr.go
```diff
@@ -18,13 +18,10 @@ package envoy
 
 import (
 	"syscall"
-
-	"github.com/tetratelabs/log"
 )
 
 func sysProcAttr() *syscall.SysProcAttr {
-	log.Warn("Running on a non-Linux system, cannot set child process kill signal. " +
```
codefromthecrypt, line 26: this was the most unhelpful log

pkg/binary/envoytest/util.go
```diff
@@ -41,7 +41,8 @@ func RunAndTerminateWithDebug(t *testing.T, debugDir string, debug func(r *envoy
 
 	r := envoy.NewRuntime(o)
 
-	debug(r)
+	e := debug(r)
+	require.NoError(t, e)
```
codefromthecrypt, line 45: this ensures we didn't miss anything

codefromthecrypt (commented):
notes

pkg/binary/envoy/debug/admin.go
```diff
@@ -36,10 +36,11 @@ var adminAPIPaths = map[string]string{
...
 	e := envoyAdminDataCollection{r.GetAdminAddress, r.GetWorkingDir()}
 	r.RegisterPreTermination(e.retrieveAdminAPIData)
+	return nil
```
mathetake, line 43: looks like we always return `error` so can we remove?

mathetake: ah - nvm.

pkg/binary/envoy/debug/lsof.go
```diff
@@ -83,26 +73,25 @@ func (o *openFilesDataCollection) retrieveOpenFilesData() error {
...
 	}
 
 	result := make([]OpenFileStat, 0)
-	// print open file stats for all envoy instances
-	// relevant fields of the process
+	// print open file stats for all envoy relevant fields of the process
 	username, err := envoyProcess.UsernameWithContext(ctx)
-	if logDebugOnError(ctx, err, "username", pid) {
-		return nil
+	if w := wrapError(ctx, err, "username", pid); w != nil {
```
mathetake, line 83: wrapError is defined in node.go but used both in lsof.go and node.go which seems weird to me..

pkg/binary/envoy/debug/lsof.go
```diff
@@ -126,8 +115,8 @@ func (o *openFilesDataCollection) retrieveOpenFilesData() error {
 		result = append(result, ofStat)
 	}
 
-	if err = json.NewEncoder(f).Encode(result); err != nil {
-		return fmt.Errorf("error writing JSON to file %v: %w", f, err)
+	if len(result) == 0 {
+		return nil // don't write a file on an unsupported platform
 	}
-	return nil
+	return writeJSON(result, filepath.Join(o.lsofDir, "lsof.json"))
```
mathetake, line 121: same goes for writeJSON

pkg/binary/envoy/debug/lsof.go
codefromthecrypt: I can extract them into a different file

mathetake: sg

pkg/binary/envoy/debug/debug.go
```diff
@@ -0,0 +1,60 @@
+// Copyright 2020 Tetrate
```
codefromthecrypt, line 1:
```suggestion
// Copyright 2021 Tetrate
```

codefromthecrypt:
@mathetake I noticed we only used tetrate logging in a couple places after this, so went ahead and used standard logging for the same. deps are a lot clearer now.

mathetake (approved):

codefromthecrypt:

PR #213, opened 2021-05-12
Prefers quitting vs using an incoherent homedir

codefromthecrypt:
Recent refactoring deferred lookup of the user home dir until after
GETENVOY_HOME is checked. This prefers being strict vs using an
undefined value on error determining the user home.

### DIFF
--- a/pkg/globals/globals.go
+++ b/pkg/globals/globals.go
@@ -14,13 +14,6 @@
 
 package globals
 
-import (
-	"path/filepath"
-
-	"github.com/mitchellh/go-homedir"
-	"github.com/tetratelabs/log"
-)
-
 // RunOpts support invocations of "getenvoy run" and "getenvoy extension run"
 type RunOpts struct {
 	// EnvoyPath is the exec.Cmd path to "envoy". Defaults to "$HomeDir/builds/$flavor/$version/$platform/bin/envoy"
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
code coverage is the same before or after this. I didn't try to fake `homedir.Dir()` as it is messy to do that. Also, a future CLI lib may make this unnecessary. Meanwhile there are other ways to increase coverage dramatically and with less work and for more broad situations.

codefromthecrypt:
This also removes a third party dep. Its only use was cross-compilation in macOS without CGO. To the degree that still exists, it is not relevant as we don't require cross-compilation from macOS: releases build on ubuntu.

pkg/globals/globals.go
```diff
@@ -41,23 +34,11 @@ type RunOpts struct {
 type GlobalOpts struct {
 	// RunOpts are inlined to allow tests to override parameters without changing ENV variables or flags
 	RunOpts
-	// HomeDir most importantly contains envoy binaries fetched from ManifestURL. Defaults to DefaultHomeDir
+	// HomeDir most importantly contains envoy binaries fetched from ManifestURL. Defaults to homedir.Dir/.getenvoy
```
mathetake, line 37: what is `homedir.Dir`?

mathetake (approved):

pkg/globals/globals.go
codefromthecrypt: drifted as we no longer use that lib
```suggestion
	// HomeDir most importantly contains envoy binaries fetched from ManifestURL. Defaults to $HOME/.getenvoy
```

PR #214, opened 2021-05-12
Stops generating JSON structs with protobuf

### LINKED ISSUE #204: Stop using protobuf to parse json
Go has a good json library, yet we are using protobuf generated code in order to parse a relatively simple json structure represented by manifest.proto. Rather than using proto to describe and runtime parse json, we can at least consider using normal json libs as they are far smaller and less conflicty. Moreover, we wouldn't need any of the "api" tooling at that point. If we want to provide a proto, we can have an external "interop" test that ensures the shape is the same and doesn't cost the maintenance proto libs cause. It would be even better to not use proto at all in order to specify json as there are other tools such as swagger or json schema for this.

codefromthecrypt:
We currently define our JSON format via a protobuf file. This changes to
stop using codegen for JSON marshaling. Instead, this uses normal structs
and go's standard library. In other words, we no longer route JSON through
protobuf libraries.

The overall rationale is less complexity, bulk and maintenance. It is easier to
manage fields manually vs the automation to match proto (never used) to its
 json representation.

Real outcomes are 14% less go code and coincidentally 14% smaller binaries.
The reduction is coincidental because the binary size is mostly due to less
bundling of unneeded dependencies (protobuf libraries).

Before:
```bash
$ cloc . |grep Go
Go                              55            642            995           3556
$ du -sk build/bin/*/*/getenvoy
12116	build/bin/darwin/amd64/getenvoy
11284	build/bin/linux/amd64/getenvoy
```

After:
```bash
$ cloc . |grep Go
Go                              54            560            953           3049
$ du -sk build/bin/*/*/getenvoy
10328	build/bin/darwin/amd64/getenvoy
9696	build/bin/linux/amd64/getenvoy
```

Moreover, the make process is significantly simpler and requires less tool
downloading.

The only side-effect is double-maintenance, I suggest we delete the
manifest.proto. Until then, when adding fields, we need to add both to the
manifest.proto and manifest.go

Fixes #204

NOTES:

I considered finding a proto -> json compiler, but the tools add even more heft.

Ex https://github.com/mitchellh/protoc-gen-go-json still requires runtime proto deps.  In other words it adds more files and steps than before:

Ex https://github.com/chrusty/protoc-gen-jsonschema, then generate structs with https://github.com/atombender/go-jsonschema. This is even more complicated and go-jsonschema doesn't work out of the box with schema generated by protoc-gen-jsonschema, even if it works with normal schema files.

The final resolution is that the cure is worse than the disease. The cost of using tools is higher than looking at files on occasion. If we later need schema based, we should define it in json schema instead of starting with proto. Meanwhile it is over-engineering and worse to force everything through the proto ecosystem.


pkg/manifest/manifest.go
```diff
@@ -0,0 +1,46 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package manifest
+
+// Manifest is derived from /manifest.proto
+type Manifest struct {
```
codefromthecrypt, line 18: this file now canonically defines the json form

pkg/manifest/print_test.go
```diff
@@ -36,48 +34,45 @@ standard-fips1402:1.10.0/linux-glibc     1.10.0
...
 }
 
-var goodManifest = &api.Manifest{
+var goodManifest = &Manifest{
 	ManifestVersion: "v0.1.0",
-	Flavors: map[string]*api.Flavor{
+	Flavors: map[string]*Flavor{
 		"standard": {
-			Name:          "standard",
-			FilterProfile: "standard",
```
codefromthecrypt, line 44: we never used FilterProfile in code, so removed it.

codefromthecrypt:
refactored the description to be very clear about the improvements here

mathetake (approved):
Make sense to me. Thanks.


@musaprg PTAL.

musaprg (approved):
Looks good to me :+1:

I'm wondering if there is any other way to define the manifest schema (such as JSON schema), but anyway +1 to this change.

codefromthecrypt:
@musaprg thanks for the review

> I'm wondering if there is any other way to define the manifest schema (such as JSON schema), but anyway +1 to this change.

yes we can use json-schema or use openapi/swagger. both can work and have library support in go. If interested we can convert the proto to one of these as a follow-up PR!

PR #216, opened 2021-05-13
Drops compile dep on archiver/v3

codefromthecrypt:
This removes the dependency on "archiver/v3" from the binary because
* It has no support for streaming https://github.com/mholt/archiver/pull/199, so we had to buffer to disk first
* Including "archiver/v3" makes it easy to accidentally bloat the binary (we did this by 25% before)
* The result is simpler dependencies

Before:
```bash
$ du -sk build/bin/*/*/getenvoy
8020	build/bin/darwin/amd64
7528	build/bin/linux/amd64
```

```bash
$ du -sk build/bin/*/*/getenvoy
7728	build/bin/darwin/amd64
7264	build/bin/linux/amd64
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/StackExchange/wmi v0.0.0-20210224194228-fe8f1750fd46 // indirect
 	github.com/go-ole/go-ole v1.2.5 // indirect
 	github.com/mattn/go-runewidth v0.0.12 // indirect
-	github.com/mholt/archiver/v3 v3.5.0
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/schollz/progressbar/v3 v3.8.0
 	github.com/shirou/gopsutil/v3 v3.21.4
--- a/go.sum
+++ b/go.sum
@@ -19,8 +19,6 @@ github.com/StackExchange/wmi v0.0.0-20210224194228-fe8f1750fd46 h1:5sXbqlSomvdjl
 github.com/StackExchange/wmi v0.0.0-20210224194228-fe8f1750fd46/go.mod h1:3eOhrUMpNV+6aFIbp5/iudMxNCF27Vw2OZgy4xEx0Fg=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
-github.com/andybalholm/brotli v1.0.0 h1:7UCwP93aiSfvWpapti8g88vVVGp2qqtGyePsSuDafo4=
-github.com/andybalholm/brotli v1.0.0/go.mod h1:loMXtMfwqflxFJPmdbJO0a3KNoPuLBgiu3qAvBg8x/Y=
 github.com/armon/circbuf v0.0.0-20150827004946-bbbad097214e/go.mod h1:3U/XgcO3hCbHZ8TKRvWD2dDTCfh9M9ya+I9JpbB7O8o=
 github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da/go.mod h1:Q73ZrmVTwzkszR9V5SSuryQ31EELlFMUz1kKyl939pY=
 github.com/armon/go-radix v0.0.0-20180808171621-7fddfc383310/go.mod h1:ufUuZ+zHj4x4TnLV4JWEpy2hxWSpsRywHrMgIH9cCH8=
@@ -42,9 +40,6 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/dgrijalva/jwt-go v3.2.0+incompatible/go.mod h1:E3ru+11k8xSBh+hMPgOLZmtrrCbhqsmaPHjLKYnJCaQ=
 github.com/dgryski/go-sip13 v0.0.0-20181026042036-e10d5fee7954/go.mod h1:vAd38F8PWV+bWy6jNmig1y/TA+kYO4g3RSRF0IAv0no=
-github.com/dsnet/compress v0.0.1 h1:PlZu0n3Tuv04TzpfPbrnI0HW/YwodEXDS+oPKahKF0Q=
-github.com/dsnet/compress v0.0.1/go.mod h1:Aw8dCMJ7RioblQeTqt88akK31OvO8Dhf5JflhBbQEHo=
-github.com/dsnet/golib v0.0.0-20171103203638-1ea166775780/go.mod h1:Lj+Z9rebOhdfkVLjJ8T6VcRQv3SXugXy999NBtR9aFY=
 github.com/fatih/color v1.7.0/go.mod h1:Zm6kSWBoL9eyXnKyktHP6abPY2pDugNf5KwzbycvMj4=
 github.com/fsnotify/fsnotify v1.4.7/go.mod h1:jwhsz4b93w/PPRr/qN1Yymfu8t87LnFCMoQvtojpjFo=
 github.com/ghodss/yaml v1.0.0/go.mod h1:4dBDuWmgqj2HViK6kFavaiC9ZROes6MMH2rRYeMEF04=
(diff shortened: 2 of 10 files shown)


pkg/internal/targz.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2019 Tetrate
```
codefromthecrypt, line 1:
```suggestion
// Copyright 2021 Tetrate
```

pkg/internal/targz.go
```diff
@@ -0,0 +1,75 @@
...
+	"archive/tar"
+	"compress/gzip"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// TarGZ makes archives the source as the dstination filename. This avoids a large (~3MB) dependency on archiver/v3.
+func TarGZ(dst, src string) error { // dst, src order like io.Copy
```
codefromthecrypt, line 27: note I'm using implicit test coverage for now, and when I looked at it, the places not touched will be hard, if possible to reproduce (ex tar err when the input is a unix pipe). However, I'm fine making unit tests for this anyway if we want to even if it doesn't improve statement coverage.

pkg/test/manifest/server.go
```diff
@@ -26,19 +26,19 @@ import (
...
 
 const archiveFormat = ".tar.gz"
+const buildsPath = "/builds/"
 
 // RequireManifestTestServer serves "/manifest.json", which contains download links a compressed archive of artifactDir
 func RequireManifestTestServer(t *testing.T, m *api.Manifest) *httptest.Server {
 	s := &server{t: t}
 	h := httptest.NewServer(s)
-	s.manifest = rewriteDownloadLocations(t, h.URL, m)
+	s.rewriteDownloadLocations(h.URL, *m) //nolint
```
codefromthecrypt, line 41: this lint error will go away when we merge #214

test/e2e/getenvoy_run_test.go
```diff
@@ -23,10 +23,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/mholt/archiver/v3"
```
codefromthecrypt, line 26: some time in the future, I want to make a separate go.mod file for the e2e tests, so that this dep doesn't clutter main

codefromthecrypt:
ok I retouched this with an updated description

codefromthecrypt:
I need to verify this manually prior to merge because GH actions doesn't actually download anything due to caching... `1.17.1/darwin is already downloaded`

codefromthecrypt:
We have used xz compression except for the first release (which was gz), so I had to add a dependency for that.

pkg/binary/envoy/fetch_test.go
```diff
@@ -17,26 +17,134 @@ package envoy
...
+		io.Copy(w, f)
+		w.WriteHeader(200)
+	}
+	t.Run("tar missing bin/envoy", func(t *testing.T) {
+		e := fetchEnvoy(envoyPath, server.URL, internal.GzipReader)
+		require.EqualError(t, e, fmt.Sprintf(`bin/envoy not found in %s`, server.URL))
+	})
+}
+
+func TestFetchEnvoy(t *testing.T) {
```
codefromthecrypt, line 93: ended up having to add more tests as only now I realized we are downloading both xz and gz archives

codefromthecrypt:
ok all good manually and also in CI

codefromthecrypt:
Added a commit to split e2e from the top-level go.mod as it now makes a difference

test/e2e/go.mod
```diff
@@ -0,0 +1,12 @@
+module github.com/tetratelabs/getenvoy/e2e
```
codefromthecrypt, line 1: this is the new go.mod for e2e tests. As you can see archiver/v3 is here and we can also add anything we want/need without fear of accidentally bloating the binary. cc @musaprg @mathetake

test/e2e/main_test.go
```diff
@@ -12,39 +12,42 @@
...
+	path, err := readGetEnvoyPath()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, `failed to start e2e tests due to an invalid "getenvoy" binary: %v`, err)
 		os.Exit(1)
 	}
-	e2e.GetEnvoyPath = path
+	getEnvoyPath = path
+
+	// Now, lookup the version we are to use
+	referenceTxt := "../../pkg/reference.txt"
```
codefromthecrypt, line 39: This is reading the file instead of using module embedding intentionally as it is cleaner than mixing the top-level module into this.

test/e2e/getenvoy_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package e2e_test
+package e2e
```
codefromthecrypt, line 15: put in same package as utilities to avoid exporting everything

go.mod
```diff
@@ -5,20 +5,15 @@ go 1.16
 
 require (
 	github.com/StackExchange/wmi v0.0.0-20210224194228-fe8f1750fd46 // indirect
-	github.com/andybalholm/brotli v1.0.2 // indirect
 	github.com/go-ole/go-ole v1.2.5 // indirect
-	github.com/klauspost/compress v1.12.2 // indirect
-	github.com/klauspost/pgzip v1.2.5 // indirect
 	github.com/mattn/go-runewidth v0.0.12 // indirect
-	github.com/mholt/archiver/v3 v3.5.0
```
codefromthecrypt, line 13: now that e2e is split, there's no reference here anymore nor any indirect stuff in go.sum

codefromthecrypt:
I rewrote the code and now it increases coverage albeit slightly (85.8% -> 85.9%)

codefromthecrypt:
it doesn't seem like our implementation of progress bar likes to be tapped into a read stream (rather the write side). I'll look into it prior to merge.

codefromthecrypt:
will pull out the general log and test fixture refactoring into a separate PR as it is unrelated to the archive dep, though needed for me to figure out what was going on.

codefromthecrypt:
will rebase after merging https://github.com/tetratelabs/getenvoy/pull/220

pkg/internal/tar.go
```diff
@@ -47,47 +48,94 @@ func NewDecompressor(src string, r io.Reader) (io.Reader, error) {
...
-	// No support for streaming https://github.com/mholt/archiver/pull/199
-	archive := filepath.Join(src, "archive.tar")
-	err := archiver.DefaultTar.Archive([]string{filepath.Join(src, root)}, archive)
-	if err != nil {
-		return err
-	}
-	defer os.Remove(archive) //nolint
+	basePath := filepath.Dir(root)
+	// Recurse through the path including all files and directories
+	return fs.WalkDir(src, root, func(path string, d os.DirEntry, err error) error {
```
codefromthecrypt, line 104: trivia: this is a go 1.16 form

codefromthecrypt:
oddly coverage is exactly the same 85.6%

mathetake (approved):

codefromthecrypt:
#220 removed a lot of the bloat, so I updated the description to be more precise with what's going on. The last commit includes cleanups requested before and some new ones as I noticed progress bar behaves differently depending on the HTTP response (ex if content length is available or not)

codefromthecrypt:
aggregate statement coverage is exactly the same before and after this: 86.9%

go.sum
```diff
@@ -19,8 +19,6 @@ github.com/StackExchange/wmi v0.0.0-20210224194228-fe8f1750fd46 h1:5sXbqlSomvdjl
 github.com/StackExchange/wmi v0.0.0-20210224194228-fe8f1750fd46/go.mod h1:3eOhrUMpNV+6aFIbp5/iudMxNCF27Vw2OZgy4xEx0Fg=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
-github.com/andybalholm/brotli v1.0.0 h1:7UCwP93aiSfvWpapti8g88vVVGp2qqtGyePsSuDafo4=
```
codefromthecrypt, line 22: reduction in lines here simplify our understanding of dependencies that can be used

internal/tar/tar.go
```diff
@@ -12,82 +12,153 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package tar avoids a large (~3MB) dependency on archiver/v3. These are special-cased to the needs of getenvoy.
```
codefromthecrypt, line 15:
```suggestion
// Package tar avoids third-party dependencies (ex archiver/v3) and are special-cased to the needs of getenvoy.
```

internal/tar/tar.go
```diff
@@ -12,82 +12,153 @@
...
-// Untar unarchives, stripping the base directory inside the "src" archive. Ex on "/foo/bar", "dst" will have "bar/**"
-func Untar(dst string, src io.Reader) error { // dst, src order like io.Copy
-	// No support for streaming https://github.com/mholt/archiver/pull/199
-	tmpDir, err := ioutil.TempDir("", "getenvoy-")
-	if err != nil {
-		return err
+// newDecompressor returns an "xz" or "gzip" decompression function based on bytes in the stream.
+func newDecompressor(r io.Reader) (io.ReadCloser, error) {
+	br := bufio.NewReader(r)
+	h, err := br.Peek(xz.HeaderLen)
```
codefromthecrypt, line 83: @mathetake doing this now and backfilled some tests around corner cases

internal/tar/tar_test.go
```diff
@@ -12,9 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package tar_test
+package tar
```
codefromthecrypt, line 15: to test un-exported functions

mholt:
FYI archiver v4 (coming soon from https://github.com/mholt/archiver/pull/302) has an all-new stream oriented API. :+1:

PR #217, opened 2021-05-14
Backfills tests in CLI

codefromthecrypt:
This cleans up some sentences and backfills tests to raise test coverage a bit (currently 85.8%)


pkg/cmd/app_test.go
```diff
@@ -37,19 +36,9 @@ func TestGetEnvoyValidateArgs(t *testing.T) {
 		args        []string
 		expectedErr string
 	}{
-		{
-			name:        "--home-dir empty",
```
codefromthecrypt, line 41: removed empty tests because it is an odd use case, and it is hard to differentiate empty from never set in practice

pkg/cmd/doc_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+func TestGetEnvoyDoc(t *testing.T) {
+	tempDir, deleteTempDir := morerequire.RequireNewTempDir(t)
+	defer deleteTempDir()
+
+	c, _, _ := newApp(&globals.GlobalOpts{})
+	c.SetArgs([]string{"doc", "-o", tempDir, "-l", "/reference/"})
+	require.NoError(t, c.Execute())
+
+	for _, file := range []string{"getenvoy.md", "getenvoy_list.md", "getenvoy_fetch.md", "getenvoy_run.md"} {
```
codefromthecrypt, line 38: these tests are here mainly to know what we were writing in the first place

codefromthecrypt:
@musaprg FYI as this is more housekeeping in nature, I'm not bothering you about approval, but if you want to take a look or find something afterwards we should change, let me know.

cmd/getenvoy/main.go
```diff
@@ -15,14 +15,36 @@
...
 )
 
 func main() {
-	if cmdutil.Execute(cmdutil.NewRoot(&globals.GlobalOpts{})) != nil {
-		os.Exit(1)
+	os.Exit(run(os.Stdout, os.Stderr, os.Args))
+}
+
+// run handles all error logging and coding so that no other place needs to.
+func run(stdout, stderr io.Writer, args []string) int {
```
codefromthecrypt, line 31: I moved the error printing logic here so that we don't need to test stderr formatting of an error on every test. This was distracting, spread logic in different places and generally made it harder to get good test coverage.

pkg/cmd/app_test.go
```diff
@@ -205,11 +196,11 @@ func requireSetenv(t *testing.T, key, value string) func() {
 	}
 }
 
-// newRootCommand initializes a command with buffers for stdout and stderr.
-func newRootCommand(o *globals.GlobalOpts) (c *cobra.Command, stdout, stderr *bytes.Buffer) {
+// newApp initializes a command with buffers for stdout and stderr.
+func newApp(o *globals.GlobalOpts) (c *cobra.Command, stdout, stderr *bytes.Buffer) {
```
codefromthecrypt, line 200: chose name app as the root command represents the application. also, in urlfave it is called app

pkg/cmd/errors.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2019 Tetrate
```
codefromthecrypt, line 1:
```suggestion
// Copyright 2021 Tetrate
```

Issue #115, opened 2021-03-03
Introduce CLI completion to getenvoy


musaprg:
Currently, we have to exec `getenvoy fetch` to know which version is available before executing `getenvoy run`. It is a small but actual pain for me. It would be nice to introduce CLI completion (bash-like or zsh-like) to `getenvoy`, especially `getenvoy run`. For instance, getenvoy can complete available "flavors" when typed "tab" after `getenvoy run`.

GetEnvoy uses https://github.com/spf13/cobra, which has a way to generate CLI completion for some popular shells (bash, zsh, fish, PowerShell) [^1]. It could be done with it.

[^1]: https://github.com/spf13/cobra/blob/master/shell_completions.md

codefromthecrypt:
I have personally found cobra very clunky. I wonder what you think about this? https://github.com/urfave/cli/blob/master/docs/v2/manual.md

codefromthecrypt:
should be easy to get off cobra now

codefromthecrypt:
@musaprg @llinder either of you want to try this after #218? here's an example project that bundles auto-completion https://github.com/rliebz/tusk

codefromthecrypt:
I'm going to deprioritize this for now though someone else can grab it.

Note that auto-completion of built-in commands is usually easier and different than auto-completion of dynamic args (ex envoy versions). In each case, how to install is somewhat shell specific, so testing this well is a fair amount of work unless we punt and assume things just work.

PR #218, opened 2021-05-14
Changes to urfave CLI, implements version command and one-page doc

codefromthecrypt:
This changes to urfave CLI, which is easier to work with than Cobra.

Pros of both:
* long source history, large amount of stars, reasonably frequent releases
* support the platforms we need (macOS, Windows, linux)
* support things we need in the future like auto-completion #115 

Pros of urfave:
* significantly less indirect dependencies (283 less lines in go.sum)
* supports ENV variables without introducing another dependency
* cleaner support for go context (haven't used it yet)
* less awkward function names and behaviors (ex. `Before` instead of `PersistentPreRunE`)
* simpler help output and markdown (ex single page vs multi-page)
* slightly smaller binary (They even [measure](https://github.com/urfave/cli/blob/45952a7d1b0e956db2099493fe1a42837ff5bd0d/internal/build/build.go#L186-L202) it!)

Cons of urfave
* doesn't yet support [argument metadata](https://github.com/urfave/cli/issues/1074)
  * We only have two args, the `<reference>` and then anything passed to envoy `-- .*`, so this isn't a biggie

https://github.com/urfave/cli/blob/master/docs/v2/manual.md


go.sum
```diff
@@ -1,120 +1,25 @@
-cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
codefromthecrypt, line 1: 283 lines gone from the go.sum file

pkg/cmd/testdata/getenvoy.md
```diff
@@ -1,27 +1,51 @@
-+++
-title = "getenvoy"
-type = "reference"
-parent = "root"
-command = "getenvoy"
-+++
-## getenvoy
+% getenvoy 8
```
codefromthecrypt, line 1: need to polish the format template, but in general it is much better IMHO to do single-page markdown for such a small utility.

pkg/cmd/run.go
```diff
@@ -31,34 +31,51 @@ import (
...
-			r.Err = c.ErrOrStderr()
+
+			r.Out = c.App.Writer
+			r.Err = c.App.ErrWriter
 
 			debug.EnableAll(r)
 
-			envoyArgs := args[1:]
-			return r.Run(c.Context(), envoyArgs)
+			envoyArgs := findEnvoyArgs(args)
```
llinder, line 58: I like the clarity of this 👍

codefromthecrypt:
I will try to finish this tomorrow morning asia time

codefromthecrypt:
once this is merged, I will help simplify the getenvoy-ui usage of "getenvoy doc" to emit the now much simpler single page. This should suffice, but if we later feel we want to edit it, we can override the template like so https://github.com/google/skia-buildbot/blob/4c87209ff9378a9befc65d5f93c4040d4fdd5922/go/urfavecli/urfavecli.go#L34 eventually, we should make markdown completely driven by the UI project instead of a hidden command as mentioned in #205

pkg/cmd/testdata/getenvoy.md
mathetake: I guess this breaks hugo builds

mathetake: we need +++ block

codefromthecrypt: magic info :) can't wait until we have things in the same repo

codefromthecrypt: agree even latest hugo can't handle this. will make a change

codefromthecrypt: this isn't really hugo related as normal markdown breaks rendering. It isn't broken, per-se, just it adds a useless line. I raised https://github.com/urfave/cli/issues/1275

codefromthecrypt:
I have a WIP branch for the website that correctly renders now https://github.com/tetratelabs/getenvoy.io/compare/new-usage

Screen Shot 2021-05-21 at 9 28 09 AM

pkg/cmd/app_test.go
```diff
@@ -178,25 +178,23 @@ func requireSetenv(t *testing.T, key, value string) func() {
...
 	c, stdout, stderr := newApp(o)
-	c.AddCommand(&cobra.Command{
-		Use: "test",
-		RunE: func(cmd *cobra.Command, _ []string) error {
-			return nil
-		},
-	})
-	c.SetArgs(append(args, "test"))
-	err := c.Execute()
+	c.Commands = append(c.Commands, &cli.Command{Name: "test", Action: func(context *cli.Context) error {
```
mathetake, line 193: nit
```suggestion
	c.Commands = append(c.Commands, &cli.Command{Name: "test", Action: func(cli.Context) error {
```

mathetake (approved):

codefromthecrypt:
thanks for the review @llinder and @mathetake!

PR #222, opened 2021-05-18
Reduces code used to implement version

codefromthecrypt:


Makefile
```diff
@@ -30,7 +30,7 @@ COVERAGE_REPORT := $(COVERAGE_DIR)/coverage.html
 GOOS := $(shell go env GOOS)
 GOARCH := $(shell go env GOARCH)
 
-GO_LD_FLAGS := -ldflags="-s -w -X github.com/tetratelabs/getenvoy/pkg/version.version=$(GETENVOY_TAG)"
+GO_LD_FLAGS := -ldflags="-s -w -X github.com/tetratelabs/getenvoy/internal/version.Current=$(GETENVOY_TAG)"
```
codefromthecrypt, line 33: @llinder FYI: this can override variables at build time, but not const (var, not const)

pkg/cmd/app.go
```diff
@@ -39,7 +39,7 @@ func NewApp(o *globals.GlobalOpts) *cobra.Command {
 		DisableAutoGenTag: true, // removes autogenerate on ___ from produced docs
 		Short:             "Fetch and run Envoy",
 		Long:              `Manage Envoy lifecycle including fetching binaries and collection of process state.`,
-		Version:           version.Build.Version, // TODO: Add version support on the command
```
codefromthecrypt, line 42: old comment as -v is by default there (and how I tested this)

```bash
$ make bin GETENVOY_TAG=asdas
CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="-s -w -X github.com/tetratelabs/getenvoy/internal/version.Current=asdas" -o build/bin/linux/amd64/getenvoy ./cmd/getenvoy/main.go
CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 go build -ldflags="-s -w -X github.com/tetratelabs/getenvoy/internal/version.Current=asdas" -o build/bin/darwin/amd64/getenvoy ./cmd/getenvoy/main.go

$ build/bin/darwin/amd64/getenvoy -v
getenvoy version asdas
```

internal/version/version.go
```diff
@@ -0,0 +1,19 @@
+// Copyright 2021 Tetrate
```
mathetake, line 1: so now what's the criteria for deciding whether a package should sit in interna/ or pkg/?

codefromthecrypt: over time, everything except main.go will be internal, as we don't have a public api

codefromthecrypt:
will rebase after #224 so we don't need to update two things with the same value

.goreleaser.yaml
```diff
@@ -15,7 +15,7 @@
 project_name: getenvoy
 builds:
 - binary: getenvoy
-  ldflags: "-s -w -X github.com/tetratelabs/getenvoy/pkg/version.version={{.Version}}"
+  ldflags: "-s -w -X github.com/tetratelabs/getenvoy/internal/version.Current={{.Version}}"
```
codefromthecrypt, line 18: only one time now!

PR #224, opened 2021-05-18
Deduplicates make config and removes orphaned config

codefromthecrypt:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,13 +29,10 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
 
       - name: "Run GoReleaser"
-        uses: goreleaser/goreleaser-action@v2
-        with:
-          version: latest
-          args: release --rm-dist
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           HOMEBREW_GETENVOY_GITHUB_TOKEN: ${{ secrets.GETENVOY_CI_GITHUB_TOKEN }}
+        run: make release
 
   e2e:
     name: Run e2e tests (${{ matrix.runner.os }})
@@ -72,9 +69,9 @@ jobs:
           INPUT_VERSION: tags/${{ env.RELEASE_TAG }}
         run: |  # extract getenvoy to location used in `make e2e`. don't chmod because tar.gz should be correct.
           curl -s https://raw.githubusercontent.com/dsaltares/fetch-gh-release-asset/0.0.6/fetch_github_asset.sh | bash
-          GOOS=$(uname -s|tr '[:upper:]' '[:lower:]')
-          mkdir -p build/bin/${GOOS}/amd64
-          tar -C build/bin/${GOOS}/amd64 -xf ${INPUT_FILE} getenvoy
+          dir=dist/getenvoy_$(go env GOOS)_$(go env GOARCH)
+          mkdir -p $dir
+          tar -C $dir -xf ${INPUT_FILE} getenvoy
 
       - name: "Run e2e tests using released `getenvoy` binary"
         run: make e2e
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -72,9 +72,9 @@ jobs:
           INPUT_VERSION: tags/${{ env.RELEASE_TAG }}
         run: |  # extract getenvoy to location used in `make e2e`. don't chmod because tar.gz should be correct.
           curl -s https://raw.githubusercontent.com/dsaltares/fetch-gh-release-asset/0.0.6/fetch_github_asset.sh | bash
-          GOOS=$(uname -s|tr '[:upper:]' '[:lower:]')
-          mkdir -p build/bin/${GOOS}/amd64
-          tar -C build/bin/${GOOS}/amd64 -xf ${INPUT_FILE} getenvoy
+          dir=dist/getenvoy_$(go env GOOS)_$(shell go env GOARCH)
```
codefromthecrypt, line 75:
```suggestion
          dir=dist/getenvoy_$(go env GOOS)_$(go env GOARCH)
```

codefromthecrypt:
coverage has annoyed me enough I will fix it then rebase over..

codefromthecrypt:
will rebase after #225

codefromthecrypt:
ok I also polished the coverage setup as it can be slightly simpler

Makefile
```diff
@@ -18,89 +18,51 @@
...
 coverage:
 	@echo "--- coverage ---"
-	@rm -rf coverage
-	@mkdir coverage
-	@go test -coverprofile=coverage/coverage.txt -covermode=atomic --coverpkg $(COVERAGE_PACKAGES) $(TEST_PKG_LIST)
-	@go tool cover -html=coverage/coverage.txt -o coverage/coverage.html
-
-##@ Code quality and integrity
+	@go test -coverprofile=coverage.txt -covermode=atomic --coverpkg $(COVERAGE_PACKAGES) $(TEST_PACKAGES)
+	@go tool cover -func coverage.txt
```
codefromthecrypt, line 59: since we now use codecov.io for visualizing in html, this tool is default (which aggregates whereas go's html does not)

Makefile
```diff
@@ -18,89 +18,51 @@
...
-GO_LD_FLAGS := -ldflags="-s -w -X github.com/tetratelabs/getenvoy/pkg/version.version=$(GETENVOY_TAG)"
-
-TEST_PKG_LIST ?= $(shell go list ./... | grep -v github.com/tetratelabs/getenvoy/test/e2e)
-GO_TEST_OPTS ?=
-GO_TEST_EXTRA_OPTS ?=
-
-E2E_PKG_LIST ?= ./test/e2e
-# Run only one test at a time, in verbose mode, so that failures are easy to diagnose.
-# Note: -failfast helps as it stops at the first error. However, it is not a cacheable flag, so runs won't cache.
-E2E_OPTS ?= -parallel 1 -v -failfast
```
codefromthecrypt, line 38: in general I removed all the variables as we should focus on the makefile working by default (overrides unnecessary). Before, all the variables cluttered the file and made it easy to not see what actually was happening

.github/workflows/release.yaml
```diff
@@ -29,13 +29,10 @@ jobs:
...
 
       - name: "Run GoReleaser"
-        uses: goreleaser/goreleaser-action@v2
-        with:
-          version: latest
-          args: release --rm-dist
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           HOMEBREW_GETENVOY_GITHUB_TOKEN: ${{ secrets.GETENVOY_CI_GITHUB_TOKEN }}
+        run: make release
```
codefromthecrypt, line 35: we now use bingo, so there will be exact match between the goreleaser we linted with and what we actually release with

codefromthecrypt:
again happy for feedback post merge

PR #225, opened 2021-05-18
Simplifies coverage configuration and replaces manual analysis with codecov.io

### LINKED ISSUE #186: Coverage report isn't very useful
**Describe the bug**
We have a coverage target, but the output isn't very useful as it has no summary data. It looks like this:

Screen Shot 2021-04-20 at 9 11 56 AM

**To Reproduce**
make coverage
open build/coverage/coverage.html

**Expected behavior**
I would expect this to give some sort of summary

codefromthecrypt:
This removes so many variable no one would set, fixes coverage packages and makes sure it builds into the ./coverage directory. Coverage is now analyzed with codecov.io

Fixes #186

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -62,13 +62,8 @@ jobs:
       - name: "Run e2e tests using the `getenvoy` binary"
         run: make e2e
 
-      - name: "Generate test coverage report"
+      - name: "Generate and upload coverage report"
         if: runner.os == 'Linux'  # no need to do this per operating system
-        run: make coverage
-
-      - name: "Upload test coverage report"
-        uses: actions/upload-artifact@v2
-        if: runner.os == 'Linux'  # no need to do this per operating system
-        with:
-          name: coverage
-          path: coverage
+        run: |  # This uses the $CODECOV_TOKEN repository secret
+          make coverage
+          bash <(curl -s https://codecov.io/bash) -f coverage/coverage.txt
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -71,4 +71,4 @@ jobs:
         if: runner.os == 'Linux'  # no need to do this per operating system
         with:
           name: coverage
-          path: build/coverage
+          path: coverage
--- a/Makefile
+++ b/Makefile
@@ -18,14 +18,10 @@
 # bingo manages go binaries needed for building the project
 include .bingo/Variables.mk
 
-ENVOY = $(shell cat internal/reference/latest.txt)
 GETENVOY_TAG ?= dev
 
 BUILD_DIR ?= build
 BIN_DIR ?= $(BUILD_DIR)/bin
-COVERAGE_DIR ?= $(BUILD_DIR)/coverage
-COVERAGE_PROFILE := $(COVERAGE_DIR)/coverage.out
-COVERAGE_REPORT := $(COVERAGE_DIR)/coverage.html
 
 GOOS := $(shell go env GOOS)
 GOARCH := $(shell go env GOARCH)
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -18,14 +18,10 @@
 # bingo manages go binaries needed for building the project
 include .bingo/Variables.mk
 
-ENVOY = $(shell cat internal/reference/latest.txt)
```
codefromthecrypt, line 21: unrelated, but unused

codefromthecrypt:
note: clicking the report won't work yet because there's no base report yet. if any problem post merge I will fix it.

.github/workflows/commit.yaml
```diff
@@ -62,13 +62,8 @@ jobs:
...
-
-      - name: "Upload test coverage report"
-        uses: actions/upload-artifact@v2
-        if: runner.os == 'Linux'  # no need to do this per operating system
-        with:
-          name: coverage
-          path: build/coverage
+        run: |  # This uses the $CODECOV_TOKEN repository secret
+          make coverage
+          bash <(curl -s https://codecov.io/bash) -f coverage/coverage.txt
```
codefromthecrypt, line 69: this is the canonical way to upload, except we are specifying a separate directory instead of letting coverage fall into the root https://github.com/codecov/example-go/blob/master/.github/workflows/ci.yml#L18

README.md
```diff
@@ -1,4 +1,6 @@
 [![Build](https://github.com/tetratelabs/getenvoy/workflows/build/badge.svg)](https://github.com/tetratelabs/getenvoy)
+[![Coverage](https://codecov.io/gh/tetratelabs/getenvoy/branch/master/graph/badge.svg)](https://codecov.io/gh/tetratelabs/getenvoy)
+[![Go Report Card](https://goreportcard.com/badge/github.com/tetratelabs/getenvoy)](https://goreportcard.com/report/github.com/tetratelabs/getenvoy)
```
codefromthecrypt, line 3: added this because, well, we are A+

codefromthecrypt:
I'm going to merge this as it is easy to revert, but doubt anyone will want it to not exist. Meanwhile, it has been tedious to calculate coverage on every PR.

codefromthecrypt:
As usual a couple bumps along the way. The correct snippet for GH actions is this:

```yaml
      - name: "Generate coverage report" # only once (not per OS)
        run: make coverage

      - name: "Upload coverage report" # only on master push and only once (not per OS)
        if: github.event_name == 'push' && github.ref == 'refs/heads/master' && runner.os == 'Linux'
        env:
```

The correct Makefile entry is this, noting that the last line allows manual review inclusive of aggregate count  (even if that aggregate doesn't exactly match what codecov says, it is still better than having to log in)

```Makefile
TEST_PACKAGES ?= $(shell go list ./... | grep -v github.com/tetratelabs/getenvoy/test/e2e)
COVERAGE_PACKAGES ?= $(shell echo $(TEST_PACKAGES)| tr -s " " ",")
.PHONY: coverage
coverage:
	@echo "--- coverage ---"
	@go test -coverprofile=coverage.txt -covermode=atomic --coverpkg $(COVERAGE_PACKAGES) $(TEST_PACKAGES)
	@go tool cover -func coverage.txt
```

And this is interesting security-wise because we don't need to add repo oauth. This all works using a token issued by codecov (which you can use manually if you want).

cc @llinder @nacx hope this helps!

codefromthecrypt:
one thing I didn't mention in the description explicitly is why codecov. The problem described in #186 was surprising to me to find out, but basically go doesn't have good coverage tools, at least not that emit html. I asked a lot of people and the best I could find for aggregate coverage was in text format.

Ex. running `make coverage` now, you can see aggregate like this
```
total:									(statements)			86.8%
```

nevermind that this number doesn't match codecov, by almost 10pct 🗡️ 

Anyway, the html output was not helpful in getting a sense of overall coverage. Again, this is unlike other languages who have decent tools. In other words codecov is primarily a cheap way to get better html without eliminating the ability to run coverage locally (ex via `make coverage`). In fact, this tool is completely disconnected as you are literally uploading the output of that process.

I chose codecov.io because a lot of other go projects use it and also it is non-interference (it just analyzes the go coverage output). This means any problem you just delete the README badge and stop uploading. Very low "exit risk" I didn't bother analyzing other saas as it looked good enough, and not risky, and well used.

Here's a summary which is more about a coverage analyzer in general, though some is specific to the "clean exit" path of codecov:
 
Why not a couple months earlier?
* we would have 3-4x code and lose time adding coverage to things we would delete later
* coverage cheated by skipping packages because the tests would hang
* project ownership wasn't great and you need that to keep an eye out to avoid random test coverage missions

Why now?
* this is more helpful than me offline passing coverage info, especially as we are still digging out of tech debt
* the coverage numbers, even the smaller one reported by codecov vs offline, are not embarassing

What about bad?
* codecov is down so uploads fail. This fails master (not PR) builds, we catch the exit code in GH actions and make it warn instead
* codecov starts annoying people with stupid info, like raised by 0.1%. We monitor this and tune thresholds
* codecov becomes 90% annoyance. We delete upload and README badge
* code or maintenance grows mainly in service to codecov. We delete upload and README badge

Meanwhile, as a severe hesitant about dev SaaS, I'm still ok hopefully some of the above makes that clear!

PR #229, opened 2021-05-21
Completes transition to internal packaging

codefromthecrypt:
We formerly used historical conventions for project layout. After
summarizing earlier that the code here is only for use in CLI, we
started to move code under the internal package here
https://github.com/tetratelabs/getenvoy/pull/222#pullrequestreview-661637691

This completes the work, only leaving "e2e" and "main.go" exposed. We
have some other thinking to do about "e2e", but for now it should be ok
to leave as a separate dir.

See https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface
See https://github.com/hashicorp/terraform/pull/28723


.goreleaser.yaml
```diff
@@ -14,22 +14,22 @@
...
-  env:
-  - CGO_ENABLED=0
-  goos:
-    - linux
-    - darwin
-  goarch:
-    - amd64
+  - binary: getenvoy
+    ldflags: "-s -w -X github.com/tetratelabs/getenvoy/internal/version.Current={{.Version}}"
+    main: .
```
codefromthecrypt, line 19: cool thing is because we now re-use goreleaser in `make e2e`, we don't need to worry if this will break late during release.

internal/cmd/run.go
```diff
@@ -23,10 +23,10 @@ import (
 
 	"github.com/urfave/cli/v2"
 
+	"github.com/tetratelabs/getenvoy/internal/binary/envoy"
+	"github.com/tetratelabs/getenvoy/internal/binary/envoy/debug"
+	"github.com/tetratelabs/getenvoy/internal/globals"
 	internalreference "github.com/tetratelabs/getenvoy/internal/reference"
```
mathetake, line 29: now this sounds weird since we have all internal..
```suggestion
	"github.com/tetratelabs/getenvoy/internal/reference"
```

mathetake: also in internal/cmd/fetch.go

codefromthecrypt: agreed. I will hunt for these now

mathetake (approved):
LGTM after fix for the comment above

codefromthecrypt:
thanks again @mathetake!

PR #230, opened 2021-05-21
Consolidates where we handle versions

codefromthecrypt:


internal/version/version.go
```diff
@@ -14,6 +14,12 @@
...
 package version
 
-// Current is the version of GetEnvoy, used in help statements and HTTP requests via "User-Agent".
-// Override this via "-X github.com/tetratelabs/getenvoy/internal/version.Current=XXX"
-var Current = "dev"
+import _ "embed" // We embed the Envoy version so that we can cache it in CI
+
+// Main is the version of GetEnvoy, used in help statements and HTTP requests via "User-Agent".
+// Override this via "-X github.com/tetratelabs/getenvoy/internal/version.Main=XXX"
+var Main = "dev"
```
mathetake, line 21: I prefer `version.GetEnvoy` to `version.Main`, which seems consistent with `version.Envoy` 😄  WDYT

codefromthecrypt: works for me

mathetake (approved):
🥳

PR #232, opened 2021-05-24
Simplifies CLI experience

codefromthecrypt:
Simplifies CLI experience

This performs a series of changes similiar to others we've done in order
to polish the experience for a release. Most of these are about reducing
terminology, jargon or code not used to support them.

### `getenvoy list` -> `getenvoy versions`
To reduce the amount of vocab and keep words aligned, this replaces
`getenvoy list` with `getenvoy versions`. The output is only the version
numbers and only those valid for the platform.

### `getenvoy fetch` -> `getenvoy install`
Most tools do not use the term fetch for downloading and extracting
something into the local cache. This uses the term "install" instead,
similar to `nvm` and `brew`.

### Simpler directories
Before, we filed downloaded versions of Envoy in a complicated path
under $ENVOY_HOME. We now follow `nvm`'s example, and simply file under
`$ENVOY_HOME/versions/$version`. This has parity both with the `getenvoy
versions` command and also arg0 of `getenvoy run` and `getenvoy
install`.

`getenvoy run` now archives to `$ENVOY_HOME/runs/$timestamp.tar.gz`

### No more "build" "reference" and "flavor" jargon
Before, we allowed installing and running Envoy on an invalid platforms.
We made a complicated "reference" syntax which allows you to download
invalid platforms, and forced users to use them instead of version
numbers. This completely ends this practice and all associated jargon.

* You can no longer install or run an invalid binary for the host
* There's no more reference or flavor concept, only version
* There is no more teasing people by listing versions they can't run.
* Those who want to fetch versions they can't run can use `jq` and
  `curl`.
* Those who want to have "flavors" can override the manifest URL

### No more `getenvoy run` `--` collisions
Cobra required us to add `--` in the commandline of `getenvoy run`. This
had two problems. One, it is not intuitive and distracting. Two, it
collides with Envoy's `--` arg. As we no longer need Cobra, we can
delete this and with it the collission.

See http://tclap.sourceforge.net/manual.html

follow-up from #195 (reducing jargon and terminology)
follow-up from #203 (making flavor optional first)


internal/binary/envoy/list.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2019 Tetrate
```
codefromthecrypt, line 1:
```suggestion
// Copyright 2021 Tetrate
```

internal/cmd/run_test.go
```diff
@@ -93,14 +88,14 @@ func TestGetEnvoyRun(t *testing.T) {
...
-			c, stdout, stderr := newApp(&o.GlobalOpts)
+			// Run "getenvoy run 1.17.1 -c envoy.yaml"
+			c, stdout, stderr := newApp(o)
 			err := c.Run(test.args)
 
 			// Verify the command invoked, passing the correct default commandline
 			require.NoError(t, err)
 
-			// We expect getenvoy to print the context it will run, and Envoy to execute the same, except adding the
+			// We expect getenvoy to printVersions the context it will run, and Envoy to execute the same, except adding the
```
codefromthecrypt, line 98:
```suggestion
			// We expect getenvoy to print the context it will run, and Envoy to execute the same, except adding the
```

internal/binary/envoy/install.go
```diff
@@ -31,47 +32,64 @@ import (
...
+		fmt.Fprintln(o.Out, version, "is already downloaded") //nolint
 	default:
 		// TODO: figure out how to get a stat error that isn't file not exist so we can test this
 		return "", err
 	}
-	return verifyEnvoy(platformPath)
+	return verifyEnvoy(installPath)
+}
+
+// LocateBuild returns the downloadLocationURL of the associated envoy binary in the manifest using the input key
```
mathetake, line 69: what exactly does "input key" refer to?

internal/binary/envoy/install.go
```diff
@@ -31,47 +32,64 @@ import (
...
+		return "", errorNoVersions
+	}
+
+	// Error if the only released "flavor" or it has no version
+	if m.Flavors["standard"] == nil || len(m.Flavors["standard"].Versions) == 0 {
+		return "", errorNoVersions
+	}
+
+	// Error if the version doesn't exist or has no builds for this platform
+	builds := m.Flavors["standard"].Versions[version]
```
mathetake, line 83:
```suggestion
	// Error if the only released "flavor" or it has no version
	f, ok := m.Flavors["standard"]
	if !ok || len(f.Versions) == 0 {
		return "", errorNoVersions
	}

	// Error if the version doesn't exist or has no builds for this platform
	builds := f.Versions[version]
```

internal/binary/envoy/debug/node.go
```diff
@@ -187,7 +187,7 @@ func printProcessTable(out io.Writer, parsed []*proc) error {
 func (n *nodeCollection) networkInterfaces() error {
 	result, err := net.Interfaces()
 	if err != nil {
-		return fmt.Errorf("unable to fetch network Interfaces: %w", err)
+		return fmt.Errorf("unable to read network Interfaces: %w", err)
```
mathetake, line 190: I think `list` would be more precise in this context: https://github.com/golang/go/blob/master/src/net/interface.go#L99

mathetake: `read` makes me feel like I'm reading frames from NIC literally, nit 😄

internal/binary/envoy/debug/node.go
```diff
@@ -221,7 +221,7 @@ var typeMap = map[uint32]string{
 func (n *nodeCollection) activeConnections() error {
 	cs, err := net.Connections("all")
 	if err != nil {
-		return fmt.Errorf("unable to fetch network interfaces: %w", err)
+		return fmt.Errorf("unable to read network interfaces: %w", err)
```
mathetake, line 224: same here

internal/binary/envoy/list.go
```diff
@@ -0,0 +1,52 @@
...
+	"sort"
+
+	"github.com/tetratelabs/getenvoy/internal/manifest"
+)
+
+// PrintVersions retrieves the manifest from the passed location and writes it to the passed writer
+func PrintVersions(m *manifest.Manifest, goos string, w io.Writer) {
+	p := manifest.BuildPlatform(goos)
+	// print nothing if the only released "flavor" or the platform doesn't exist
+	if m.Flavors["standard"] == nil || p == "" {
```
mathetake: Let's use `f, ok := m.Flavors["standard"]` here and the below for indexing with "standard".

internal/binary/envoy/list_test.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/tetratelabs/getenvoy/internal/manifest"
+)
+
+func TestPrintVersions(t *testing.T) {
```
mathetake: I don't see there's a case for checking the sort is working correctly? (the order of map entries is not deterministic).

internal/errors/errors.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package cmd
+package errors
```
mathetake, line 15: `NewValidationError` makes feel wtf does "validation" mean, and the reality is that it is basically just aliasing `fmt.Errorf`. Do we really need this package now?

internal/binary/envoy/install_test.go
```diff
@@ -90,16 +85,47 @@ func TestUntarEnvoy(t *testing.T) {
...
 
 	o.ManifestURL += "/mannyfest.json"
-	_, e := FetchIfNeeded(&o.GlobalOpts, o.reference)
+
+	_, e := InstallIfNeeded(&o.GlobalOpts, runtime.GOOS, version.Envoy)
 	require.EqualError(t, e, "received 404 status code from "+o.ManifestURL)
 	require.Empty(t, o.Out.(*bytes.Buffer))
 }
 
+func TestInstallIfNeeded_Validates(t *testing.T) {
```
mathetake, line 99: this test took me a few minutes figure out what these cases actually test.. the difficult part is that the manifest setup here in `setupTest` is not defined in this _test.go but declared in the nested `RequireManifestTestServer`.. Can we improve here? e.g. having `RequireManifestTestServer(t *testing.T, m *manifest.Manifest)` signature for `RequireManifestTestServer`?

mathetake (commented):
First shot 🔫 ! Mostly just nits in tests. Nice work.. will do the second round next morning.

internal/errors/errors.go
codefromthecrypt: validate is pre-flight, basically something that happens before download. I would like to represent it because we do actually use this type. It is used to print a usage hint if we know it is argument related.. any ideas?

internal/binary/envoy/install_test.go
codefromthecrypt: I really hate spending even more time on manifest server as it will be deleted in the next PR. can we stop maybe?

codefromthecrypt:
Thanks for the review @mathetake I will go through your feedback and anything not voided by the next PR I'll throw in, and comment why if not! https://github.com/tetratelabs/getenvoy/pull/234

internal/binary/envoy/install.go
codefromthecrypt: reworded though next PR deletes this

internal/binary/envoy/debug/node.go
codefromthecrypt: def better

internal/binary/envoy/list.go
codefromthecrypt: this code will die, but addressed this since you took time to suggest it!

internal/binary/envoy/list_test.go
codefromthecrypt: the test has this, so it should break on non-deterministic, esp since the data is added in reverse

```go
		{
			name:     "darwin",
			goos:     "darwin",
			manifest: goodManifest,
			expected: `1.18.3
1.14.6
`,
```

internal/binary/envoy/install_test.go
codefromthecrypt: maybe this wasn't in the same file before, but setupTest is in the same file now for sure.. I renamed it setupInstallTest to make it more clear and did a pass to ensure nothing else is importing it.

codefromthecrypt:
ok PTAL when you are free again. thanks @mathetake

internal/errors/errors.go
codefromthecrypt: ps this might help https://github.com/tetratelabs/getenvoy/blob/master/main.go#L48

in general we still have way too many packages, ex "cmd" vs "binary" etc and maybe this one stands out because I used it before. Once the main entanglement caused by the object model is done, I'd be happy to reduce this and more.

Ex. you'll notice the "cmd" thing makes the hierarchy more bizarre also
```
internal
internal/globals
internal/cmd
internal/transport
internal/test
internal/test/morerequire
internal/test/manifest
internal/manifest
internal/version
internal/binary
internal/binary/envoytest
internal/binary/envoy
internal/binary/envoy/debug
internal/errors
internal/tar
```

codefromthecrypt: https://github.com/tetratelabs/getenvoy/issues/235 but we can still discuss "errors" here. Hopefully the additional doc I put on that type makes sense. I don't know a better word for validate at the moment, so if you can suggest one, I'm all ears also!

codefromthecrypt: I went with up-front validation for now. There's some more refactoring we can do later wrt where validation occurs, but this should get the distraction out of this PR!

mathetake (approved):
@codefromthecrypt Thanks! LGTM

codefromthecrypt:
thanks for all the help!

PR #234, opened 2021-05-24
Implements new schema

### LINKED ISSUE #231: Dramatically simplify build list
Currently, the build list is called "manifest.proto" and it is represented only as json. It uses a map of map approach to try to keep keys unique, but that leads to very awkward code that has literally had TODOs on it like so:

```go
	for _, flavor := range deterministicFlavors(manifest.Flavors) {
		for _, version := range deterministicVersions(flavor.Versions) {
			for _, build := range deterministicBuilds(version.Builds) {
				ref := Reference{flavor.Name, version.Name, platformFromEnum(build.Platform)}
				fmt.Fprintf(w, "%s\t%s\n", ref.String(), version.Name)
			}
		}
	}
```

Over time, I think it has become clear this is a really difficult way to get a list of builds. Also, we can simplify things by optimizing for go and stop trying to normalize for other things (manifest.json is only used here).

Here's a WIP json schema which I believe will generate far easier to deal with data and help us also deal with arch in the future. We should host that schema here, host the data file here also, removing the two step process currently used to validate the build list.

Naming the actual file"builds.json" or "envoy_builds.json" will also help make it possible to know the point of the file.
```json
{
  "$schema": "https://json-schema.org/draft-06/schema#",
  "type": "array",
  "title": "Envoy builds",
  "description": "List of Envoy builds available to download",
  "items": {
    "$ref": "#/definitions/EnvoyBuild"
  },
  "examples": [
    [
      {
        "version": "1.14.6",
        "os": "darwin",
        "arch": "amd64",
        "url": "https://dl.getenvoy.io/public/raw/files/getenvoy-envoy-1.14.6.p0.gca0d611-1p70.g486d4e2-darwin-release-x86_64.tar.xz"
      },
      {
        "version": "1.18.3",
        "os": "linux",
        "arch": "amd64",
        "url": "https://dl.getenvoy.io/public/raw/files/getenvoy-envoy-1.18.3.p0.g98c1c9e-1p77.gb76c773-linux-glibc-release-x86_64.tar.xz"
      }
    ]
  ],
  "definitions": {
    "EnvoyBuild": {
      "type": "object",
      "properties": {
        "version": {
          "type": "string",
          "title": "Version",
          "description": "Version of Envoy. Ex. 1.18.4"
        },
        "os": {
          "type": "string",
          "enum": [
            "darwin",
            "linux",
            "windows"
          ],
          "title": "OS",
          "description": "Supported operating systems (subset of GOOS)"
        },
        "arch": {
          "type": "string",
          "enum": [
            "amd64",
            "arm64"
          ],
          "title": "Arch",
          "description": "Supported architectures (subset of GOARCH)"
        },
        "url": {
          "type": "string",
          "format": "uri",
          "title": "Download URL",
          "description": "Location of the tar.gz or tar.xz that bundles Envoy. Minimum contents are $version/bin/envoy"
        }
      },
      "required": [
        "arch",
        "os",
        "url",
        "version"
      ],
      "title": "Envoy build",
      "description": "List of Envoy builds available to download"
    }
  }
}
```

codefromthecrypt:
This adds a new, simpler json schema for envoy versions.

Ex.
```json
    {
      "latestVersion": "1.18.3",
      "versions": {
        "1.14.6": {
          "releaseDate": "2021-04-16",
          "tarballs": {
            "darwin/amd64": "https://getenvoy.io/versions/1.14.6/envoy-1.14.6-darwin-x86_64.tar.gz",
            "linux/amd64": "https://getenvoy.io/versions/1.14.6/envoy-1.14.6-linux-x86_64.tar.gz"
          }
        },
        "1.18.3": {
          "releaseDate": "2021-05-11",
          "tarballs": {
            "darwin/amd64": "https://getenvoy.io/versions/1.18.3/envoy-1.18.3-darwin-x86_64.tar.gz",
            "linux/amd64": "https://getenvoy.io/versions/1.18.3/envoy-1.18.3-linux-x86_64.tar.gz",
          }
        }
      }
    }
```

* validate with json-schema compliant tools (such as most IDEs)
* supports arm64 (not built here or published yet as we have no Envoy binary to test with)
* supports non-standard version lists via $ENVOY_VERSIONS_URL
* simplifies main code by avoiding trees of pointers we had before in `*manifest.Manifest`
* guarantees "latest version" is coherent.
* untangles package hierarchy a bit

*NOTE: after check-in, we need to mount site/envoy_versions.json on https://www.getenvoy.io/* and fix the default URL

Fixes #231


internal/cmd/run.go
```diff
@@ -66,10 +65,10 @@ $ getenvoy run %s --config-path ./bootstrap.yaml`, latestversion.Envoy),
 // initializeRunOpts allows us to default values when not overridden for tests.
 // The version parameter correlates with the globals.GlobalOpts EnvoyPath which is installed if needed.
 // Notably, this creates and sets a globals.GlobalOpts WorkingDirectory for Envoy, and any files that precede it.
-func initializeRunOpts(o *globals.GlobalOpts, goos, version string) error {
+func initializeRunOpts(o *globals.GlobalOpts, platform, version string) error {
 	runOpts := &o.RunOpts
 	if o.EnvoyPath == "" { // not overridden for tests
-		envoyPath, err := envoy.InstallIfNeeded(o, goos, version)
+		envoyPath, err := envoy2.InstallIfNeeded(o, platform, version)
```
codefromthecrypt, line 71: there's a lot of envoy2 fuzz. I'll fix that in next commit after feedback

internal/cmd/testdata/getenvoy_help.txt
```diff
@@ -13,6 +13,7 @@ COMMANDS:
    install   Download and install a <version> of Envoy
 
 GLOBAL OPTIONS:
-   --home-dir value  GetEnvoy home directory (location of downloaded artifacts, caches, etc) (default: ${HOME}/.getenvoy) [$GETENVOY_HOME]
-   --help, -h        show help (default: false)
-   --version, -v     print the version (default: false)
+   --home-dir value            GetEnvoy home directory (location of downloaded versions and run archives) (default: ${HOME}/.getenvoy) [$GETENVOY_HOME]
+   --envoy-versions-url value  URL of Envoy versions JSON (default: https://raw.githubusercontent.com/tetratelabs/getenvoy/new-json/site/envoy_versions.json) [$ENVOY_VERSIONS_URL]
```
codefromthecrypt, line 17: this default will be immediately followed up on the next PR because we host the site in a separate repo getenvoy.io

internal/version/envoy.txt
```diff
@@ -1 +1 @@
-1.17.1
\ No newline at end of file
+1.18.3
```
codefromthecrypt, line 1: the tests for site/** caught this btw

site/envoy_versions_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/tetratelabs/getenvoy/internal/version"
+)
+
+const envoyVersionsPath = "./envoy_versions.json"
+
+// TestEnvoyVersionsJson ensures tarball URLs in the Envoy versions JSON appear correct.
+// These are not fetched to avoid causing excess load.
+func TestEnvoyVersionsJson(t *testing.T) {
```
codefromthecrypt, line 33: similar to the latest manifest checker, which will become useless except to support old versions of getenvoy as long as we choose to do that.

internal/version/envoy.txt
mathetake: do we need now? why not just read the latest version from envoy_versions.json? (I guess you plan to do in a separate PR).

internal/envoy/install.go
```diff
@@ -42,14 +41,14 @@ func InstallIfNeeded(o *globals.GlobalOpts, p, v string) (string, error) {
...
-		m, e := manifest.GetManifest(o.ManifestURL)
+		ev, e := GetEnvoyVersions(o.EnvoyVersionsURL)
 		if e != nil {
 			return "", e
 		}
 
-		tarballURL, e := tarballURL(m, p, v)
-		if e != nil {
-			return "", e
+		tarballURL := ev.Versions[v].Tarballs[p]
```
mathetake, line 49: nit this would panic when `version, ok := ev.Versions[v]; !ok`.

site/envoy_versions_test.go
```diff
@@ -0,0 +1,67 @@
...
+	require.NotEmptyf(t, evs.LatestVersion, "latest version isn't in %s", envoyVersionsPath)
+	require.Containsf(t, evs.Versions, evs.LatestVersion, "latest version isn't in the version list of %s", envoyVersionsPath)
+	require.Equalf(t, evs.LatestVersion, version.Envoy, "version.Envoy doesn't match latest version in %s", envoyVersionsPath)
+
+	for _, tc := range tests {
+		name := fmt.Sprintf("%s-%s", tc.version, tc.platform)
+		tarballURL := tc.tarballURL
+		t.Run(name, func(t *testing.T) {
+			require.Regexpf(t, "https://.*.tar.(gz|xz)", tarballURL, "expected an https tar.gz or xz %s", tarballURL)
+			res, err := http.Head(tarballURL)
```
mathetake, line 59: I think you missed `require.NoError(t, err)` here

site/envoy_versions_test.go
```diff
@@ -0,0 +1,67 @@
...
+	require.NoErrorf(t, err, "error parsing json from %s", envoyVersionsPath)
+
+	type testCase struct{ version, platform, tarballURL string }
+
+	var tests []testCase
+	for v, ev := range evs.Versions {
+		for p, tb := range ev.Tarballs {
+			tests = append(tests, testCase{v, p, tb})
+		}
+	}
```
mathetake, line 48: would like to put these lines after three requires below.

internal/envoy/versions.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	return result, nil
+}
+
+// PrintVersions retrieves the Envoy versions from the passed location and writes it to the passed writer
+func PrintVersions(vs version.EnvoyVersions, p string, w io.Writer) {
+	// Build a list of versions for this platform
+	var versions []string
+	for v, t := range vs.Versions {
+		if len(t.Tarballs[p]) > 0 {
```
mathetake, line 58: Can we validate this condition in sites tests against envoy_versions.json?

internal/test/server.go
```diff
@@ -99,8 +89,8 @@ func (s *server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func (s *server) GetManifest() []byte {
-	data, err := json.Marshal(s.manifest)
+func (s *server) GetVersions() []byte {
```
mathetake, line 92: be more consistent 😄 
```suggestion
func (s *server) GetEnvoyVersions() []byte {
```

mathetake: and no need to be exported

mathetake (commented):
finished first round and generally looks good. getting much neater and more readable 😄

internal/envoy/install.go
codefromthecrypt: ah subtle but because this is no longer a graph of pointers, it "just works" missing tests though, so added them

site/envoy_versions_test.go
codefromthecrypt: yep!

site/envoy_versions_test.go
codefromthecrypt: agreed

internal/envoy/versions.go
codefromthecrypt: ah yeah def. we'll keep this here, too, though JIC someone re-targetted the binary

internal/version/envoy.txt
codefromthecrypt: renamed and reworded docs around "last known envoy". This was indeed an issue of confusion, so thanks for thinking about it!

internal/envoy/install_test.go
```diff
@@ -162,6 +162,20 @@ func TestInstallIfNeeded(t *testing.T) {
 	require.Contains(t, out.String(), "100% |████████████████████████████████████████|")
 }
 
+func TestInstallIfNeeded_NotFound(t *testing.T) {
```
codefromthecrypt, line 165: @mathetake here are the tests that show how good life is when the graph isn't nested pointers anymore!

internal/version/version.go
```diff
@@ -20,6 +20,32 @@ import _ "embed" // We embed the Envoy version so that we can cache it in CI
 // Override this via "-X github.com/tetratelabs/getenvoy/internal/version.GetEnvoy=XXX"
 var GetEnvoy = "dev"
 
-// Envoy is the default version to download. This is embedded for re-use in build and CI scripts.
-//go:embed envoy.txt
-var Envoy string
+// LastKnownEnvoy is the last known Envoy version, used to ensure help statements aren't out-of-date.
```
codefromthecrypt, line 23: hopefully, these docs better explain the compile vs runtime concern

site/envoy_versions.json
```diff
@@ -0,0 +1,40 @@
+{
+  "latestVersion": "1.18.3",
+  "versions": {
+    "1.14.6": {
+      "releaseDate": "2021-04-16",
```
codefromthecrypt, line 5:
```suggestion
      "releaseDate": "2020-12-08",
```

internal/version/version.go
mathetake: +1 thanks!

internal/test/server.go
```diff
@@ -99,8 +89,8 @@ func (s *server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func (s *server) GetManifest() []byte {
-	data, err := json.Marshal(s.manifest)
+func (s *server) GetEnvoyVersions() []byte {
```
mathetake, line 92:
```suggestion
func (s *server) getEnvoyVersions() []byte {
```

internal/envoy/versions.go
mathetake, line 58: what if the only tarball contained here is for a different platform?

codefromthecrypt: this is the length of the string for this platform...

codefromthecrypt: I'll switch syntax..

mathetake: ah 😂

internal/envoy/install_test.go
mathetake: super!

mathetake (approved):
image

internal/envoy/versions.go
```diff
@@ -55,7 +55,7 @@ func PrintVersions(vs version.EnvoyVersions, p string, w io.Writer) {
 	// Build a list of versions for this platform
 	var versions []string
 	for v, t := range vs.Versions {
-		if len(t.Tarballs[p]) > 0 {
+		if _, ok := t.Tarballs[p]; ok {
```
mathetake, line 58: much better! 😄

codefromthecrypt:
thanks for all the help, this project is starting to polish!

PR #236, opened 2021-05-25
Updates to latest Netlify links

codefromthecrypt:
The one about CLI is a little lying. There's a chicken/egg.. we need to serve it here before we can point to it otherwise, we blackhole it for a sec.


site/getenvoy_installer.sh
```diff
@@ -0,0 +1,401 @@
+#!/bin/sh
```
codefromthecrypt, line 1: This is a copy. After updating the _redirects, I'll go back and delete the old one

mathetake (approved):

site/README.md
```diff
@@ -0,0 +1,8 @@
+# Envoy site
+
+The following files are hosted on https://getenvoy.io, specifically via [Netlify redirects](https://github.com/tetratelabs/getenvoy.io/blob/master/site/static/_redirects):
+
+Latest master merge:
+* https://getenvoy.io/cli -> [./getenvoy_installer.sh](getenvoy_installer.sh)
```
codefromthecrypt, line 6: ps we can also call this "install.sh" similar to https://github.com/nvm-sh/nvm#install--update-script

PR #237, opened 2021-05-25
Simplifies code that adds user-agent

codefromthecrypt:


internal/envoy/http.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2019 Tetrate
```
codefromthecrypt, line 1:
```suggestion
// Copyright 2021 Tetrate
```

internal/envoy/versions.go
```diff
@@ -22,17 +22,16 @@ import (
...
 // GetEnvoyVersions returns a version map from a remote URL. eg globals.DefaultEnvoyVersionsURL.
 func GetEnvoyVersions(envoyVersionsURL string) (version.EnvoyVersions, error) {
 	result := version.EnvoyVersions{}
 	// #nosec => This is by design, users can call out to wherever they like!
-	resp, err := transport.Get(envoyVersionsURL)
-	if err != nil {
-		return result, err
+	resp, e := httpGet(envoyVersionsURL)
+	if e != nil {
+		return result, e
```
mathetake, line 34: why not use `err` ? 😄

mathetake (approved):

internal/envoy/versions.go
codefromthecrypt: switched everything to this form so no more inconsistency!

codefromthecrypt:
Thanks for the nudge @mathetake

Issue #238, opened 2021-05-25
regenerate site/install.sh after release


codefromthecrypt:
We unfortunately overridded goreleaser naming conventions formerly like this:

```
archives:
  - name_template: "getenvoy_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
    replacements:
      darwin: Darwin
      linux: Linux
      amd64: x86_64
```

This is distracting and busy work especially as we add new platforms. We now release defaults (which naturally map to docker platform and our json schema).

After we release current master, we need to regenerate and push `site/install.sh` to catch the naming convention updates.
Ex. using latest commit of https://github.com/goreleaser/godownloader

```bash
./godownloader --repo=tetratelabs/getenvoy  .goreleaser.yaml  > site/install.sh 
```

codefromthecrypt:
https://github.com/tetratelabs/getenvoy/pull/240 is ready with the fix

PR #239, opened 2021-05-25
Makes the build arm64 ready

codefromthecrypt:
This prepares the goreleaser to build arm64 even if we don't have any
envoy builds to test with, yet. Note #238 needs to happen post-release
to make sure the install script looks in the right place.


mathetake (approved):

codefromthecrypt:
thanks for the look!

PR #241, opened 2021-05-25
Disables auto-changelog and uses more defaults in goreleaser

codefromthecrypt:
Changes from defaults should be commented or removed

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -27,7 +27,8 @@ builds:
     goarch:
       - amd64
       - arm64
-# No changelog as we write release notes manually until refactoring settles down.
+# Create a draft release in GitHub. We can write release notes manually.
+draft: true
 brews:
   - name: getenvoy
     tap:
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -39,7 +39,7 @@ brews:
       token: "{{ .Env.HOMEBREW_GETENVOY_GITHUB_TOKEN }}"
     folder: Formula
     homepage: https://github.com/tetratelabs/getenvoy
-    description: Download and run Envoy
+    description: Install and run Envoy
     license: Apache-2.0
     install: |  # TODO: add shell completion post-install
       bin.install "getenvoy"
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
I noticed our release pipeline is backwards. We release before e2e testing the binary we made. We should do "goreleaser build" then "make e2e" then "goreleaser release" I think

mathetake:
> I noticed our release pipeline is backwards. We release before e2e testing the binary we made. We should do "goreleaser build" then "make e2e" then "goreleaser release" I think

I think this is intentional in order to test the "uploaded" binary works correctly vs testing the locally built binary..

mathetake:
testing the locally built binary in the release workflow is basically a duplication of the e2e in the main branch's workflow..

.goreleaser.yaml
```diff
@@ -25,26 +27,21 @@ builds:
...
+    tap:
       owner: tetratelabs
       name: homebrew-getenvoy
       # Use token other than the default one (which is limited in scope to
       # https://github.com/tetratelabs/getenvoy only)
       token: "{{ .Env.HOMEBREW_GETENVOY_GITHUB_TOKEN }}"
     folder: Formula
     homepage: https://github.com/tetratelabs/getenvoy
-    description: Distribute, deploy and operate a fleet of Envoys
+    description: Download and run Envoy
```
mathetake, line 42: nit this is not in line with the "install" command 😄 
```suggestion
    description: Install and run Envoy
```

codefromthecrypt:
I'll save the e2e on release refactoring for a different PR, but glad the conversation started. We should be testing the binary we built "before" releasing it ideally. ex "goreleaser build" only after e2e "goreleaser release". The only test we should need to do post-release is that the tar made as an attachment to the releases page. The tar would have exactly the same bytes as we tested.

mathetake:
yeah maybe be testing shasum is enough for post release. Thanks.

mathetake (approved):

codefromthecrypt:
thanks for the help again!

PR #242, opened 2021-05-26
Updates Envoy 1.14 version and ensures e2e test local json

codefromthecrypt:
This ensures pull requests run against the local json file. Doing so
allows us to update the latest release and test it in the same PR.

At first I was thinking we should only change GH actions like this:
```
      - name: "Run e2e tests using the `getenvoy` binary"
        run: make e2e
        env:  # Override getenvoy.io URL as we need to be able to test the changes prior to deployment
          ENVOY_VERSIONS_URL: https://raw.githubusercontent.com/${{ github.repository }}/${{ steps.image_and_branch_name.outputs.LOCAL_BRANCH }}/site/envoy_versions.json
```

However, we may end up needing to use travis also, and also local
dev should be coherent if people are to figure out PR failures.


codefromthecrypt:
The key change here is you can see the updated version in the e2e output:

```
=== CONT  TestGetEnvoyVersions
1.18.3
1.17.3
1.16.4
1.15.5
1.14.7
--- PASS: TestGetEnvoyVersions (0.01s)
```

site/envoy_versions.json
```diff
@@ -1,11 +1,11 @@
 {
   "latestVersion": "1.18.3",
   "versions": {
-    "1.14.6": {
```
codefromthecrypt, line 4: I replaced this not added because we've not yet released a version of getenvoy that reads this file. When we do, we should encourage use of latest patches.

e2e/main_test.go
```diff
@@ -16,53 +16,101 @@ package e2e
...
 	}
 	getEnvoyPath = path
 
+	versionLine, _, err := getEnvoy("--version").exec()
+	if err != nil {
+		exitOnInvalidBinary(err)
+	}
+
+	if _, ok := os.LookupEnv(envoyVersionsUrlEnvKey); !ok && strings.Contains(versionLine, "SNAPSHOT") {
+		s, err := mockEnvoyVersionsServer()
```
mathetake, line 51: nit you forgot to close this mock server 😄

codefromthecrypt: linter yelled at me because we call os.Exit :)

codefromthecrypt: added a comment

site/envoy_versions.json
```diff
@@ -1,11 +1,11 @@
...
-      "releaseDate": "2020-12-08",
+    "1.14.7": {
+      "releaseDate": "2021-05-16",
```
mathetake, line 5: sounds like we are lier 😄 1.14 not released in 2021 actually

mathetake: ah - security patched. nvm 😄

mathetake: maybe better add verification of releaseDate fields in test.site?

codefromthecrypt: yeah should be April.. agreed.. maybe there's a easy way to read the tag date from https://github.com/envoyproxy/envoy/releases/tag/v1.14.7

codefromthecrypt:
I will fix verification of the release tag before merge. agreed @mathetake this is a good idea and we don't want to lie by accident

codefromthecrypt:
ok the validator is now more strict!

.golangci.yml
```diff
@@ -34,7 +34,7 @@ linters:
     - "gofmt"
     - "gocyclo"
     - "goimports"
-    - "golint"
```
codefromthecrypt, line 37: golint is dead. long live revive

site/envoy_versions_test.go
```diff
@@ -28,9 +29,24 @@ import (
 
 const envoyVersionsPath = "./envoy_versions.json"
 
+// GitHubRelease includes a subset of fields we use from https://docs.github.com/en/rest/reference/repos#releases
+type GitHubRelease struct {
```
mathetake, line 33:
```suggestion
type gitHubRelease struct {
```

mathetake (approved):

site/envoy_versions_test.go
```diff
@@ -71,3 +90,36 @@ func TestEnvoyVersionsJson(t *testing.T) {
...
+		return nil, fmt.Errorf("error unmarshalling GitHub Releases: %w", err)
+	}
+
+	m := map[string]string{}
+	for _, r := range releases { //nolint:gocritic
+		if r.Draft || r.PreRelease {
+			continue
+		}
+		// clean inputs "v1.15.4" -> "2021-05-11T19:11:09Z" into "1.15.4" -> "2021-05-11"
+		m[r.Name[1:]] = r.PublishedAt[0:10]
```
mathetake, line 122:
```suggestion
		m[strings.TrimPrefix(r.Name, "v")] = r.PublishedAt[0:10]
```

codefromthecrypt:
Thanks for the help @mathetake. As always it is better than before, and as always you should be thanked even if you always help! 🙇

Issue #106, opened 2020-10-12
`getenvoy run` to start the latest version by default


rakyll:
`getenvoy run` requires users to set a target as an argument but target names are verbose. I find myself often having to `getenvoy list` to copy/paste the target name in development time. Would it be reasonable if `getenvoy run` was running the latest version compatible to the host platform by default? If users want to use specific a version/platform, they can keep passing an argument.

codefromthecrypt:
I agree and confirm that the current version doesn't do this and it should

codefromthecrypt:
we could possibly also shortcut a minimal config like so:

```bash
$ getenvoy run standard:1.17.1 -- --config-yaml "admin: {access_log_path: '/dev/stdout', address: {socket_address: {address: '127.0.0.1', port_value: 0}}}"
```

so that doing `getenvoy run` does the same thing..


https://github.com/tetratelabs/getenvoy/pull/182/files#r615385360

codefromthecrypt:
I think basically we should fetch the manifest and rank the latest version present. The only other way is to use the "last known" build which will be out of date as soon as getenvoy cli is cut. ex currently it is "1.17.1" The latter approach is more stable, but has a risk of being out of date. In any case, we can fallback to this. any thoughts?

codefromthecrypt:
Here's an idea cc @anuraaga who helped collaborate

1. switch to a `getenvoy use` model: remove the version arg from `getenvoy run` completely. This also removes arg ambiguity
2. Default to the version in ~/.getenvoy/envoy-version. If not present persist envoy_versions.json/latestVersion and install it.
3. To support parallel use, first look at $PWD/.envoy-version

In the future, we can add `getenvoy upgrade` which could fetch the latest patch or upgrade the version entirely. There are some scenarios where blind upgrading doesn't work. Another way is to support "N.N.x" syntax in the version file, which will always check for latest patch. We don't have to do this now, but this is thinking ahead.

codefromthecrypt:
while bike shed, I'm kindof partial to .envoy_version vs .envoy-version just to match envoy_versions.json. That or switch the file convention of envoy_versions.json to envoy-versions.json

nacx:
I like the idea of `getenvoy use`, especially for development environments. This is similar to other tooling that is becoming popular like `rbenv`, `jenv`, etc. configuring it once and being able to quickly run it is great.

nacx:
Just thinking out loud here:
[...]

PR #244, opened 2021-05-27
Changes envoy versions json from snake_case to kebab-case

codefromthecrypt:
see #106


codefromthecrypt:
will follow-up immediately on getenvoy.io repo. I can't raise before here as netlify redirect checker would break

mathetake (approved):

PR #245, opened 2021-05-27
Adds platform to User-Agent

codefromthecrypt:
Adding platform will tell us how many are using architectures besides
amd64. For example, we will see attempts of usage in analytics, even if
they end up eventually failing. This allows us to plan better.


e2e/command_test.go
```diff
@@ -40,7 +38,16 @@ type cmdBuilder struct {
 
 // getEnvoy returns a new command builder.
 func getEnvoy(arg0 string) *cmdBuilder { //nolint:golint
-	return &cmdBuilder{exec.Command(getEnvoyPath, arg0)} //nolint:gosec
+
+	var args []string
+	if arg0 == "--version" {
+		args = []string{arg0}
+	} else {
+		args = []string{"--internal-user-agent", userAgent, arg0}
```
mathetake, line 46: is it intentional to have "--internal-user-agent" before arg0?

codefromthecrypt: yeah it is a global flag

codefromthecrypt: I could rename arg0 to "command" but #246 will drift this anyway :D

mathetake: I mean, if it is a global flag, it should work even put this after arg0?

mathetake (approved):

e2e/command_test.go
codefromthecrypt: nope.. urfave is neater. global flags are before the command. This is helpful also because you know nothing will ever interfere with args if a subcommand defines no flags

```
USAGE:
   getenvoy [global options] command [command options] [arguments...]
```

mathetake: neat!

codefromthecrypt:
thanks for the look and the good question @mathetake!

PR #246, opened 2021-05-27
Adds `getenvoy installed` to list installed versions

codefromthecrypt:
Before, we only had a command to know the available versions. This tells
us what's installed. Later, we can print out which is the default
version.

Rationale for "installed" is that it is intuitive alongside "install" vs other options. Also, the fact that "getenvoy" is not an env-based name makes it look nicer (IMHO)

Ex.
```bash
getenvoy versions (existing command)
getenvoy install 1.18.3 (existing command)
getenvoy installed (new command)
```

Vs. Mixing different output based on a flag like `pyenv install`
```bash
pyenv install -l
pyenv install  3.9.5
pyenv versions
```

Vs. not really describing versions are the thing 'ls'ed like `nvm`
```bash
nvm ls-remote
nvm install v16.2.0
nvm ls
```

### DIFF
--- a/e2e/getenvoy_install_test.go
+++ b/e2e/getenvoy_install_test.go
@@ -43,8 +43,7 @@ func TestGetEnvoyInstall(t *testing.T) {
 		stdout, stderr, err := getEnvoy("--home-dir", homeDir, "install", version.LastKnownEnvoy).exec()
 
 		require.NoError(t, err)
-		require.Equal(t, version.LastKnownEnvoy+` is already downloaded
-`, stdout)
+		require.Equal(t, version.LastKnownEnvoy+" is already downloaded\n", stdout)
 		require.Empty(t, stderr)
 	})
 }
--- a/e2e/getenvoy_install_test.go
+++ b/e2e/getenvoy_install_test.go
@@ -15,7 +15,9 @@
 package e2e
 
 import (
+	"fmt"
 	"path/filepath"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -46,4 +48,13 @@ func TestGetEnvoyInstall(t *testing.T) {
 		require.Equal(t, version.LastKnownEnvoy+" is already downloaded\n", stdout)
 		require.Empty(t, stderr)
 	})
+}
+
+func TestGetEnvoyInstall_UnknownVersion(t *testing.T) {
+	stdout, stderr, err := getEnvoy("install", "1.1.1").exec()
+
+	require.EqualError(t, err, "exit status 1")
+	require.Empty(t, stdout)
+	require.Equal(t, fmt.Sprintf(`error: couldn't find version "1.1.1" for platform "%s/%s"
+`, runtime.GOOS, runtime.GOARCH), stderr)
 }
--- a/e2e/getenvoy_install_test.go
+++ b/e2e/getenvoy_install_test.go
@@ -1,0 +1,50 @@
+// Copyright 2021 Tetrate
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


e2e/getenvoy_install_test.go
```diff
@@ -0,0 +1,50 @@
+// Copyright 2020 Tetrate
```
codefromthecrypt, line 1:
```suggestion
// Copyright 2021 Tetrate
```

e2e/getenvoy_install_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+		require.FileExists(t, filepath.Join(homeDir, "versions", version.LastKnownEnvoy, "bin", "envoy"))
+	})
+
+	t.Run("already installed", func(t *testing.T) {
+		stdout, stderr, err := getEnvoy("--home-dir", homeDir, "install", version.LastKnownEnvoy).exec()
+
+		require.NoError(t, err)
+		require.Equal(t, version.LastKnownEnvoy+` is already downloaded
+`, stdout)
```
mathetake, line 47: prefer this for readability 
```suggestion
		require.Equal(t, version.LastKnownEnvoy+" is already downloaded\n", stdout)
```

internal/cmd/installed_test.go
```diff
@@ -0,0 +1,92 @@
...
+func TestGetEnvoyInstalled_NothingYet(t *testing.T) {
+	o, cleanup := setupTest(t)
+	defer cleanup()
+
+	c, stdout, stderr := newApp(o)
+	o.Out = stdout
+	err := c.Run([]string{"getenvoy", "installed"})
+
+	require.NoError(t, err)
+	require.Equal(t, `No envoy versions installed, yet
```
mathetake, line 34: same here

e2e/getenvoy_install_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/tetratelabs/getenvoy/internal/test/morerequire"
+	"github.com/tetratelabs/getenvoy/internal/version"
+)
+
+// TestGetEnvoyInstall needs to always execute, so we run it in a separate home directory
+func TestGetEnvoyInstall(t *testing.T) {
+	homeDir, removeHomeDir := morerequire.RequireNewTempDir(t)
+	defer removeHomeDir()
+
```
mathetake, line 31: See no test case for the attempt to install non-existing version 😄

internal/cmd/installed.go
```diff
@@ -0,0 +1,61 @@
...
+
+			// Sort so that new versions appear first
+			sort.Slice(rows, func(i, j int) bool {
+				return rows[i] > rows[j]
+			})
+
+			fmt.Fprintln(o.Out, "VERSION") //nolint
+			for _, vr := range rows {
+				fmt.Fprintln(o.Out, vr) //nolint
+			}
```
mathetake, line 57: nit how about creating string first, and then call the print syscall once?
```suggestion
			fmt.Fprintln(o.Out, "VERSION") //nolint
			out := "VERSION\n"
			for _, vr := range rows {
				out += vr + "\n"
			}
			fmt.Fprint(o.Out, out)
```

mathetake: and better use `strings.Join(rows, "\n")`

e2e/getenvoy_install_test.go
codefromthecrypt: agreed!

internal/cmd/installed.go
codefromthecrypt: sure!

codefromthecrypt:
Thanks @mathetake much less code now, except tests where there should be more :D

mathetake (approved):
img

PR #247, opened 2021-05-27
[WIP] Add self-hosted ARM64 runner

musaprg:
This PR introduces workflows for self-hosted ARM64 GHA runners. This PR is opened but still WIP because of validating the workflow status of GHA itself.

Signed-off-by: Kotaro Inoue <k.musaino@gmail.com>

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -30,7 +30,7 @@ jobs:
     strategy:
       fail-fast: false  # don't fail fast as sometimes failures are operating system specific
       matrix:
-        os: [ ubuntu-latest, macos-latest, self-hosted ] ]
+        os: [ ubuntu-latest, macos-latest, self-hosted ]
     steps:
       - name: "Checkout"
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
ps another cheaper way is to use travis for this dimension. Here's an another project I setup. We only need to run `make e2e` https://github.com/openzipkin/docker-alpine/blob/master/.travis.yml

codefromthecrypt:
regardless the main thing we need is a tarball URL to an arm64 binary of envoy. until we have that, it doesn't really matter as the only thing we would try to test would fail.

codefromthecrypt:
actually I will close this out with travis, as I know how to set it up and it will be free.

codefromthecrypt:
https://github.com/tetratelabs/getenvoy/pull/249

codefromthecrypt:
non-obvious problem with self-hosted runners is non-employees can't just fork the repo and test with their own GH account, unless they also pay and setup a runner. The travis alternative not only makes it free for us, but it also makes it free for those who run builds on their forks, which is something we would want to encourage vs the alternative of bashing our CI (consuming quota etc)

codefromthecrypt:
https://github.com/tetratelabs/getenvoy/pull/250 will address this. thanks for the start and please that branch's site/envoy_versions.json when there's a build we can test with (even a temporary one will help)

Issue #82, opened 2020-01-19
Arm64 builds


beriberikix:
It would be very handy to have an Arm64 build, especially for use on a Raspberry Pi. The CNCF provides builds via https://cncf.ci/ that could used as a baseline.

dkrutsko:
With the release of Amazon's [m6g](https://aws.amazon.com/blogs/aws/new-m6g-ec2-instances-powered-by-arm-based-aws-graviton2) instances, it would be really nice to have aarch64 builds available so we can take advantage of the cost savings without manually having to compile Envoy.

lizan:
I'm going to work on this in next few weeks. Need to setup CI environment with AWS arm64 instances.

morlay:
here a workflow for building envoy of istio.

https://github.com/querycap/istio-envoy-arm64

could merge it to here too?

cageyv:
Amazon also release new [t4g](https://aws.amazon.com/ru/about-aws/whats-new/2020/09/announcing-new-amazon-ec2-t4g-instances-powered-by-aws-graviton2-processors/) instances. Graviton2 expansion:) I also think, that it would be really nice to have aarch64

codefromthecrypt:
let's keep only one issue on arm64. I know this one came first, but #109 has more recent activity so closing in favor of that. please follow it and 👍 accordingly!

PR #249, opened 2021-05-28
Enable travis in preparation of arm64 builds

codefromthecrypt:
This adds travis configuration which we can validate for amd64. Once we have an envoy binary, all we need to do is switch the arch to "amd64" or "arm64-graviton2" I think we can use the latter since we don't require our tests to run docker. It is supposedly very fast. https://blog.travis-ci.com/2020-09-11-arm-on-aws

See #82

### DIFF
--- a/.licenserignore
+++ b/.licenserignore
@@ -1,3 +1,5 @@
+.travis.yaml
+
 .editorconfig
 
 /LICENSE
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
I think travis has to be merged once in order to be able to build (chicken/egg). https://docs.travis-ci.com/user/tutorial/

so, I'll merge this with amd64 (redundantly) then raise a PR to switch it to amd64.

codefromthecrypt:
ok this is green now https://travis-ci.com/github/tetratelabs/getenvoy/branches

PR #250, opened 2021-05-28
Enables e2e testing of arm64 builds

codefromthecrypt:
By using travis instead of a self-hosted github action, we don't pay anything to test arm64 and neither do non-employees who fork our repository.

This uses a temporary version list until our main one includes arm64

#109

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,16 @@
 # See https://docs.travis-ci.com/user/reference/overview/#for-a-particular-travisyml-configuration
 # run `travis lint` prior to check-in!
 os: linux   # required for arch different than amd64
-arch: amd64  # TODO arm64-graviton2 # we only only test archs not already tested with GH actions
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
good news is that everything works as expected. the e2e ran (valid goreleaser build) and failed because there's no arch

Screen Shot 2021-05-28 at 10 23 38 AM

codefromthecrypt:
I temporarily disabled travis so we don't use our quota on unrelated PRs. There's no action to take until we get a tarball

https://travis-ci.com/github/tetratelabs/getenvoy/settings

codefromthecrypt:
ok this all works. basically the binary is fine, just we are missing the tarball urls. This means the binary will be ready to support arm64 even if the main envoy-versions.json isn't updated until post-release

See https://github.com/codefromthecrypt/temprepo for details

mathetake (approved):

PR #251, opened 2021-05-28
Removes need to specify an envoy version when running

### LINKED ISSUE #106: `getenvoy run` to start the latest version by default

codefromthecrypt:
This removes the version argument so that you can `getenvoy run`
without knowing which version you want first.

Ex.
```bash
$ getenvoy run --config-yaml "admin: {access_log_path: '/dev/stdout', address: {socket_address: {address: '127.0.0.1', port_value: 0}}}"
looking up latest version
downloading https://dl.getenvoy.io/public/raw/files/getenvoy-envoy-1.18.3.p0.g98c1c9e-1p77.gb76c773-darwin-release-x86_64.tar.xz
--snip--
```

The version of Envoy run is chosen from $ENVOY_VERSION (value),
$PWD/.envoy-version (file contents), $GETENVOY_HOME/version (file
contents).

The ENV allows people who really want to control the version to do so,
and without effecting the arg parsing of `envoy`.

When first run $GETENVOY_HOME/version is initialized the the remote
version stored as "lastestVersion" in the envoy-versions.json file.
Specifically, that's here: https://getenvoy.io/envoy-versions.json

To ensure arguments don't interfere (ex. envoy also has a '-h'), this
removes the '-h' argument from `getenvoy`, replaced with a `help`
command that is run by default and can be command-scoped.

Ex. `getenvoy` or `getenvoy help` show help, and `getenvoy help run` is
scoped as you'd expect.

Fixes #106

### DIFF
--- a/e2e/getenvoy_run_test.go
+++ b/e2e/getenvoy_run_test.go
@@ -38,7 +38,7 @@ func TestGetEnvoyRun(t *testing.T) {
 
 	c := getEnvoy(`run`)
 	// Below is the minimal config needed to run envoy
-	c.args(`--config-yaml`, `admin: {access_log_path: '/dev/stdout', address: {socket_address: {address: '127.0.0.1', port_value: 0}}}`)
+	c.args("--config-yaml", "admin: {access_log_path: '/dev/stdout', address: {socket_address: {address: '127.0.0.1', port_value: 0}}}")
 
 	stdout, stderr, terminate := c.start(t, terminateTimeout)
 
--- a/e2e/getenvoy_run_test.go
+++ b/e2e/getenvoy_run_test.go
@@ -26,7 +26,6 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/tetratelabs/getenvoy/internal/tar"
-	"github.com/tetratelabs/getenvoy/internal/version"
 )
 
 const terminateTimeout = 2 * time.Minute
@@ -39,8 +38,7 @@ func TestGetEnvoyRun(t *testing.T) {
 
 	c := getEnvoy(`run`)
 	// Below is the minimal config needed to run envoy
-	// TODO allow implicit version #106
-	c.args(version.LastKnownEnvoy, `--config-yaml`, `admin: {access_log_path: '/dev/stdout', address: {socket_address: {address: '127.0.0.1', port_value: 0}}}`)
+	c.args(`--config-yaml`, `admin: {access_log_path: '/dev/stdout', address: {socket_address: {address: '127.0.0.1', port_value: 0}}}`)
 
 	stdout, stderr, terminate := c.start(t, terminateTimeout)
 
--- a/internal/cmd/install_test.go
+++ b/internal/cmd/install_test.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Tetrate
+// Copyright 2021 Tetrate
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -30,12 +30,12 @@ func TestGetEnvoyInstall_VersionValidates(t *testing.T) {
 	tests := []struct{ name, version, expectedErr string }{
 		{
 			name:        "version empty",
-			expectedErr: fmt.Sprintf(`invalid <version> argument: "" should look like "%s"`, version.LastKnownEnvoy),
+			expectedErr: fmt.Sprintf(`invalid [version] argument: "" should look like "%s"`, version.LastKnownEnvoy),
 		},
 		{
 			name:        "version invalid",
 			version:     "a.b.c",
-			expectedErr: fmt.Sprintf(`invalid <version> argument: "a.b.c" should look like "%s"`, version.LastKnownEnvoy),
+			expectedErr: fmt.Sprintf(`invalid [version] argument: "a.b.c" should look like "%s"`, version.LastKnownEnvoy),
 		},
 	}
 
--- a/internal/cmd/install_test.go
+++ b/internal/cmd/install_test.go
@@ -1,0 +1,56 @@
+// Copyright 2020 Tetrate
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/tetratelabs/getenvoy/internal/version"
+)
+
+func TestGetEnvoyInstall_VersionValidates(t *testing.T) {
+	o, cleanup := setupTest(t)
+	defer cleanup()
+
+	tests := []struct{ name, version, expectedErr string }{
+		{
+			name:        "version empty",
+			expectedErr: fmt.Sprintf(`invalid <version> argument: "" should look like "%s"`, version.LastKnownEnvoy),
+		},
+		{
+			name:        "version invalid",
+			version:     "a.b.c",
+			expectedErr: fmt.Sprintf(`invalid <version> argument: "a.b.c" should look like "%s"`, version.LastKnownEnvoy),
+		},
+	}
+
+	for _, test := range tests {
+		test := test // pin! see https://github.com/kyoh86/scopelint for why
+
+		t.Run(test.name, func(t *testing.T) {
+			c, stdout, stderr := newApp(o)
+			err := c.Run([]string{"getenvoy", "install", test.version})
+
+			// Verify the command failed with the expected error
+			require.EqualError(t, err, test.expectedErr)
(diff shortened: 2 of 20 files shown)


e2e/getenvoy_run_test.go
```diff
@@ -39,8 +38,7 @@ func TestGetEnvoyRun(t *testing.T) {
 
 	c := getEnvoy(`run`)
 	// Below is the minimal config needed to run envoy
-	// TODO allow implicit version #106
-	c.args(version.LastKnownEnvoy, `--config-yaml`, `admin: {access_log_path: '/dev/stdout', address: {socket_address: {address: '127.0.0.1', port_value: 0}}}`)
+	c.args(`--config-yaml`, `admin: {access_log_path: '/dev/stdout', address: {socket_address: {address: '127.0.0.1', port_value: 0}}}`)
```
jcchavezs, line 41: nit: I know you just copied it but ` is usually used for strings with new lines.

codefromthecrypt: ah yeah we aren't using double quotes. lemme change that and anything else you find non-idiomatic!

internal/cmd/install_test.go
```diff
@@ -0,0 +1,56 @@
...
+		test := test // pin! see https://github.com/kyoh86/scopelint for why
+
+		t.Run(test.name, func(t *testing.T) {
+			c, stdout, stderr := newApp(o)
+			err := c.Run([]string{"getenvoy", "install", test.version})
+
+			// Verify the command failed with the expected error
+			require.EqualError(t, err, test.expectedErr)
+			// GetEnvoy handles logging of errors, so we expect nothing in stdout or stderr
+			require.Empty(t, stdout)
```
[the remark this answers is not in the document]
codefromthecrypt: ah problem is we had panic bugs everywhere and require helps avoid that #137 I'd prefer to use just require for assertions as panic hunting wastes time.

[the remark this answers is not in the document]
jcchavezs: Notice require will immediately fail the test whereas assert will continue with the other assertions until the end. This is better feedback as you won't need to fix error by error but fix them all together.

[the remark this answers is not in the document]
codefromthecrypt: wounds are too fresh, only recently had the luxury of being without soft asserts. I really dislike scanning code up and down to make sure no soft assert isn't relying on state above, and that if that one fails softly another assert isn't going to panic because the reference usually enforced above is no longer good.

life's too hard as it is. TL;DR; I hear ya but can't in good faith go back at this point.

internal/cmd/run.go
```diff
@@ -87,3 +105,26 @@ func initializeRunOpts(o *globals.GlobalOpts, platform, version string) error {
 	}
 	return nil
 }
+
+// setHomeEnvoyVersion makes sure the $GETENVOY_HOME/version exists and globals.GlobalOpts HomeEnvoyVersion contains it.
+func setHomeEnvoyVersion(o *globals.GlobalOpts) error {
+	// See if there's an existing default version
+	homeVersionFile := filepath.Join(o.HomeDir, "version")
+	if data, err := os.ReadFile(homeVersionFile); err == nil {
+		o.HomeEnvoyVersion = string(data)
```
mathetake, line 114: maybe better validate the format of the data here?

internal/cmd/run_test.go
```diff
@@ -109,19 +71,65 @@ envoy args:%[3]s --admin-address-path admin-address.txt`, o.WorkingDir, o.EnvoyP
...
+	defer cleanup()
+
+	o.HomeEnvoyVersion = ""
+	require.NoError(t, os.WriteFile(filepath.Join(o.HomeDir, "version"), []byte("a.a.a"), 0600))
+
+	c, _, _ := newApp(o)
+	err := c.Run([]string{"getenvoy", "run"})
+
+	// Verify the command failed with the expected error
+	require.EqualError(t, err, fmt.Sprintf(`invalid version in "$GETENVOY_HOME/version": "a.a.a" should look like "%s"`, version.LastKnownEnvoy))
```
mathetake, line 118: Add check `o.HomeEnvoyVersion` is set to the wanted value?

internal/cmd/run.go
```diff
@@ -87,3 +105,26 @@ func initializeRunOpts(o *globals.GlobalOpts, platform, version string) error {
 	}
 	return nil
 }
+
+// setHomeEnvoyVersion makes sure the $GETENVOY_HOME/version exists and globals.GlobalOpts HomeEnvoyVersion contains it.
+func setHomeEnvoyVersion(o *globals.GlobalOpts) error {
```
mathetake, line 110: Prefer have this function as a method of `globals.GlobalOpts`.. WDYT?

internal/envoy/version.go
```diff
@@ -0,0 +1,76 @@
...
+	"strings"
+
+	"github.com/tetratelabs/getenvoy/internal/globals"
+	"github.com/tetratelabs/getenvoy/internal/version"
+)
+
+var (
+	versionVarName      = "$ENVOY_VERSION"
+	wdVersionFileName   = filepath.Join("$PWD", ".envoy-version")
+	homeVersionFileName = filepath.Join("$GETENVOY_HOME", "version")
```
mathetake, line 30: I was confused by these names so I would suggest 
```suggestion
	versionSourceVersionVar      = "$ENVOY_VERSION"
	versionSourceWDVersionFile   = filepath.Join("$PWD", ".envoy-version")
	versionSourceHomeVersionFile = filepath.Join("$GETENVOY_HOME", "version")
```

internal/envoy/version.go
```diff
@@ -0,0 +1,76 @@
...
+		return
+	} else if !os.IsNotExist(err) {
+		return
+	}
+
+	// Priority 3: $GETENVOY_HOME/version
+	source = homeVersionFileName
+	v = homeVersion
+	err = nil
+	return
```
mathetake, line 69: less lines, the better
```suggestion
	// Priority 1: $ENVOY_VERSION
	if v, ok := os.LookupEnv("ENVOY_VERSION"); ok {
		return v, versionSourceVersionVar, nil
	}

	// Priority 2: $PWD/.envoy-version
	data, err := os.ReadFile(".envoy-version")
	if err == nil {
		return string(data), versionSourceWDVersionFile, nil
	} else if !os.IsNotExist(err) {
		return "", "", err
	}

	// Priority 3: $GETENVOY_HOME/version
	return homeVersion, versionSourceHomeVersionFile, nil
```

mathetake (commented):
First shot. Nice work!

internal/cmd/run.go
codefromthecrypt: oh yeah.. true we can't assume people are using the schema. good point.

internal/cmd/run.go
codefromthecrypt: so far I've not had any cmd arg parsing in the globals file.. that's why this is here..

internal/envoy/version.go
codefromthecrypt: how about using the word default instead of source? I find the word source awkward as it feels like source code..

internal/envoy/version.go
codefromthecrypt: yes completely agree. especially there are if statements we can never cover in non-absurd tests :D let's keep the line count down.

internal/envoy/version.go
mathetake: sounds better to me 😄

internal/cmd/run.go
mathetake: OK, got it

internal/cmd/install_test.go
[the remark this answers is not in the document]
codefromthecrypt: ps after a weekend of pondering, I think there were a few things on #137 that led things to be very painful. Even if some are invalidated now, the summary of fixing failed assertions > fixing panics holds.

first, consider a flow like this:

assert not empty versions
assert not empty versions.labels

The overall problem is that the second assert will panic if it isn't a hard assert. There are many forms of this problem especially around if foo, err := stuff. Ex you soft assert err is not nil then if that changes later panic on foo.

1. iirc ginkgo (which luckily is no longer here) would crash strangely on panic. 
2. library upgrades sometimes change structs from chains to chain of pointers
3. panics aren't well formatted, and because they happen later, it is time consuming to figure out what soft assert failed leading to this

The general optimization, in other words, is code for less difficult upgrades, not for "auto-filling tests". Ex I'm aware of the practice of writing a test with a bunch of failed assertions and then try to finish the code. It is indeed slightly faster to do that vs click re-run. However, that's optimizing not just for folks who code like that, but it is in reverse from the concerns above. Basically it makes it quick to add code not well considered and tricky to fix bugs later.

So anyway I really appreciate your feedback and hope that this doesn't come off draconian or anything I generally want to accept changes requested even if they are low-level framework, but I think the "cure" of changing everything to soft asserts isn't something I would do personally as I can't in good faith say this is a nice thing to do for future maintainers including me.

codefromthecrypt:
@mathetake PTAL sorry this change looks bigger, but >350 lines are tests

mathetake (approved):
img

e2e/getenvoy_run_test.go
codefromthecrypt: done

codefromthecrypt:
Thanks for the feedback and help @jcchavezs and @mathetake another round on the way soon

PR #253, opened 2021-06-02
Repurposes install -> use <version>

### LINKED ISSUE #248: command overhaul: 3 commands!
Hopefully, the last cleanup in pursuit of simplest path (before we release stuff!)

goals:
* intuitive as possible (little jargon)
* zero steps (no "run this before that")
* auto-completion friendly (avoid commands that share same first character)

end state:
* `getenvoy versions` (lists installed versions, -a includes remote ones not yet installed)
* `getenvoy use <version>` (assign current version, installing as necessary)
* `getenvoy run -c envoy.yaml` (run current version, installing as necessary)
current version is $ENVOY_VERSION, $PWD/.envoy-version or $ENVOY_HOME/version, which defaults to envoy-versions.json/latestVersion

current state (still not released):
* `run <version> -c envoy.yaml` (run installing on demand as necessary)
* `versions` (lists all versions)
* `install <version>` (install if not already)
* `installed` (lists all installed versions)
current version requires looking at the versions list

action 1: remove `<version>` from run, and read $ENVOY_VERSION $PWD/.envoy-version or $ENVOY_HOME/version
* simplifies `run` syntax and the need to know what version you'd want
* removes all questions about args to envoy: everything after `run` is interpreted by envoy
* $ENVOY_VERSION makes runs in CI for a non-latest version single-step
* on first run $ENVOY_HOME/version is populated from envoy-versions.json/latestVersion

action 2: repurpose `install -> use <version>`
* if we introduce a command to assign a version, there's a chance it isn't there, ends up the same as install
* rename install to use, and upon `use <version>` install it as necessary
* new change is to update $PWD/.envoy-version or $ENVOY_HOME/version if present

action 3: don't make a flag to write .envoy-version
* `getenvoy use 1.18.3 --local` is only 1 more char than `echo 1.18.3 > .envoy-version`
* by waiting to define a flag, we don't need to choose between bad jargon choices ex "local" or "cwd"
* by waiting, we can get more input from end users

action 4: fold `installed` into `versions -a/--all`
* `getenvoy install` collides on auto-complete with `getenvoy installed` which is annoying
* `getenvoy installed` output is different than `getenvoy versions` mainly because we didn't have release date locally
* during install, we can persist the "releaseDate" field into the $ENVOY_HOME/versions/$version directory create date
* at that point `getenvoy versions` shows the installed ones and `getenvoy versions -a` shows all including remote, both knowing the release date

codefromthecrypt:
This removes `getenvoy install` for `getenvoy use`. The most essential change is that this assigns the version `getenvoy run` uses upon invocation. Specifically, the new change is it updates $PWD/.envoy-version or $ENVOY_HOME/version if present.

Notes:
* This intentionally doesn't crash `getenvoy list` when a version is targeted with a version file, but not present yet
* `getenvoy run` continues to install on-demand as it did before
* For edge cases that need to pre-install multiple versions, simply run `getenvoy use` in a loop, with the intended default version last.

Changes to `getenvoy versions` output:
* this indicates what version is current and why similar to `pyenv`
* this removes header and error when no versions are present for parsing

Usage example:
```bash
$ getenvoy versions
* 1.18.3 2021-05-11 (set by $GETENVOY_HOME/version)
$ getenvoy versions -a
* 1.18.3 2021-05-11 (set by $GETENVOY_HOME/version)
  1.17.3 2021-05-11
  1.16.4 2021-05-11
  1.15.5 2021-05-11
  1.14.7 2021-04-15
$ getenvoy use 1.14.7
downloading https://dl.getenvoy.io/public/raw/files/getenvoy-envoy-1.14.7.p0.gbc15868-1p74.gbb8060d-darwin-release-x86_64.tar.xz
$ getenvoy versions
  1.18.3 2021-05-11
* 1.14.7 2021-04-15 (set by $GETENVOY_HOME/version)
$ echo 1.15.5 > .envoy-version
$ getenvoy versions
  1.18.3 2021-05-11
  1.14.7 2021-04-15
$ getenvoy run
downloading https://dl.getenvoy.io/public/raw/files/getenvoy-envoy-1.15.5.p0.g623479e-1p74.gbb8060d-darwin-release-x86_64.tar.xz
--snip--
$ getenvoy versions
  1.18.3 2021-05-11
* 1.15.5 2021-05-11 (set by $PWD/.envoy-version)
  1.14.7 2021-04-15
$ ENVOY_VERSION=1.18.3 getenvoy versions
* 1.18.3 2021-05-11 (set by $ENVOY_VERSION)
  1.15.5 2021-05-11
  1.14.7 2021-04-15
```

Fixes #248


e2e/getenvoy_use_test.go
```diff
@@ -26,32 +27,39 @@ import (
...
 
 	t.Run("not yet installed", func(t *testing.T) {
-		stdout, stderr, err := getEnvoy("--home-dir", homeDir, "install", version.LastKnownEnvoy).exec()
+		stdout, stderr, err := getEnvoy("--home-dir", homeDir, "use", version.LastKnownEnvoy).exec()
 
 		require.NoError(t, err)
 		require.Regexp(t, `downloading https:.*tar.*`, stdout)
 		require.Empty(t, stderr)
 
+
```
codefromthecrypt, line 42:
```suggestion
```

internal/cmd/versions_cmd_test.go
```diff
@@ -35,72 +36,171 @@ func TestGetEnvoyVersions_NothingYet(t *testing.T) {
...
+}
 
-	oneTwoTwo := filepath.Join(o.HomeDir, "versions", "1.2.2")
-	require.NoError(t, os.MkdirAll(oneTwoTwo, 0700))
-	morerequire.RequireSetMtime(t, oneTwoTwo, "2020-12-31")
+func TestGetEnvoyVersions_CurrentVersion(t *testing.T) {
+	o, cleanup := setupTestVersions(t)
+	defer cleanup()
+
+	// These tests depend on prior state, so execute sequentially (don't change the order).
```
codefromthecrypt, line 64: @jcchavezs fyi based on your nudging on the former PR, I made this a table test because all the expected results can fit on a screen, and scanning the difference is easier this way. It works well also because there is no conditional in the actual test block, even if some of the setup is gnarly to enclose in an anonymous function. In other words, for this one, feels better this way!

internal/envoy/install_test.go
```diff
@@ -127,28 +128,6 @@ func TestInstallIfNeeded_Validates(t *testing.T) {
 	}
 }
 
-// progressReader emits a progress bar, when the length is known
-func TestProgressReader(t *testing.T) {
```
codefromthecrypt, line 131: moved these as they were sandwiched between other tests

codefromthecrypt:
updated the description with a usage walkthrough

mathetake:
will start the review in a couple hours..

internal/envoy/version_test.go
```diff
@@ -154,3 +213,10 @@ func TestCurrentVersion_Validates(t *testing.T) {
 		require.EqualError(t, err, fmt.Sprintf(`invalid version in "$ENVOY_VERSION": "c.c.c" should look like "%s"`, version.LastKnownEnvoy))
 	})
 }
+
+func requireWorkingDirVersion(t *testing.T, v string) (removeTempDir func()) {
+	tempDir, removeTempDir := morerequire.RequireNewTempDir(t)
```
mathetake, line 218: I'm a bit surprised to see that removeTempDir is properly assigned even though it looks like `removeTempDir` is shadowed here by `:=`.. I didn't know this behavior..https://play.golang.org/p/m0H6KZAm6aw

internal/cmd/versions_cmd_test.go
```diff
@@ -35,72 +36,171 @@ func TestGetEnvoyVersions_NothingYet(t *testing.T) {
...
+	// The different version values are intentional to ensure overrides work properly.
+	for _, tt := range []struct {
+		name, expected string
+		setup          func() func()
+	}{
+		{"no current version", `  1.2.2 2021-01-31
```
mathetake, line 70: Could you explicitly give the field name for cases? Now it is a bit hard to read..

```suggestion
		{name: "no current version", expected:`  1.2.2 2021-01-31
```

internal/envoy/version_test.go
codefromthecrypt: I'll try make it more explicit.. this is a bit of a pain because we can leak a temp dir if not careful..

mathetake (approved):

internal/envoy/version_test.go
codefromthecrypt: ps this is another reason why I had it sequentially before :D anyway this function will look terrible but more clear

internal/cmd/versions_cmd_test.go
codefromthecrypt: I can't read the tests anymore especially as go makes the indent nuts. I'm reverting matrixifying this test

internal/envoy/version_test.go
codefromthecrypt: I extracted a new function for this

internal/test/morerequire/morerequire.go
```diff
@@ -72,3 +72,25 @@ func RequireSetenv(t *testing.T, key, value string) func() {
 		require.NoError(t, err, `error reverting env variable %s=%s`, key, previous)
 	}
 }
+
+// RequireChdirIntoTemp creates a new temp directory and cleans it up on with the returned function.
+func RequireChdirIntoTemp(t *testing.T) (cleanup func()) {
+	var cleanups []func()
+	cleanup = func() {
```
codefromthecrypt, line 79: @mathetake this is best I could figure out for how to deal with not leaking

internal/test/morerequire/morerequire.go
```diff
@@ -72,3 +72,25 @@ func RequireSetenv(t *testing.T, key, value string) func() {
 		require.NoError(t, err, `error reverting env variable %s=%s`, key, previous)
 	}
 }
+
+// RequireChdirIntoTemp creates a new temp directory and cleans it up on with the returned function.
```
codefromthecrypt, line 76:
```suggestion
// RequireChdirIntoTemp creates a new temp directory and cleans it up on the returned function.
```

internal/cmd/versions_cmd_test.go
codefromthecrypt: I ended up reverting this as it made things worse. We definitely use matrices for simple tests and will continue to!

internal/test/morerequire/morerequire.go
```diff
@@ -72,3 +72,25 @@ func RequireSetenv(t *testing.T, key, value string) func() {
...
+	cleanups = append(cleanups, func() {
+		require.NoError(t, os.Chdir(wd))
+	})
+
+	tempDir, removeTempDir := RequireNewTempDir(t)
+	cleanups = append(cleanups, removeTempDir)
+
+	require.NoError(t, os.Chdir(tempDir))
+	return
+}
```
mathetake, line 96:
```suggestion

func RequireChdirIntoTemp(t *testing.T) func() {
	wd, err := os.Getwd()
	require.NoError(t, err)
	tempDir, removeTempDir := RequireNewTempDir(t)
	return func() {
		require.NoError(t, os.Chdir(wd))
		removeTempDir()
		require.NoError(t, os.Chdir(tempDir))
	}
}
```

How about this?

codefromthecrypt: it forgot to chdir :D but sure can composite the func I think that's the thing you mainly want!

codefromthecrypt: added a commit similar!

codefromthecrypt: https://github.com/tetratelabs/getenvoy/commit/01331b8dd1bed331a273ec7b3a4c28a4d0fa8420

PR #257, opened 2021-06-02
Buckets user-agent only to release or non-release without an internal flag

codefromthecrypt:
This also captures that netlify won't help us know which user-agents are in use without custom edge handlers.

### DIFF
--- a/internal/envoy/install.go
+++ b/internal/envoy/install.go
@@ -36,7 +36,7 @@ func InstallIfNeeded(o *globals.GlobalOpts, p, v string) (string, error) {
 	switch {
 	case os.IsNotExist(err):
 		var ev version.EnvoyVersions // Get version metadata for what we will install
-		ev, err = GetEnvoyVersions(o.EnvoyVersionsURL, o.UserAgent)
+		ev, err = GetEnvoyVersions(o.EnvoyVersionsURL, p, v)
 		if err != nil {
 			return "", err
 		}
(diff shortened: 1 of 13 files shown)


internal/envoy/install.go
```diff
@@ -55,7 +55,7 @@ func InstallIfNeeded(o *globals.GlobalOpts, p, v string) (string, error) {
 		}
 
 		fmt.Fprintln(o.Out, "downloading", tarballURL) //nolint
-		if err = untarEnvoy(installPath, tarballURL, o.UserAgent); err != nil {
+		if err = untarEnvoy(installPath, tarballURL, p, v); err != nil { //nolint
```
codefromthecrypt, line 58: lint hates me.. it doesn't like `=` and also doesn't like `:=`. To avoid making randomly `e`, I'm punting //nolint

internal/envoy/install.go
```diff
@@ -36,7 +36,7 @@ func InstallIfNeeded(o *globals.GlobalOpts, p, v string) (string, error) {
 	switch {
 	case os.IsNotExist(err):
 		var ev version.EnvoyVersions // Get version metadata for what we will install
-		ev, err = GetEnvoyVersions(o.EnvoyVersionsURL, o.UserAgent)
+		ev, err = GetEnvoyVersions(o.EnvoyVersionsURL, p, v)
```
mathetake, line 39: I realized that we invoke httpGet twice in this path.. here and the inside of untarEnvoy below.. would it be better to call once and reuse the result?

mathetake (commented):
LGTM modulo the comment above

internal/envoy/install.go
codefromthecrypt: it is invoked twice, but different URLs, the tarball is a different hostname in real life (even if in mock we are using the same origin). In any case, it is different URL for "envoy-versions.json" vs "/path/to/tar.xz". Make sense?

mathetake: ah i see, thanks for clarifying!

PR #265, opened 2021-06-03
Runs Envoy in the current working directory

### LINKED ISSUE #263: Revert `getenvoy run` to excute envoy in the CWD
Right now, `getenvoy run` executes envoy in a different directory (the "run" directory). This was a mistake in hindsight because it makes running envoy more difficult as the config file must be an absolute path.

Ex. this won't work
```
$ getenvoy run -c ./my-config.yml
```

This will
```bash
$ getenvoy run -c $PWD/my-config.yml
```

However, the $PWD trick won't work with included files, as those are resolved by envoy based on the cwd, not relative to the path of the config file.

Ex. relative paths like this:
```yaml
                          vm_config:
                            vm_id: "my_vm_id"
                            runtime: "envoy.wasm.runtime.v8"
                            code:
                              local:
                                filename: "./examples/http_body/main.go.wasm"
```

We need to change the behavior to execute in the current working directory and backfill an E2E test that proves file resolution relative to the config file works.

codefromthecrypt:
This makes sure Envoy is run in the current working directory, so that
it is easier to use and configuration can locate relative paths.

Fixes #263

### DIFF
--- a/e2e/admin_test.go
+++ b/e2e/admin_test.go
@@ -70,7 +70,7 @@ func (c *adminClient) getJSON(path string, v interface{}) error {
 	return json.Unmarshal(body, v)
 }
 
-func get(url string) ([]byte, error) {
+func httpGet(url string) ([]byte, error) {
 	resp, err := http.Get(url)
 	if err != nil {
 		return nil, err
--- a/e2e/admin_test.go
+++ b/e2e/admin_test.go
@@ -22,37 +22,20 @@ import (
 	"net/http"
 )
 
-// adminAPI represents Envoy Admin API.
-type adminAPI interface {
-	isReady() (bool, error)
-	getStats() (*stats, error)
-}
-
-// stats represents Envoy response to `/stats?format=json` endpoint.
-type stats struct {
-	Metrics []metric `json:"stats"`
-}
-
-// metric represents recorded value of a single metric.
-type metric struct {
-	Name  string  `json:"name"`
-	Value float64 `json:"value"`
-}
-
-// newClient returns a new client for Envoy Admin API.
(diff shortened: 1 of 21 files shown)


e2e/admin_test.go
```diff
@@ -22,37 +22,20 @@ import (
 	"net/http"
 )
 
-// adminAPI represents Envoy Admin API.
-type adminAPI interface {
```
codefromthecrypt, line 26: fyi even package private interfaces interfere with linters who otherwise can find dead code!

e2e/admin_test.go
```diff
@@ -61,23 +44,40 @@ func (c *client) isReady() (bool, error) {
...
 	if err != nil {
-		return nil, err
+		return err
 	}
-	var s stats
-	err = json.Unmarshal(body, &s)
+	return json.Unmarshal(body, v)
+}
+
+func get(url string) ([]byte, error) {
```
mathetake, line 73: nit want to prefix http.. 😄 
```suggestion
func httpGet(url string) ([]byte, error) {
```

internal/envoy/run_test.go
```diff
@@ -37,7 +37,8 @@ func TestRuntime_Run(t *testing.T) {
 
 	runsDir := filepath.Join(tempDir, "runs")
 	fakeTimestamp := "1619574747231823000"
```
mathetake, line 39: do we need this then?

mathetake (approved):

internal/envoy/run_test.go
codefromthecrypt: nope.. killable

Issue #105, opened 2020-09-30
Support getenvoy CLI on Windows


sunjayBhatia:
With the Alpha release of Envoy on Windows approaching, we think it would be useful for users to eventually be able to use the getenvoy CLI tool to fetch and run Envoy etc. once Envoy on Windows is GA. We are opening this issue is to start the conversation about the steps needed to get to Windows support in getenvoy.

As far as I can tell, the high-level steps I can glean from looking around the getenvoy repos are:
- Augment https://github.com/tetratelabs/getenvoy-package to support building on Windows
- Augment this repo to build the getenvoy CLI on Windows and have it fetch Envoy-Windows builds
- Update documentation to add usage examples for Windows

getenvoy/Tetrate package distribution channels could also be augmented for Windows, distributing Envoy via chocolatey or some other package manager

If there is anything high-level that maintainers of getenvoy see is critical and missing, please let us know (and of course if you are willing to support getenvoy on Windows at all!)

sunjayBhatia:
We're getting close to a Beta release of Envoy on Windows as well, do the maintainers have any thoughts on this work?

lizan:
@sunjayBhatia Sorry for being late on this, yes we definitely love to see Windows support of GetEnvoy! The outline above looks great to me.

sunjayBhatia:
Awesome, thanks @lizan, we will get some PRs going to hopefully incrementally get this going

codefromthecrypt:
we are also using GH actions consistently here, now, and would need to add windows to the matrix for that to work. Are you still interested in contributing?

codefromthecrypt:
It looks like windows is officially supported with the most recent version of envoy. We may be able to play a trick to test this by extracting the binary from the upstream docker image https://blog.envoyproxy.io/general-availability-of-envoy-on-windows-267e4544994a

codefromthecrypt:
added this issue to ask envoy to retain the binaries they already build https://github.com/envoyproxy/envoy/issues/16830

sunjayBhatia:
Sorry unfortunately internal company priorities changed for me so I'm light on time to contribute to this at the moment!

codefromthecrypt:
@sunjayBhatia no problem. Tetrate bought me a windows laptop so I can look into this personally.. and I am now. I hope by the end of the week things will be working even if using a temporary envoy binary tarball. Will make notes here with progress.

codefromthecrypt:
[...]

PR #267, opened 2021-06-05
Fixes compilation break on Windows

codefromthecrypt:
This doesn't fix all the tests or make binaries avaiable. It just makes
it possible to compile on windows.

See #105


internal/envoy/debug/node.go
```diff
@@ -48,7 +48,7 @@ type nodeCollection struct {
 }
 
 // Don't wait forever. This has hung on macOS before
-const processTimeout = 3 * time.Second
+const processTimeout = 300 * time.Second
```
codefromthecrypt, line 51:
```suggestion
const processTimeout = 3 * time.Second
```

internal/envoy/debug/lsof_stat.go
```diff
@@ -0,0 +1,38 @@
...
+
+package debug
+
+import (
+	"fmt"
+	"syscall"
+)
+
+func newOpenFileStat(command string, pid int32, user string, fd uint64, name string) *openFileStat {
+	ofstat := openFileStat{
```
codefromthecrypt, line 25: I went back and forth on this in my head.. I feel like the constructor pattern feels less odd here when guarding against OS, but there is code duplication initializing the struct and that is easy to drift in the windows variant. Another way is to make a "populateNodeAndSize" function or similar and reduce the surface area.. wdyt?

codefromthecrypt:
on second thought I think this shouldn't be a PR as we shouldn't merge anything until there is CI to cover it. Even if e2e tests can't complete, yet due to missing binary, we can run unit tests

codefromthecrypt:
closing until ready

internal/envoy/debug/lsof_stat.go
codefromthecrypt: this is no longer relevant as we dumped lsof anyway

PR #269, opened 2021-06-07
Simplifies hooks into shutdown-only

codefromthecrypt:
Before, we had several types of hooks, but most were only needed to
support tee'ing stdout and stderr into log files. This centralizes that
logic and renames things around "shutdown hooks" which are easier to
understand and congruent with envoy console output.

Ex.
```
[2021-06-07 10:07:43.912][3272041][info][main] [source/server/server.cc:790] shutting down server instance
```


internal/envoy/runtime.go
```diff
@@ -46,16 +47,39 @@ type Runtime struct {
...
+	}
+	// Envoy's run directory is mutable, so it is fine to write the admin address there.
+	r.adminAddressPath = filepath.Join(r.opts.RunDir, "admin-address.txt")
+	r.cmd.Args = append(r.cmd.Args, flag, r.adminAddressPath)
+	return nil
+}
+
 // GetAdminAddress returns the current admin address in host:port format, or empty if not yet available.
-// Exported for debug.EnableEnvoyAdminDataCollection, which is always on due to debug.EnableAll.
+// Exported for shutdown.enableEnvoyAdminDataCollection, which is always on due to shutdown.EnableAll.
```
codefromthecrypt, line 82:
```suggestion
// Exported for shutdown.enableEnvoyAdminDataCollection, which is always on due to shutdown.EnableHooks.
```

internal/cmd/usage_md_test.go
```diff
@@ -56,6 +56,7 @@ func TestUsageMarkdownMatchesCommands(t *testing.T) {
 	want = strings.ReplaceAll(want, "   ", "") // remove leading indent until urfave/cli#1275
 
 	have, err := os.ReadFile(siteMarkdownFile)
+
```
mathetake, line 59: seems unnecessary change 😄

internal/envoy/shutdown/shutdown.go
```diff
@@ -24,20 +24,8 @@ import (
...
-		enableEnvoyLogCollection,
-		enableNodeCollection,
-	} {
-		if err := debug(r); err != nil {
-			errs = append(errs, err)
-		}
-	}
-	return errs
-}
+// EnableHooks is a list of functions that enable hooks in this package
```
mathetake, line 27: > in this package

curious, is there any other "hooks" outside this package?

mathetake (approved):
LGTM

image

internal/envoy/shutdown/shutdown.go
codefromthecrypt: no there aren't.. I'll take out the unnecessary words!

codefromthecrypt:
Thanks for the 🕶️

PR #270, opened 2021-06-07
Centralizes use of go context

codefromthecrypt:
Before, we had many call paths that could accept a go context, but we
didn't give one. Since urfave supports context, we should use it.

This centralizes shutdown hook deadline as now+5s and logs accordingly.

Ex.

```
invoking shutdown hooks with deadline [2021-06-07 13:49:47.226]
[2021-06-07T05:49:42.233Z] "GET /contention HTTP/1.1" 200 - 0 95 1 - "192.168.0.116" "Go-http-client/1.1" "-" "127.0.0.1:62884" "-"
sending interrupt to envoy (pid=61302)
[2021-06-07 13:49:47.269][4272150][warning][main] [external/envoy/source/server/server.cc:622] caught SIGINT
```

This also makes fetching parallel per #271


codefromthecrypt:
https://github.com/tetratelabs/getenvoy/pull/271 layers over this and does things in parallel

mathetake (approved):

codefromthecrypt:
Thanks for the look @mathetake!

PR #279, opened 2021-06-11
Makes code more OS portable

codefromthecrypt:
Almost ready for #105


internal/cmd/run_test.go
```diff
@@ -95,7 +97,7 @@ func TestGetEnvoyRun_ReadsHomeVersionFile(t *testing.T) {
 	runWithoutConfig(t, c)
 
 	// No implicit lookup
-	require.NotContains(t, o.Out.(*bytes.Buffer).String(), "looking up latest version\n")
+	require.NotContains(t, o.Out.(*bytes.Buffer).String(), "looking up latest version"+lf)
```
mathetake, line 100: I guess we could use `fmt.Sprintln` to produce a portable string with a trailing new line..

codefromthecrypt: except it doesn't :) inside it does this

```
	p.buf.writeByte('\n')
```

codefromthecrypt: it is very strange that go decided to not include portable linefeed support 🤷

mathetake: funny 😂

codefromthecrypt: that said, we could make our own Sprintln that does..

internal/cmd/run_test.go
```diff
@@ -29,11 +28,14 @@ import (
 
 	rootcmd "github.com/tetratelabs/getenvoy/internal/cmd"
 	"github.com/tetratelabs/getenvoy/internal/globals"
+	"github.com/tetratelabs/getenvoy/internal/moreos"
 	"github.com/tetratelabs/getenvoy/internal/test"
 	"github.com/tetratelabs/getenvoy/internal/test/morerequire"
 	"github.com/tetratelabs/getenvoy/internal/version"
 )
 
+const lf = moreos.LineSeparator
```
mathetake, line 37: what does "f" of lf mean here?

mathetake (approved):

internal/cmd/run_test.go
codefromthecrypt: meh... problem with doing this is that IDEs won't syntax highlight the format string

internal/cmd/run_test.go
codefromthecrypt: renamed to ls!

codefromthecrypt: actually "ln" looks nicer as matches println

codefromthecrypt:
ok also backfilled a ln test, though it will need a slight adjustment when completing windows it shows why it is there.

internal/moreos/moreos_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	f, err := os.Stat(bin)
+	require.NoError(t, err)
+
+	require.False(t, isExecutable(f))
+}
+
+func TestLineSeparator(t *testing.T) {
+	stdout := new(bytes.Buffer)
+	cmd := exec.Command("echo", "cats")
```
codefromthecrypt, line 58: in windows, we need to have an if statement and do `"cmd.exe", "/c", "echo cats"`. This proves that it works and helps ensure we catch newline things without running envoy

codefromthecrypt:
Thanks for looking  and asking questions @mathetake !

PR #281, opened 2021-06-16
Refactors JSON schema and adds SHA256 verification of tarballs

codefromthecrypt:
This change allows us increased confidence that tarballs downloaded were
what we expected. Notably, `getenvoy use N.N.N` will fail now, if the
`sha256sum` doesn't match what's in the json key for that filename.

This refactors out Envoy-specific terminology from the JSON schema used
to validate envoy-versions.json. It is now called
release-versions-schema.json.

This adds a new key sha256sums in the schema and backfills data for the
existing versions.

The key name "sha256sums" comes from the `sha256sum` command. This is
SHA-256 instead of SHA-512 as it is a ubuquitous hash algo, for example,
used by default also in goreleaser.

This doesn't use a more generic term like "hash" because doing so would
cause unnecessary confusion on what hash it is (ex MD-5?).

The map "sha256sums" allows for easy indexing, and using the filename
instead of the URL allows less changes when rehosting the exact same
files.


site/usage.md
```diff
@@ -8,7 +8,7 @@ choose one, invoke `getenvoy use 1.18.3`. This installs into
 
 You may want to override `$ENVOY_VERSIONS_URL` to supply custom builds or
 otherwise control the source of Envoy binaries. When overriding, validate
-your JSON first: https://getenvoy.io/envoy-versions-schema.json
+your JSON first: https://getenvoy.io/release-versions-schema.json
```
codefromthecrypt, line 11: I will add a redirect for this post review and pre-merge

internal/envoy/install.go
```diff
@@ -84,19 +94,42 @@ func verifyEnvoy(installPath string) (string, error) {
...
 
-func untarEnvoy(ctx context.Context, dst, url, p, v string) error { // dst, src order like io.Copy
-	resp, err := httpGet(ctx, url, p, v)
+type digester struct {
+	r io.Reader
+	h hash.Hash
+	e error
+}
+
+func (d *digester) Read(p []byte) (n int, err error) {
```
codefromthecrypt, line 103: I couldn't find another non-buffering way to do an incremental digest

internal/version/version.go
```diff
@@ -12,40 +12,62 @@
...
 package version
 
 import _ "embed" // We embed the Envoy version so that we can cache it in CI
 
+var getEnvoy = "dev"
+
 // GetEnvoy is the version of the CLI, used in help statements and HTTP requests via "User-Agent".
-// Override this via "-X github.com/tetratelabs/getenvoy/internal/version.GetEnvoy=XXX"
-var GetEnvoy = "dev"
+// Override this via "-X github.com/tetratelabs/getenvoy/internal/version.getEnvoy=XXX"
```
codefromthecrypt, line 23: I checked this works

internal/version/version.go
```diff
@@ -12,40 +12,62 @@
...
+
+// SHA256Sum is a SHA-256 lower-hex hash. Ex. "1274f55b3022bc1331aed41089f189094e00729981fe132ce00aac6272ea0770"
+type SHA256Sum string
+
+// ReleaseDate is the publish date of the release Version. Ex. "2021-05-11"
+type ReleaseDate string
+
+// Release primarily maps available Tarballs for a Version
+type Release struct {
+	ReleaseDate
```
mathetake, line 69: IMHO we should avoid embedding structs as much as possible unless it's necessary (for example, in the case where we want to provide default implementation of interfaces via embedding).

mathetake: often this make it difficult at call sites.

internal/version/version.go
```diff
@@ -12,40 +12,62 @@
...
+type Tarball string
+
+// TarballURL is the HTTPS URL to the Tarball. SHA256Sums must include its base name.
+type TarballURL string
+
+// SHA256Sum is a SHA-256 lower-hex hash. Ex. "1274f55b3022bc1331aed41089f189094e00729981fe132ce00aac6272ea0770"
+type SHA256Sum string
+
+// ReleaseDate is the publish date of the release Version. Ex. "2021-05-11"
+type ReleaseDate string
```
mathetake, line 65: I'm skeptical that we need these aliasing.. I see a lot of `string(...)` anywhere.

internal/cmd/versions.go
```diff
@@ -80,10 +80,8 @@ func NewVersionsCmd(o *globals.GlobalOpts) *cli.Command {
 }
 
 type versionReleaseDate struct {
-	// version ex "1.15.5"
-	version string
-	// releaseDate ex "2021-05-11"
-	releaseDate string
+	version.Version
+	version.ReleaseDate
```
mathetake, line 84: same here for embedding and aliasing.

internal/version/version.go
mathetake: my point is I think having descriptive field or variable names is enough vs bother doing type conversion a lot

mathetake (commented):
other than the comment above, the change LGTM!

internal/envoy/install.go
```diff
@@ -48,16 +52,22 @@ func InstallIfNeeded(ctx context.Context, o *globals.GlobalOpts, p, v string) (s
 			return "", fmt.Errorf("couldn't find version %q for platform %q", v, p)
 		}
 
+		tarball := version.Tarball(path.Base(string(tarballURL)))
+		sha256Sum := ev.SHA256Sums[tarball]
+		if sha256Sum == "" {
```
mathetake, line 57: more precisely I suggest check `len(sha256Sum) == 64` 😄

internal/version/version.go
codefromthecrypt: I kindof agree, though I would say it is pretty limited and this found a bug or two.

codefromthecrypt: the maintenance of this vs json schema is far less hairy with type aliases

internal/version/version.go
codefromthecrypt: that's fine I just noticed it was less typing and work

internal/version/version.go
codefromthecrypt: the other thing is the these are coordinated map keys. the aliases do help on that part

mathetake: OK fair enough! better leave them as-is

site/envoy-versions-schema.json
```diff
@@ -1,96 +0,0 @@
-{
```
mathetake, line 1: is this going to be a breaking change?

codefromthecrypt: we don't have anything except docs referring to the location of this file. but I will rename it back for this PR anyway!

internal/version/version.go
```diff
@@ -12,40 +12,62 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package version declares types for each string to keep strict coupling to the JSON schema
```
codefromthecrypt, line 15: this is a comment which can be used to squarely blame adrian for the type aliases later. If they become a problem I won't complain much about reverting them.. I also usually find type aliases annoying

site/envoy-versions-schema.json
```diff
@@ -73,7 +98,7 @@
       },
       "required": [
         "releaseDate",
-        "downloads"
+        "tarballs"
```
codefromthecrypt, line 101: typo!

site/envoy-versions-schema.json
codefromthecrypt: ps to answer no.. it isn't a breaking change, but before our json schema had a bug in it, so there's that also :D

codefromthecrypt:
3Q for the help @mathetake!

PR #283, opened 2021-06-18
Add the reference to Envoy in the run command

peterj:


codefromthecrypt:
ah the reference is is no longer used since v0.3.0. you may want to upgrade https://github.com/tetratelabs/getenvoy/releases/tag/v0.3.0

peterj:
hmmm, I thought I upgraded to 0.4.0 and it didn’t work there.

peterj:
Looks like I was running an older version .. nevermind. closing this.

PR #285, opened 2021-07-01
Fix check recipe so it won't hang

mathetake:


codefromthecrypt (approved):
hopefully I won't bomb you with reviews this long, but great anyway :D

PR #288, opened 2021-07-06
Changes env naming convention to be more normal

codefromthecrypt:


.github/workflows/release.yaml
```diff
@@ -30,7 +30,6 @@ jobs:
       - name: "Run GoReleaser"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          HOMEBREW_FUNC-E_GITHUB_TOKEN: ${{ secrets.FUNC-E_CI_GITHUB_TOKEN }}
```
codefromthecrypt, line 33: will re-introduce homebrew later

mathetake (approved):

PR #289, opened 2021-07-06
Adds basic website

codefromthecrypt:
This creates a basic website which we can add logos etc to later! I will handle netlify and dns for func-e.io post-merge.

Screen Shot 2021-07-06 at 2 13 01 PM


.bingo/Variables.mk
```diff
@@ -35,6 +35,12 @@ $(GORELEASER): $(BINGO_DIR)/goreleaser.mod
 	@echo "(re)installing $(GOBIN)/goreleaser-v0.166.0"
 	@cd $(BINGO_DIR) && $(GO) build -mod=mod -modfile=goreleaser.mod -o=$(GOBIN)/goreleaser-v0.166.0 "github.com/goreleaser/goreleaser"
 
+HUGO := $(GOBIN)/hugo-v0.85.0
```
mathetake, line 38: nice 😄

internal/cmd/usage_md_test.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/tetratelabs/func-e/internal/globals"
 )
 
-const siteMarkdownFile = "../../site/usage.md"
+const siteMarkdownFile = "../../USAGE.md"
```
mathetake, line 28: is this going to break the existing redirect from getenvoy.io?

mathetake: though it's not a problem at all😄

mathetake (approved):

internal/cmd/usage_md_test.go
codefromthecrypt: yep it will and I'll fix it before anyone notices :D

codefromthecrypt:
thanks for the fast turnaround! we need this in order to setup a stable homebrew. I'll follow-up the loose ends on site details before that now.

internal/cmd/usage_md_test.go
codefromthecrypt: https://github.com/tetratelabs/getenvoy.io/pull/182

PR #290, opened 2021-07-06
Fixes scss and broken links

codefromthecrypt:
A few glitches:

1. we weren't locally using the "extended" hugo, but that was more a parity issue
2. there were a couple 404s
3. maybe we weren't checking out submodules..


codefromthecrypt:
hmm still no dice.. looking into it...

codefromthecrypt:
ok the deploy preview looks ok now. Hopefully it is fixed!

mathetake (approved):

PR #291, opened 2021-07-06
Switches to a simpler theme

codefromthecrypt:
Actual deployment may not work until DNS propagates (maybe tomorrow)

Notably, this hosts the usage docs inline like before!

Screen Shot 2021-07-06 at 7 42 15 PM


codefromthecrypt:
https://deploy-preview-291--func-e.netlify.app/ looks alright so I'm hoping when DNS propagates so will func-e.io 🤞

mathetake (approved):
img

PR #292, opened 2021-07-06
Removes URL pinning

codefromthecrypt:
It is possible this will allow https://func-e.netlify.app/ to work until func-e.io DNS propagates.. It is also possible it won't!

interestingly docs argue whether this is even needed or not.
* netlify->hugo leaves it out https://docs.netlify.com/configure-builds/common-configurations/hugo/
* hugo->netlify puts it in https://gohugo.io/hosting-and-deployment/hosting-on-netlify/

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,15 +1,7 @@
 [build]
   base = "site"
   publish = "public"
+  command = "git submodule update --init && hugo --gc --minify"
 
 [build.environment]
   HUGO_VERSION = "0.85.0"
-
-[context.production]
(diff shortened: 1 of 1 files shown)


mathetake (approved):

codefromthecrypt:
preview went to full domain name, which is worse than before because at least preview worked last time :)

I officially give up until DNS is synced 🏳️ 

Screen Shot 2021-07-06 at 2 13 01 PM

codefromthecrypt:
yay all good https://func-e.io/

mathetake:
Super!

PR #295, opened 2021-07-07
Removes unnecessary redirect

codefromthecrypt:
we serve install.sh directly from the CDN, so there's no need to add complicated flags


mathetake (approved):
better!

PR #296, opened 2021-07-08
Adds missing robots.txt

codefromthecrypt:


mathetake (approved):

codefromthecrypt:
I noticed that while GH actions is skipping on site change (as it should) travis isn't. I'm going to look into why now to make sure we don't aimlessly burn our free arm64 minutes

PR #298, opened 2021-07-08
Optimizes Travis check

codefromthecrypt:
This does two things to conserve arm64 time.
 * run `make check` *after* checking if the build should run!
 * cache the same directories we do in GH actions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,14 +9,21 @@ language: go
 go:
   - 1.16.5
 
+cache:
+  directories: # ~/.func-e/versions is cached so that we only re-download once: for TestFuncEInstall
+    - $HOME/.func-e/versions
+    - $HOME/go/pkg/mod
+    - $HOME/go/bin/*-v*
+
 git:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,14 +9,21 @@ language: go
...
+  directories: # ~/.func-e/versions is cached so that we only re-download once: for TestFuncEInstall
+    - $HOME/.func-e/versions
+    - $HOME/go/pkg/mod
+    - $HOME/go/bin/*-v*
+
 git:
   depth: false  # TRAVIS_COMMIT_RANGE requires full commit history.
 
 before_install: |  # Prevent test build of a documentation or GitHub Actions only change.
-  make check || travis_terminate 1
```
codefromthecrypt, line 16: although this wouldn't have taken a long time with caching setup, it is still better to abort the build vs spend limited minutes

mathetake (approved):

codefromthecrypt:
yay neither GHA nor netlify checked this!

PR #301, opened 2021-07-10
Simplifies version override so it is more like other packages

codefromthecrypt:
This makes the main version more normal. Ex.
```bash
$ go build -ldflags "-s -w -X main.version=foo" .

$ ./func-e -v
func-e version foo
```

This also gets rid of some shell games in static text tests.


mathetake (approved):

codefromthecrypt:
thanks for the look @mathetake !

PR #304, opened 2021-07-11
Refactors SHA-256 code to isolate problem on Windows

codefromthecrypt:
It is currently difficult to tell if tar is related to the SHA-256
mismatch on windows, but not Linux. Better tests make it easier to
figure out and fix.

### DIFF
--- a/internal/tar/tar.go
+++ b/internal/tar/tar.go
@@ -34,6 +34,8 @@ import (
 	"github.com/tetratelabs/func-e/internal/version"
 )
 
+const pathSeparator = string(os.PathSeparator)
+
 type digester struct {
 	r io.Reader
 	h hash.Hash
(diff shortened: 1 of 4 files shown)


internal/tar/tar.go
```diff
@@ -19,15 +19,46 @@ import (
...
+	return
+}
+
+// UntarAndVerify is like Untar, except it errors if the stream has a different signature than the given SHA-256.
+func UntarAndVerify(dst string, src io.Reader, sha256Sum version.SHA256Sum) error { // dst, src order like io.Copy
+	d := digester{src, sha256.New()}
+	if err := Untar(dst, &d); err != nil {
+		return err
+	}
+	sum := version.SHA256Sum(fmt.Sprintf("%x", d.h.Sum(nil)))
```
mathetake, line 55: maybe better use hex.EncodeToString 😄 

```suggestion
	sum := version.SHA256Sum(hex.EncodeToString(d.h.Sum(nil)))
```

mathetake (approved):

internal/tar/tar.go
codefromthecrypt: nice!

codefromthecrypt:

PR #306, opened 2021-07-11
Completes Windows EOL handling

codefromthecrypt:
This replaces `moreos.LineSeparator` with `moreos.Sprintf` and `Fprintf` in order to address all places where newlines need to change.

See https://github.com/golang/go/issues/28822

### DIFF
--- a/internal/cmd/run_test.go
+++ b/internal/cmd/run_test.go
@@ -34,8 +34,6 @@ import (
 	"github.com/tetratelabs/func-e/internal/version"
 )
 
-const ln = moreos.LineSeparator
-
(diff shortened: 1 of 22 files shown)


internal/cmd/run_test.go
```diff
@@ -34,7 +34,7 @@ import (
 	"github.com/tetratelabs/func-e/internal/version"
 )
 
-const ln = moreos.LineSeparator
```
codefromthecrypt, line 37: I think this was a bad name in hindsight and eol is better

codefromthecrypt:
I changed the approach so that it is less code!

mathetake (approved):
LGTM
image

PR #307, opened 2021-07-12
Refactors e2e to not rely on make

codefromthecrypt:
This refactors E2E tests to not rely on make. This gives us more flexibility in windows and also for how we download the release binary.

### DIFF
--- a/e2e/main_test.go
+++ b/e2e/main_test.go
@@ -24,7 +24,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"runtime"
 	"strings"
 	"testing"
 	"time"
@@ -34,25 +33,24 @@ import (
 
 //nolint:golint
 const (
-	funcEBinaryEnvKey      = "E2E_FUNC_E_BINARY"
+	// funcEPathEnvKey holds the path to funcEBin (defaults to pwd)
+	funcEPathEnvKey        = "E2E_FUNC_E_PATH"
 	envoyVersionsURLEnvKey = "ENVOY_VERSIONS_URL"
 	envoyVersionsJSON      = "envoy-versions.json"
 	runTimeout             = 2 * time.Minute
 )
 
 var (
-	funcEPath           = "func-e" // funcEPath holds a path to a 'func-e' binary under test.
+	funcEBin            string // funcEBin holds a path to a 'func-e' binary under test.
 	expectedMockHeaders = map[string]string{"User-Agent": "func-e/dev"}
 )
 
 // TestMain ensures the "func-e" binary is valid.
 func TestMain(m *testing.M) {
 	// As this is an e2e test, we execute all tests with a binary compiled earlier.
-	path, err := readFuncEPath()
-	if err != nil {
+	if err := readFuncEBin(); err != nil {
 		exitOnInvalidBinary(err)
 	}
-	funcEPath = path
 
 	versionLine, _, err := funcEExec("--version")
 	if err != nil {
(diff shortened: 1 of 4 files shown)


e2e/main_test.go
```diff
@@ -107,33 +105,35 @@ func mockEnvoyVersionsServer() (*httptest.Server, error) {
...
-	path := os.Getenv(funcEBinaryEnvKey)
-	if path == "" {
-		// Assemble the default created by "make bin"
-		relativePath := filepath.Join("..", "dist", fmt.Sprintf("func-e_%s_%s", runtime.GOOS, runtime.GOARCH), "func-e")
-		abs, err := filepath.Abs(relativePath)
-		if err != nil {
-			return "", fmt.Errorf("%s didn't resolve to a valid path. Correct environment variable %s", path, funcEBinaryEnvKey)
+func readFuncEBin() error {
+	path := os.Getenv(funcEPathEnvKey)
+	path = filepath.Join("..", path) // relative to project root
```
mathetake, line 112: looks like path defaults to `../${E2E_FUNC_E_PATH}`?

e2e/main_test.go
```diff
@@ -107,33 +105,35 @@ func mockEnvoyVersionsServer() (*httptest.Server, error) {
...
+	if path != "" {
```
mathetake, line 113: therefore it seems that we never reach this branch to me

e2e/main_test.go
```diff
@@ -107,33 +105,35 @@ func mockEnvoyVersionsServer() (*httptest.Server, error) {
...
 	}
+	funcEBin = filepath.Join(abs, "func-e"+moreos.Exe)
+	stat, _ := os.Stat(funcEBin)
+
 	// While "make bin" should result in correct permissions, double-check as some tools lose them, such as
 	// https://github.com/actions/upload-artifact#maintaining-file-permissions-and-case-sensitive-files
-	if !moreos.IsExecutable(stat) {
-		return "", fmt.Errorf("%s is not executable. Correct environment variable %s", path, funcEBinaryEnvKey)
+	if stat == nil || !moreos.IsExecutable(stat) {
+		return fmt.Errorf("%s is not executable. Run `go build .` or `make bin` %s", funcEBin, funcEPathEnvKey)
```
mathetake, line 133: nit: if stat == nil, then `%s is not executable` sounds a bit confusing to me since that case os.Stat itself fails.

e2e/main_test.go
codefromthecrypt: what happens is that pwd is /path/to/project/e2e, so the ../ rebases it at the project root.

ex E2E_FUNC_E_PATH=dist/whatever
/path/to/project/e2e/../dist/whatever = /path/to/project/dist

e2e/main_test.go
codefromthecrypt: oh yeah lemme check that!

e2e/main_test.go
codefromthecrypt: well it is also not executable in that case, was trying to save 3 more lines of copy/pasta

e2e/main_test.go
codefromthecrypt: this is ok. `..` will always exist.

codefromthecrypt: but you are right, the if isn't needed

e2e/main_test.go
mathetake: fair enough!

e2e/main_test.go
mathetake: ah - I confused != with == 😄

e2e/main_test.go
mathetake: so we cannot pass the full path to `E2E_FUNC_E_PATH`?

e2e/main_test.go
codefromthecrypt: fixing it

e2e/main_test.go
mathetake: I mean, absolute path

codefromthecrypt: hmm maybe this would fail if an absolute path were given instead of a relative one. Ex /usr/local/bin. I'll rework it

codefromthecrypt:
ok I reworked as I think what you said all the questions means it should be done differently :D I hope it is better now!

mathetake (approved):
Thank you!

mathetake:
Poor Travis 😄

codefromthecrypt:
travis wasn't cheating by executing `make bin` first :D

PR #309, opened 2021-07-12
Adds Windows support

### LINKED ISSUE #105: Support getenvoy CLI on Windows

codefromthecrypt:
Before this change, we already had windows binaries available and some
functionality working. This makes the last adjustments needed for
Windows to pass both unit and end-to-end tests. While I can currently
use `make` personally, we shouldn't rely on it in Windows because paths
often have a space in them which makes make die. For that reason, the
GitHub Action is setup to use normal go instead.

Fixes #105

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -82,3 +82,37 @@ jobs:
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: bash <(curl -s https://codecov.io/bash)
+
+  test-windows:
+    name: "Run unit tests (windows-latest)"
+    runs-on: windows-latest
+    timeout-minutes: 90  # instead of 360 by default
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v2
+
+      - name: "Install Go"
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: "Cache dependencies"
+        uses: actions/cache@v2
+        with:
+          # This combines unrelated caches because actions/cache@v2 doesn't support multiple
+          # instances, rather a combined path. https://github.com/actions/cache/issues/16
+          path: |  # ~\.func-e\versions is cached so that we only re-download once: for TestFuncEInstall
+            ~\.func-e\versions
+            ~\go\pkg\mod
+          key: test-windows-latest-${{ env.GO_VERSION }}-go-${{ hashFiles('internal/version/last_known_envoy.txt', 'go.sum') }}
+          restore-keys: test-windows-latest-${{ env.GO_VERSION }}-go-
+
+      - name: "Run unit tests"
+        run: go test . ./internal/...
+
+      - name: "Build the `func-e` binary"
+        run: go build --ldflags '-s -w' .
+
+      - name: "Run e2e tests using the `func-e` binary"
+        run: go test -parallel 1 -v -failfast ./e2e
+
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -35,6 +35,7 @@ import (
 // NewRunCmd create a command responsible for starting an Envoy process
 func NewRunCmd(o *globals.GlobalOpts) *cli.Command {
 	var envoyVersion version.Version
+	runDirectoryExpression := moreos.ReplacePathSeparator("$FUNC_E_HOME/runs/$epochtime")
 	cmd := &cli.Command{
 		Name:            "run",
 		Usage:           "Run Envoy with the given [arguments...] until interrupted",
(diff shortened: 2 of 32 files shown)


codefromthecrypt:
@mathetake I have some missing tests, but raising it before doing them, to address any style issues.

internal/moreos/proc_windows.go
```diff
@@ -0,0 +1,55 @@
...
+
+func processGroupAttr() *syscall.SysProcAttr {
+	return &syscall.SysProcAttr{
+		CreationFlags: syscall.CREATE_NEW_PROCESS_GROUP, // Stop Ctrl-Break propagation to allow shutdown-hooks
+	}
+}
+
+func interrupt(p *os.Process) error {
+	pid := p.Pid
+	d, err := syscall.LoadDLL("kernel32.dll")
```
mathetake, line 34: wow cool

.github/workflows/commit.yaml
```diff
@@ -82,3 +82,37 @@ jobs:
...
+
+      - name: "Cache dependencies"
+        uses: actions/cache@v2
+        with:
+          # This combines unrelated caches because actions/cache@v2 doesn't support multiple
+          # instances, rather a combined path. https://github.com/actions/cache/issues/16
+          path: |  # ~\.func-e\versions is cached so that we only re-download once: for TestFuncEInstall
+            ~\.func-e\versions
+            ~\go\pkg\mod
+          key: test-windows-latest-${{ env.GO_VERSION }}-go-${{ hashFiles('internal/version/last_known_envoy.txt', 'go.sum') }}
```
codefromthecrypt, line 107: the actions/cache example for windows uses unix slashes in the hashFiles thing, so did the same thing

internal/cmd/run.go
```diff
@@ -47,12 +48,12 @@ The first version in the below is run, controllable by the "use" command:
 The version to use is downloaded and installed, if necessary.
 
 Envoy interprets the '[arguments...]' and runs in the current working
-directory (aka $CWD) until func-e is interrupted (ex Ctrl+C, Ctrl+Break).
+directory (aka $PWD) until func-e is interrupted (ex Ctrl+C, Ctrl+Break).
```
codefromthecrypt, line 51: intentional, to make portable in powershell

mathetake (approved):
Awesome job!

codefromthecrypt:
Thanks! I'll polish for merge then

codefromthecrypt:
I'll update the install script post-merge

codefromthecrypt:
one flake was strange because envoy returned exit code 1 on ctrl-break. that's not normal and the code of func-e didn't change..

codefromthecrypt:
failed about exactly 2 minutes.. maybe it is time related

```
--- FAIL: TestFuncERun_StaticFilesystem (120.01s)
223
```

codefromthecrypt:
The flakiness is here..

Basically, after func-e receives a terminate signal, propagates it to envoy etc, it hangs until interrupted on timeout.

```
invoking shutdown hooks with deadline [2021-07-12 23:13:00.602]
sending interrupt to envoy (pid=5408)
Warning: 12 23:12:55.656][6828][warning][main] [source/server/server.cc:702] caught ENVOY_SIGTERM
[2021-07-12 23:12:55.656][6828][info][main] [source/server/server.cc:823] shutting down server instance
[2021-07-12 23:12:55.656][6828][info][main] [source/server/server.cc:769] main dispatch loop exited
2021/07/12 23:14:55 done stderr loop for func-e after running [D:\a\func-e\func-e\func-e.exe run -c envoy.yaml]
```

Ex you can see 2 minutes pass from envoy's output of main dispatch loop exited, which means go's test runner killed func-e.

In func-e's code, that we see the interrupt means `handleShutdown` exited.

This means it was blocking on
```
	// Block until the process is complete. This ensures file descriptors are closed.
	<-waitCtx.Done()
```

One explanation is that envoy didn't exit even after logs show it received and also processed the signal. If it did, `waitCancel()` would happen and unblock the above wait.

```
	// Wait in a goroutine. We may need to kill the process if a signal occurs first.
	go func() {
		defer waitCancel()
		_ = r.cmd.Wait() // Envoy logs like "caught SIGINT" or "caught ENVOY_SIGTERM", so we don't repeat logging here.
	}()
```

So, in summary, we can see in the logs that envoy got the signal, and thought it exited `main dispatch loop exited`. However, the process didn't close causing func-e to block. This is flakey and we should handle it otherwise it will curse the project.

When I'm back online, I'll look into maybe killing it on a timeout or some other way, perhaps there's an envoy bug out there also.. I am positive our code to send an interrupt on windows works fine, just process cleanup is not consistent. As you can see on above builds, sometimes it works sometimes it doesn't, and only one test it doesn't (the one who is holding an explicit config file vs anonymous yaml config)

codefromthecrypt:
I'll restart on this now

codefromthecrypt:
this depends on #312 I will kick the build a few times to ensure the flakes are gone now

codefromthecrypt:
ok 3x in a row green

codefromthecrypt:
https://github.com/tetratelabs/func-e/releases/tag/v0.6.0

PR #310, opened 2021-07-19
Builds with Go 1.16.6

codefromthecrypt:
https://github.com/golang/go/issues?q=milestone%3AGo1.16.6+label%3ACherryPickApproved

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -5,7 +5,7 @@
 name: "build"
 
 env:
-  GO_VERSION: "1.16.5"
+  GO_VERSION: "1.16.6"
 
 on:
   # We run tests on non-tagged pushes to master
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
hmm we have something to consider which is that the e2e looks for a coherent version. Actually the linux+windows vs darwin versions are on different schedules until https://github.com/envoyproxy/envoy/issues/16830 (which seems inactive).

We may want to consider using a regex vs a strict version match here in order to avoid locking up things.. WDTY @mathetake

codefromthecrypt:
also it is tricky but "latestVersion" in https://archive.tetratelabs.io/envoy/envoy-versions.json causes this.. we may want to consider not publishing at all until os/x is done as basically it ends up latest, except darwin which introduces some skew edge cases in tests and usage

codefromthecrypt:
opened https://github.com/tetratelabs/archive-envoy/issues/15 on this topic

codefromthecrypt:
tetratelabs/archive-envoy#16 works around the build outage by pinning 1.18.3

mathetake (approved):

PR #312, opened 2021-07-22
Adds hook to ensure Envoy exited

codefromthecrypt:
We currently don't ensure Envoy has exited when it said it should. Due the the nature of FS locks in windows, we can detect this happens sometimes in #309. This change adds the plumbing we can use on all supported operating systems.

### DIFF
--- a/internal/envoy/run.go
+++ b/internal/envoy/run.go
@@ -29,7 +29,7 @@ import (
 )
 
 // Run execs the binary at the path with the args passed. It is a blocking function that can be shutdown via SIGINT.
-func (r *Runtime) Run(ctx context.Context, args []string) (err error) {
+func (r *Runtime) Run(ctx context.Context, args []string) (err error) { //nolint:gocyclo
 	// We can't use CommandContext even if that seems correct here. The reason is that we need to invoke shutdown hooks,
 	// and they expect the process to still be running. For example, this allows admin API hooks.
 	cmd := exec.Command(r.opts.EnvoyPath, args...) // #nosec -> users can run whatever binary they like!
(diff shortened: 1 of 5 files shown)


internal/moreos/proc_darwin.go
```diff
@@ -29,6 +29,13 @@ func interrupt(p *os.Process) error {
 	return p.Signal(syscall.SIGINT)
 }
 
+func ensureProcessDone(p *os.Process) error {
+	if err := p.Kill(); err != nil && err != os.ErrProcessDone {
```
codefromthecrypt, line 33: it is handy we are using recent golang as this is exported

mathetake (approved):

codefromthecrypt:
going to do some more testing as maybe we can't use consolidated logic with windows due to various issues stemming from https://golang.org/issue/25965

codefromthecrypt:
oh cool there was no netlify comment on this

internal/envoy/run.go
```diff
@@ -78,15 +78,21 @@ func (r *Runtime) Run(ctx context.Context, args []string) (err error) {
...
 
-	// Block until the process is complete. This ensures file descriptors are closed.
-	<-waitCtx.Done()
-
+	// At this point, shutdown hooks have run and Envoy is interrupted.
+	// Block until it exits to ensure file descriptors are closed prior to archival.
+	// Allow up to 5 seconds for a clean stop, killing if it can't for any reason.
+	select {
+	case <-waitCtx.Done():
+	case <-time.After(5 * time.Second):
```
codefromthecrypt, line 93: so the key to unblocking windows is to allow graceful shutdown, but nuke it like this. I used 5 seconds because I haven't lost my thumb

mathetake (approved):

Issue #320, opened 2021-07-28
Sign the windows executable (func-e.exe) and installer (func-e.msi)


codefromthecrypt:
**Is your feature request related to a problem? Please describe.**
Right now, func-e.exe is not signed. Windows will suggest to not run it due to it being from an unknown publisher. If it had an authenticode signature, the warning would change to less severe, but still suggest to not run it. After the [reputation improves](https://www.digicert.com/blog/ms-smartscreen-application-reputation), it is possible Windows won't suggest to not run it.

**Describe the solution you'd like**
To do this, we would first need a code-signing certificate, probably for the whole of tetratelabs, not just here. Seems people use digicert or symantec, and some pay more for an EV cert which has a higher initial reputation. https://answers.microsoft.com/en-us/protect/forum/all/windows-defender-is-giving-a-warning-when/ee48147e-8fe2-46d5-bfba-c5d38b332360

Once we have that, we'd need to add it in PKCS#12 format as an org secret  (ex tetratelabs-cert.pfx). This allows any of our repos can use it.

Some process we make or is an existing action or otherwise automation would sign the .exe after goreleaser makes it. The options listed below work on linux, which means our releaser doesn't need to step across different operating systems to complete. These also assume some step uses curl or otherwise to download the pfx into the file `tetratelabs-cert.pfx`

### option 1: osslsigncode
[osslsigncode](https://github.com/mtrojnar/osslsigncode) seems a simpler option as it natively supports PKCS#12 and requires no conversion. It is available in ubuntu (apt) and os/x (homebrew). While it has few contributors, and doesn't release often, it is active and written in C.

Ex.
```
  osslsigncode sign -h sha256 -pkcs12 tetratelabs-cert.pfx -pass "$TETRATELABS_CERT_PASSWORD" \
    -n "func-e makes running Envoy® easy" -i https://func-e.io -t http://timestamp.digicert.com \
    -in dist/func-e_windows_amd64/func-e.exe -out dist/func-e_windows_amd64/func-e-signed.exe &&
mv dist/func-e_windows_amd64/func-e-signed.exe dist/func-e_windows_amd64/func-e.exe
```

### option 2: signcode

signcode is a part of the [mono](https://github.com/mono/mono) project and available also on ubuntu or os/x (homebrew). While it is written in c#, it releases often and has a lot of contributors.

[...]

PR #321, opened 2021-07-29
Adds Windows installer

codefromthecrypt:
This adds a Windows installer (msi) using the subset of WIX features
supported by https://wiki.gnome.org/msitools.

Dodging unsupported features brings DevX because we can create and test
the tool on OS/x.

A future PR will patch the the release created by goreleaser, uploading the MSI


codefromthecrypt:
current status: this works if built on windows or darwin, it installs and uninstalls on windows

ex. `VERSION=0.6.1 make msi`

Makefile
```diff
@@ -32,6 +32,30 @@ bin $(BIN): $(GORELEASER)
...
+WIN_BIN := dist/func-e_windows_amd64
+WIN_BIN_EXE := $(WIN_BIN)/func-e.exe
+
+$(WIN_BIN_EXE): $(GORELEASER)
+	@echo "--- win-bin ---"
+	@GOOS=windows GOARCH=amd64 $(GORELEASER) build --snapshot --single-target --rm-dist
+
+.PHONY: msi
+msi: $(WIN_BIN_EXE)
+ifeq ($(OS),Windows_NT)  # same for Windows 10 etc
```
codefromthecrypt, line 46: @mathetake earlier I was trying to explain that how in windows it is two commands to accomplish what in linux/darwin is one. This is the thing I referred to.

Hover goes a bit farther and does this and some other bits in go... not quite sure want to go there at this point, but we could certainly optimize later. https://github.com/go-flutter-desktop/hover/blob/9f74033acaeabec51906e9542d8706111b98fe31/cmd/packaging/windows-msi.go

Main thing is that make is already doing the release, so probably good enough to leave this as an `ifeq`

Makefile
```diff
@@ -32,6 +32,30 @@ bin $(BIN): $(GORELEASER)
...
+	@candle -v -arch x64 -dVersion=$(VERSION)  \
+	-dBin=$(WIN_BIN)/func-e.exe \
+	packaging/func-e.wxs
+	@light func-e.wixobj -o $(WIN_BIN)/func-e.msi -spdb
+	@rm func-e.wixobj
+else  # TODO: the exit code is 1 for some reason
```
codefromthecrypt, line 52: TODO: make sure this comment isn't lying :) I think the exit code is zero again.

packaging/func-e.wxs
```diff
@@ -0,0 +1,90 @@
...
+        </Feature>
+
+        <!-- Add func-e to the default PATH once installation succeeds.
+             This is an alternative to <Component><Environment> which doesn't work on wixl 0.101 (OS/X) -->
+        <Property Id="setx" Value="setx.exe"/>
+        <CustomAction Id="ChangePath" ExeCommand="PATH &quot;%PATH%;[INSTALLDIR]&quot;" Property="setx"
+                      Execute="deferred"/>
+        <InstallExecuteSequence>
+            <RemoveExistingProducts Before="InstallInitialize"/>
+            <!-- TODO: this doesn't cleanup the path addition on uninstall -->
```
codefromthecrypt, line 86: I will likely not address this TODO as I spent an hour and in worst case we leave func-e in the path post-install. I verified it doesn't redundantly add it on upgrade. This is the same impl as nomad, so if they manage to improve, so could we.

packaging/func-e.wxs
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- This builds a minimal package of func-e. The only required variable is Version
+     Ex. `wixl -D Version=0.6.1`
+
+     While all options used are documented, the initial version of this file was helped by:
+     * cloudflared - https://github.com/cloudflare/cloudflared/commit/38af26e232d6a5fd981d40981b702a1d06a3b8ee
+     * podmad - https://github.com/containers/podman/commit/b6b0b6e8bd0f8c75fe9411f722a43f9305c6137a
```
codefromthecrypt, line 7: I would also cite apple's swift, except it uses windows-only (ex non GNOME) features, which actually resulted in me losing time to figure that out. Luckily I commented below which features would be nice if GNOME msitools ends up supporting them https://github.com/apple/swift-installer-scripts/blob/bbcd6a3addfe9c3b62caebf1683cd39c27e7ddb9/platforms/Windows/Toolchain/windows-toolchain.wxs

Makefile
```diff
@@ -32,6 +32,30 @@ bin $(BIN): $(GORELEASER)
...
+WIN_BIN_EXE := $(WIN_BIN)/func-e.exe
+
+$(WIN_BIN_EXE): $(GORELEASER)
+	@echo "--- win-bin ---"
+	@GOOS=windows GOARCH=amd64 $(GORELEASER) build --snapshot --single-target --rm-dist
+
+.PHONY: msi
+msi: $(WIN_BIN_EXE)
+ifeq ($(OS),Windows_NT)  # same for Windows 10 etc
+	@candle -v -arch x64 -dVersion=$(VERSION)  \
```
codefromthecrypt, line 47: TODO: crash if this is empty. also this can't start with 'v'. It *must* be a N.N.N version to work with microsoft

codefromthecrypt (commented):
notes for monday's cleanup/integration for merge

smoke test of the msi should happen in a GHA when func-e.wxs changes (on windows)
```bash
# install
msiexec /i func-e.msi /qn

# run func-e which should be in the path, this will download envoy
func-e run -version

# uninstall
msiexec /x func-e.msi /qn
```

packaging/func-e.wxs
codefromthecrypt: resolved with a comment about an about to land feature that solves this

codefromthecrypt:
until #320 we can only run unattended quiet install with admin (using `msiexec`) because otherwise it silently fails due to the permissions request

codefromthecrypt:
installer icon looks great thanks to @peterj!

image

codefromthecrypt:
goreleaser can't patch an existing set of files to publish, and doesn't support MSI either per https://github.com/goreleaser/goreleaser/issues/1295

This means we need to run goreleaser and subsequently package MSI. I'll work on that in a bit

codefromthecrypt:
OK I think this will likely pass, and probably all that should be done in one PR

codefromthecrypt:
rebased over recent refactoring and pulled msi tests into their own file. This will reduce scope that requires external tool installation to only changes related to MSI.

codefromthecrypt:
OK I added minimal installer validation. Once this is in I'll refactor the release script to patch what goreleaser makes with the MSI produced later via `make msi`

codefromthecrypt:
alright I think this is polished enough. I will do the release work in another PR

PR #323, opened 2021-08-03
Adds favicons for the website

codefromthecrypt:
This adds favicons and also Web Manifest configuration for
https://func-e.io. This comes with great thanks to @peterj,
who contributed the png and ico files.


Makefile
```diff
@@ -37,7 +37,7 @@ bin $(BIN): $(GORELEASER)
 site: $(HUGO)
 	@echo "--- site ---"
 	@git submodule update
-	@cd site && $(HUGO) server -D
+	@cd site && $(HUGO) server --disableFastRender -D
```
codefromthecrypt, line 40: I added this to prevent weird glitches when you update hugo between commits

codefromthecrypt:
netlify preview URL is https://deploy-preview-323--func-e.netlify.app/

mathetake:
Oh netlify bots stopped putting preview links completely? 😄

mathetake (approved):

codefromthecrypt:
> Oh netlify bots stopped putting preview links completely? 😄
yes I think we offended them :D

mathetake:
btw I don't think we need to run unittests on changes under site/ 😄

codefromthecrypt:
> btw I don't think we need to run unittests on changes under site/ 😄

I think usually we wouldn't but I couldn't resist upgrading hugo, which tripped the root path

mathetake:
ah - ic 😄

Issue #316, opened 2021-07-26
Support func-e windows packages


codefromthecrypt:
from v0.6.0 we can run on windows
- [x] exe added to release as .zip
- [ ] exe Authenticode signed
- [x] windows installer added to release as .msi
- [ ] msi Autheticode signed
- [ ] chocolatey package?
- [x] winget package?

codefromthecrypt:
winget is sponsored by MS and might also work fine

codefromthecrypt:
https://github.com/tetratelabs/func-e/pull/321 starts this as we need MSI before we can do winget

codefromthecrypt:
installers needed to do packaging aren't yet on the default GHA runners, so I raised this https://github.com/actions/virtual-environments/issues/3857

codefromthecrypt:
https://github.com/tetratelabs/func-e/pull/332 completes everything except EXE and MSI signing

With MSI in place, we should be able to start winget support

codefromthecrypt:
updated https://github.com/tetratelabs/func-e/releases/tag/v0.6.0 to use the MSI

codefromthecrypt:
iiuc looks like winget will be available by default in a service release of Windows 10 next month https://github.com/microsoft/winget-cli/discussions/988#discussioncomment-835319

codefromthecrypt:
next steps are below, if my loose understanding of the process is correct
- [ ] figure out the correct [winget-create](https://github.com/microsoft/winget-create) syntax (needs to run on windows maybe)
- [ ] test deploy to [winget-pkgs-submission-test](https://github.com/microsoft/winget-pkgs-submission-test)
- [ ] if signing isn't required, raise an initial PR to [winget-pkgs](https://github.com/microsoft/winget-pkgs), or wait for #320 if it is required
- [ ] setup a GHA workflow to update the above similar to how we do in homebrew

I'll start on this next week

codefromthecrypt:
As far as I can tell, winget needs to run on an actual windows OS (even though choco can run on linux via [docker/mono](https://hub.docker.com/r/linuturk/mono-choco/)

codefromthecrypt:
https://github.com/microsoft/winget-pkgs/pull/24284

codefromthecrypt:
winget is now available

```cmd
C:\Users\tmp>winget.exe install func-e
Found func-e [Tetrate.func-e]
This application is licensed to you by its owner.
[...]

PR #325, opened 2021-08-03
Uses make in windows

codefromthecrypt:
This consolidates the build by using the same matrix for Linux, MacOS
and Windows. To do the latter requires installation of CygWin, and to
get GHA to use that requires overriding the `shell`. One other quirk is
that when you are running tasks via cygwin, it resets the CWD. To avoid
that, this adds the GITHUB_WORKSPACE to `.bashrc`. Finally, this dodges
various newline related problems and comments why.

Unrelated changes including updating syntax that allows us to
break-apart actions/cache and fixing some lint errors that the build
never caught because lint wasn't run on Windows.

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -35,31 +35,65 @@ jobs:
     name: "Run unit tests (${{ matrix.os }})"
     runs-on: ${{ matrix.os }}
     timeout-minutes: 90  # instead of 360 by default
+
     strategy:
       fail-fast: false  # don't fail fast as sometimes failures are operating system specific
       matrix:
-        os: [ ubuntu-latest, macos-latest ]
+        os: [ubuntu-latest, macos-latest]
+        include:
+          - os: windows-latest
+            # Until GHA allows a cygwin runner, we have to override the default bash
+            shell: C:\tools\cygwin\bin\bash.exe --login -o errexit -o nounset -o pipefail -o igncr -i {0}
+
+    defaults:
+      run:
+        shell: ${{ matrix.shell || 'bash' }}
+
     steps:
+      # This sets up the enviornment we need for Windows
+      #  * install make and cygwin's bash which resolves paths correctly for .bingo/Variables.mk
+      #  * configure git so line-feeds don't trip lint https://github.com/actions/checkout/issues/135
+      #
+      # Note: We don't need to install anything here (ex wixtoolset):
+      # https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md
+      - name: "Pre-configure (Windows)"
+        if: runner.os == 'Windows'
+        run: |
+          choco install -y --no-progress --source cygwin make bash
+          git config --global core.autocrlf input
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
main goal of this is not just coherency, but allowing us to use this to test using make as a part of #316

codefromthecrypt:
looks like it didn't like caching cygwin..

codefromthecrypt:
last didn't error, but it didn't cache either. I'm trying explicit dir for chocolatey

codefromthecrypt:
I'll pull the choco package caching out of scope, as it isn't working, nor is the dominant time in the workflow anyway.

Makefile
```diff
@@ -54,7 +54,7 @@ test:
 .PHONY: e2e
 e2e: $(BIN)
 	@echo "--- e2e ---"
-	@E2E_FUNC_E_PATH=$(PWD)/$(BIN) go test -parallel 1 -v -failfast ./e2e
```
codefromthecrypt, line 57: changing this allowed one less translation problem in cygwin. It is now allowed to be and defaults to relative to the project root.

internal/moreos/proc_windows.go
```diff
@@ -69,7 +69,7 @@ func ensureProcessDone(p *os.Process) error {
 		}
 		return os.NewSyscallError("OpenProcess", e)
 	}
-	defer syscall.CloseHandle(h)
+	defer syscall.CloseHandle(h) //nolint
```
mathetake, line 72: oh which kind of lint errors we got with this?

.github/workflows/commit.yaml
```diff
@@ -35,32 +35,66 @@ jobs:
...
-        os: [ ubuntu-latest, macos-latest ]
+        os: [ubuntu-latest, macos-latest]
+        include:
+          - os: windows-latest
+            # Until GHA allows a cygwin runner, we have to override the default bash
+            shell: C:\tools\cygwin\bin\bash.exe --login -o errexit -o nounset -o pipefail -o igncr -i {0}
+
+    defaults:
+      run:
+        shell: ${{ matrix.shell || 'bash' }}
```
mathetake, line 50: I didn't know that we can use`||` stuff in the GHA expression 😄

mathetake (approved):

internal/moreos/proc_windows.go
codefromthecrypt: good point :D It is the usual defer thing where we don't have a nested function to check the thing that closes the thing didn't also err. Pretty ok with ack'ing, but agree it should be precise!

```suggestion
	defer syscall.CloseHandle(h) //nolint:errcheck
```

codefromthecrypt:
Thanks for the look!

codefromthecrypt:
with some adjustments I was able to use the built-in make and save a lot of time and crazy config https://github.com/tetratelabs/func-e/pull/327

codefromthecrypt:
https://github.com/tetratelabs/func-e/pull/328 simplified further

PR #326, opened 2021-08-04
Uses make from GHA runner on Windows

codefromthecrypt:
while opaque, make is actually in there!


codefromthecrypt:
merging as I think no one would want things to be more complicated than necessary

Issue #314, opened 2021-07-22
replace bingo with go 1.17 `go run` commands


codefromthecrypt:
**Is your feature request related to a problem? Please describe.**
Right now, we use a modified version of bingo to pin versions without tainting GOBIN's default. For example, this is used to invoke known versions of hugo, goreleaser, golint, etc

**Describe the solution you'd like**
Inside the Makefile, declare paths for each main tool and do `go run` instead. This reduces the toolchain knowledge to run the project and conversion to make-unfriendly env like windows more straightforward.

> go run now accepts arguments with version suffixes (for example, go run example.com/cmd@v1.0.0). This causes go run to build and run packages in module-aware mode, ignoring the go.mod file in the current directory or any parent directory, if there is one. This is useful for running executables without installing them or without changing dependencies of the current module. 

**Describe alternatives you've considered**
Leaving it alone. There is some benefit to leaving it as depending on how `go run` in 1.17 is implemented, it might be inefficient to repeatedly run the same command.

**Additional context**
https://tip.golang.org/doc/go1.17#go-command

codefromthecrypt:
cc @nacx

mathetake:
this is good to know this, thanks for raising here 😄

codefromthecrypt:
It is possible that when we use this, we can avoid needing cygwin bash anymore. So, possibly worth revisiting the github action that currently installs cygwin-bash and overrides the shell in order to translate paths so bingo's Variables.mk doesn't die (#325)

nacx:
When I was trying to upgrade the monorepo to 1.17 and move away from Bingo, I found this error when trying to use Kustomize:

```
go generate ./...
+ set -o allexport
+ source ../../../../scripts/build/tools.mk
++ GOIMPORTS=golang.org/x/tools/cmd/goimports@v0.1.5
++ GOLANGCI_LINT=github.com/golangci/golangci-lint/cmd/golangci-lint@v1.41.1
++ KUSTOMIZE=sigs.k8s.io/kustomize/kustomize/v3@v3.10.0
++ LICENSER=github.com/liamawhite/licenser@v0.6.1-0.20210729145742-be6c77bf6a1f
+ set +o allexport
+ go run sigs.k8s.io/kustomize/kustomize/v3@v3.10.0 build sources/managementplane -o managementplaneoperator.yaml
[...]

PR #327, opened 2021-08-05
Restricts default make config to single-entry GOPATH

codefromthecrypt:
This restricts the default make config to single-entry GOPATH, which allows windows to work without hacks or cygwin.

Those who must use multiple-entry GOPATH will have to override GOBIN if they use our Makefile

See https://github.com/bwplotka/bingo/issues/26
This defers porting to go 1.17 #314


codefromthecrypt:
This takes the windows runner performance down from 7m to 4m!

codefromthecrypt:
merging as I don't expect any pushback, but don't mind revisions later either!

PR #330, opened 2021-08-06
Refactors release end-to-end tests to include Windows

codefromthecrypt:
Before, we didn't test Windows. This adds tests and also makes updating
the homebrew formula depend on e2e succeeding.

Moreover, this removes a shell script in favor of `gh` commands which
are included in the default images.

See https://github.com/actions/virtual-environments/blob/main/images


codefromthecrypt:
ps I tested the commands used on a fake draft release, however the real test will be the next real one. We may have some glitches but I will fix any as they occur, as always.

.github/workflows/release.yaml
```diff
@@ -32,15 +32,25 @@ jobs:
...
 
   e2e:
+    needs: func-e
     name: Run e2e tests (${{ matrix.os }})
-    needs:
-      - func-e
     runs-on: ${{ matrix.os }}
     timeout-minutes: 90  # instead of 360 by default
     strategy:
-      fail-fast: false  # don't fail fast as sometimes failures are operating system specific
```
codefromthecrypt, line 41: intentionally not allowing fail-fast on releases anymore

.github/workflows/release.yaml
```diff
@@ -51,13 +61,20 @@ jobs:
...
+          gh release download "${GITHUB_REF#refs/tags/}" -p '${{ matrix.pattern }}'
+          ${{ matrix.unzip || tar -xzf *.tar.gz && rm *.tar.gz }}
         env:  # authenticate as the release is a draft
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: "Run e2e tests using released `func-e` binary"
         run: go test -parallel 1 -v -failfast ./e2e
 
+  homebrew:
+    needs: e2e
```
codefromthecrypt, line 74: now homebrew PR won't happen if e2e fails

internal/moreos/moreos.go
```diff
@@ -26,6 +26,7 @@ import (
 
 const (
 	// Exe is the runtime.GOOS-specific suffix for executables. Ex. "" unless windows which is ".exe"
+	// See https://github.com/golang/go/issues/47567 for formalization
```
codefromthecrypt, line 29: unrelated but was missing rationale

codefromthecrypt (commented):
notes

PR #333, opened 2021-08-09
Uses separate workflow for version bumps

codefromthecrypt:
This uses a separate workflow for version bumps, starting with homebrew (winget is likely next).

This allows us to centralize flows that are not file-driven, rather release driven. Particularly, actions that result in external pull requests would end up here.

This converts the homebrew bump to explicit commands and also removes the need to say "same perms as `brew bump-formula-pr func-e --version 0.6.0`" as we literally are using that command.

This also centralizes the token and identity of those pull requests so they are attributed to the org, not the person who created the token.

### DIFF
--- a/.github/workflows/bump.yaml
+++ b/.github/workflows/bump.yaml
@@ -1,0 +1,43 @@
+# `name` value will appear "as is" in the badge.
+# See https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository
+# yamllint --format github .github/workflows/bump.yaml
+---
+name: bump
+
+# This workflow is for external packages, and is only triggered once ("released").
+# This centralizes workflows that are hard to undo, such as reversing PRs that update packages.
+#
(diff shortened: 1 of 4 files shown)


.github/workflows/bump.yaml
```diff
@@ -0,0 +1,43 @@
+# `name` value will appear "as is" in the badge.
+# See https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository
+# yamllint --format github .github/workflows/bump.yaml
+---
+name: bump
+
+# This workflow is for external packages, and is only triggered once ("released").
+# This centralizes workflows that are hard to undo, such as reversing PRs that update packages.
+#
+# One trade-off of deferring to "released" is that release notes will publish packages not yet
```
codefromthecrypt, line 10:
```suggestion
# One trade-off of deferring to "released" is that release notes will refer to packages not yet
```

mathetake (approved):

PR #334, opened 2021-08-09
Adds which command

codefromthecrypt:
This adds a `which` command that prints the path to the Envoy binary
that would be run.

Ex.
```bash
$ func-e which
/Users/adrian/.func-e/versions/1.19.0/bin/envoy
```

This allows func-e to install, but not run envoy for scenarios such as building other images.

Ex. the following are equivalent:
```bash
$ $(func-e which) --version
$ func-e run --version
```

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -22,6 +22,7 @@ This value must be constant within a `$FUNC_E_HOME`.
 | run | Run Envoy with the given [arguments...] until interrupted |
 | versions | List Envoy versions |
 | use | Sets the current [version] used by the "run" command |
+| which | Prints the path to the Envoy binary used by the "run" command |
 | --version, -v | Print the version of func-e |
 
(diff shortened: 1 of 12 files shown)


mathetake (approved):
How's the `func-e which` output look like? 😄

codefromthecrypt:
cc @noseka1 this might help you with https://github.com/noseka1/openshift-toolbox/

codefromthecrypt:
@mathetake sorry I should have put the output. this idea is to be like the `which` command, so print the path only. Ex.

```bash
$ func-e which
/Users/adrian/.func-e/versions/1.19.0/bin/envoy
```

codefromthecrypt:
@lanceplarsen similarly maybe this will help in your scripts in https://github.com/hashicorp/field-workshops-consul

Issue #338, opened 2021-08-11
Envoy 1.19 / Centos 7 / GLIBC_2_18 compatibility issue


WeirdBob:
**Describe the bug**
When trying to use func-e on centos 7, envoy 1.19 fails to start with the message: _versions/1.19.0/bin/envoy: /lib64/libc.so.6:  version `GLIBC_2.18' not found_

**To Reproduce**
1. Install func-e on centos-7
2. Try to run envoy (--version is enough)
3. See error

**Expected behavior**
Envoy is compiled with a compatible glibc version (as it was the case with previous getenvoy builds)

**Environment (please complete the relevant information):**
 - OS: Centos 7
 - Envoy Version: 1.19.0

**Additional context**
Thanks for you work even if it doesn't work (yet) in my case !

codefromthecrypt:
hi @WeirdBob getenvoy and all things associated with that are gone now. specific to this topic, we aren't using a different build pipeline: the envoy binary here was extracted from the official [docker image](https://hub.docker.com/r/envoyproxy/envoy) during [archive-envoy](https://github.com/tetratelabs/archive-envoy/blob/master/bin/car_envoy.sh#L94)

That said, this is an issue.. we don't have a CI dimension for cents. We can take a look and try to help figure out what can be done about it. Thanks for reporting!

codefromthecrypt:
@mathetake maybe you can help elaborate options on this one as you are more familiar with the parts inside. Here are some bread crumbs, though some may be irrelevant.

This isn't an attempt to find a reason to not support centos, rather show what might be in front of us given the envoy project's decisions. Custom builds are extremely expensive so, upstream adjustments are preferred vs going back into the labor pit of perpetually broken custom builds.

 * the [centos build image]https://github.com/envoyproxy/envoy-build-tools/blob/main/build_container/Dockerfile-centos) is marked as experimental, but appears to be latest centos 7
 *  the official build wants glibc >= 2.27 since [1.15](https://github.com/envoyproxy/envoy/blob/4a0dfd5df93d6aac5c04e869c3cb3b3587767b41/docs/root/version_history/v1.15.0.rst)
 * the build has an if-statement around [centos 7](
https://github.com/envoyproxy/envoy/blob/4b7389a7522959490045f7b47f58268f71450509/bazel/external/wee8.genrule_cmd#L115-L117), so this might imply we can raise support issues...

codefromthecrypt:
[...]

PR #339, opened 2021-08-11
Runs e2e on centos 8

codefromthecrypt:
centos isn't a supported runner in GHA or travis, but you can run it via container with GHA.
We need to use an internal image to avoid rate limiting and 12minutes to install prereqs

See #338

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -97,3 +97,48 @@ jobs:
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: bash <(curl -s https://codecov.io/bash)
+
+  # Github Action Runners don't support all operating systems we may need to test. Below ensures
+  # we run on the intended minimum platforms. Note: we also use Travis for arm64+ubuntu.
+  #
+  # Use only images from internal-images.yaml to avoid Docker rate limits and slow install steps.
+  test-e2e:
+    name: "Run e2e tests via containers (${{ matrix.os }})"
+    runs-on: ubuntu-latest
+    # "When you only specify a container image, you can omit the image keyword."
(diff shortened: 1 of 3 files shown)


.github/workflows/internal-images.yml
```diff
@@ -0,0 +1,52 @@
...
+#
+# To troubleshoot, run locally with docker and install go directly. Ex:
+# ```bash
+# $ docker run -ti -v $PWD:/func-e --rm centos:8
+# [root@d37da4601545 /]# cd /func-e
+# [root@d37da4601545 func-e]# eval "$(curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | GIMME_GO_VERSION=1.16.6 bash)"
+# now execute your setup commands!
+# ```
+jobs:
+  copy-images:
```
mathetake, line 22: copy? 😄

.github/workflows/commit.yaml
```diff
@@ -97,3 +97,48 @@ jobs:
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: bash <(curl -s https://codecov.io/bash)
+
+  # Github Action Runners don't support all operating systems we may need to test. Below ensures
+  # we run on the intended minimum platforms. Note: we also use Travis for arm64+ubuntu.
+  #
+  # Use only images from internal-images.yaml to avoid Docker rate limits and slow install steps.
+  test-e2e:
+    name: "Run e2e tests via containers (${{ matrix.os }})"
```
mathetake, line 106:
```suggestion
    name: "Run e2e tests via containers (${{ matrix.matrix }})"
```

.github/workflows/internal-images.yml
codefromthecrypt: hah yeah..
```suggestion
  build-and-push-images:
```

mathetake (approved):

PR #340, opened 2021-08-13
Signs windows exe and installer (msi)

### LINKED ISSUE #320: Sign the windows executable (func-e.exe) and installer (func-e.msi)

codefromthecrypt:
This signs the Windows exe and installer (msi), using a self-signed
certificate in test and a real one from Digicert in prod.

Fixes #320


codefromthecrypt:
I've tested locally and windows to make sure that the MSI ends up signed. However, the real test will be cutting a draft release after merge of v0.6.1, then deleting it after verification. This will ensure the org secrets etc are expected.

codefromthecrypt:
Thanks for the look.. I'll 🤞 and do a draft release now.

codefromthecrypt:
build had an unrelated glitch during e2e, but the signatures on both files uploaded are correct!

PR #341, opened 2021-08-13
sleep to prevent flakes  on delayed windows installer service

codefromthecrypt:
hopefully, this will prevent a flake like this https://github.com/tetratelabs/func-e/runs/3318938205

I used 2 seconds because that's the last magic number choco uses https://github.com/chocolatey/choco/blob/master/src/chocolatey/infrastructure.app/services/AutomaticUninstallerService.cs#L41


mathetake (approved):

codefromthecrypt:
fingers crossed no flakes https://github.com/tetratelabs/func-e/actions/runs/1126548259

PR #344, opened 2021-08-13
Fixes glitches in package update script

codefromthecrypt:
This reverts to the old homebrew script because handing homebrew updates
in github env is a little tricky. This also adds a missing dep.

See https://github.com/tetratelabs/func-e/runs/3319457023?check_suite_focus=true


mathetake (approved):

codefromthecrypt:
last glitch a9ccc4dcc852bc50d8ecab0dcd44566e954a984d

PR #347, opened 2021-08-20
Removes test utilities for built-ins

codefromthecrypt:
Notably, this uses the go 1.17 https://pkg.go.dev/testing#T.Setenv


mathetake (approved):

codefromthecrypt:
windows fail is probably legit as the lifecycle of the directory used in tests is different now, and windows locks files when a process is open. I will look into it

codefromthecrypt:
ok this is no longer flaking, which is great.

mathetake (approved):

PR #348, opened 2021-08-22
Cleans up tests

codefromthecrypt:
Some tests passed for the wrong reason because of drift in CLI parsing
inside the fake envoy server. This fixes that and also moves the fake
admin endpoint into the fake envoy server. This allows us to verify
admin address path parsing works as we expect it to.

### DIFF
--- a/internal/envoy/install_test.go
+++ b/internal/envoy/install_test.go
@@ -31,15 +31,13 @@ import (
 	"github.com/tetratelabs/func-e/internal/globals"
 	"github.com/tetratelabs/func-e/internal/moreos"
 	"github.com/tetratelabs/func-e/internal/test"
-	"github.com/tetratelabs/func-e/internal/test/morerequire"
 	"github.com/tetratelabs/func-e/internal/version"
 )
 
 func TestUntarEnvoyError(t *testing.T) {
 	ctx := context.Background()
-	tempDir, removeTempDir := morerequire.RequireNewTempDir(t)
+	tempDir := t.TempDir()
 	dst := filepath.Join(tempDir, "dst")
-	defer removeTempDir()
 
 	tarball, tarballSHA256sum := test.RequireFakeEnvoyTarGz(t, version.LastKnownEnvoy)
 
@@ -88,8 +86,7 @@ func TestUntarEnvoyError(t *testing.T) {
 
 // TestUntarEnvoy doesn't test compression formats because that logic is in tar.Tar
 func TestUntarEnvoy(t *testing.T) {
-	tempDir, removeTempDir := morerequire.RequireNewTempDir(t)
-	defer removeTempDir()
+	tempDir := t.TempDir()
 
 	tarball, tarballSHA256sum := test.RequireFakeEnvoyTarGz(t, version.LastKnownEnvoy)
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -210,8 +207,7 @@ func TestInstallIfNeeded_AlreadyExists(t *testing.T) {
 }
 
 func TestVerifyEnvoy(t *testing.T) {
-	tempDir, removeTempDir := morerequire.RequireNewTempDir(t)
-	defer removeTempDir()
+	tempDir := t.TempDir()
 
 	envoyPath := filepath.Join(tempDir, "versions", string(version.LastKnownEnvoy))
 	require.NoError(t, os.MkdirAll(filepath.Join(envoyPath, "bin"), 0755))
@@ -248,30 +244,20 @@ type installTest struct {
 }
 
 func setupInstallTest(t *testing.T) (*installTest, func()) {
-	var tearDown []func()
-
-	tempDir, removeTempDir := morerequire.RequireNewTempDir(t)
-	tearDown = append(tearDown, removeTempDir)
-
 	versionsServer := test.RequireEnvoyVersionsTestServer(t, version.LastKnownEnvoy)
-	tearDown = append(tearDown, versionsServer.Close)
-
+	homeDir := t.TempDir()
 	return &installTest{
-			ctx:        context.Background(),
-			tempDir:    tempDir,
-			tarballURL: test.TarballURL(versionsServer.URL, runtime.GOOS, runtime.GOARCH, version.LastKnownEnvoy),
-			GlobalOpts: globals.GlobalOpts{
-				HomeDir:          tempDir,
-				EnvoyVersionsURL: versionsServer.URL + "/envoy-versions.json",
-				Out:              new(bytes.Buffer),
-				Platform:         globals.DefaultPlatform,
-				RunOpts: globals.RunOpts{
-					EnvoyPath: filepath.Join(tempDir, "versions", string(version.LastKnownEnvoy), binEnvoy),
-				},
+		ctx:        context.Background(),
+		tempDir:    t.TempDir(),
+		tarballURL: test.TarballURL(versionsServer.URL, runtime.GOOS, runtime.GOARCH, version.LastKnownEnvoy),
+		GlobalOpts: globals.GlobalOpts{
+			HomeDir:          homeDir,
+			EnvoyVersionsURL: versionsServer.URL + "/envoy-versions.json",
+			Out:              new(bytes.Buffer),
+			Platform:         globals.DefaultPlatform,
+			RunOpts: globals.RunOpts{
+				EnvoyPath: filepath.Join(homeDir, "versions", string(version.LastKnownEnvoy), binEnvoy),
 			},
-		}, func() {
-			for i := len(tearDown) - 1; i >= 0; i-- {
-				tearDown[i]()
-			}
-		}
+		},
+	}, versionsServer.Close
 }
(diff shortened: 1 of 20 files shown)


internal/envoy/run_test.go
```diff
@@ -57,7 +57,7 @@ func TestRuntime_Run(t *testing.T) {
 			name: "func-e Ctrl+C",
 			args: []string{"-c", "envoy.yaml"},
 			// Don't warn the user when they exited the process
-			expectedStdout:   moreos.Sprintf("starting: %s -c envoy.yaml %s\nGET /ready HTTP/1.1\n", fakeEnvoy, adminFlag),
```
codefromthecrypt, line 60: we don't expect func-e to make HTTP requests, but if it accidentally did, this would show up!

internal/envoy/shutdown/admin_test.go
```diff
@@ -74,5 +62,5 @@ func runWithShutdownHook(t *testing.T, runDir string, hook func(r *envoy.Runtime
 		if fakeInterrupt != nil {
 			fakeInterrupt()
 		}
-	}, r, stderr, args...)
+	}, r, stderr, "-c", "envoy.yaml")
```
codefromthecrypt, line 65: fake envoy should crash if config isn't passed. this was crashing but we never checked the error

internal/test/testdata/fake_envoy.go
```diff
@@ -26,24 +29,43 @@ func main() {
 	// Echo the same first line Envoy would. This also lets code scraping output know signals are trapped
 	os.Stderr.Write([]byte("initializing epoch 0" + lf)) //nolint
 
-	// Validate a haveConfig is passed just like Envoy
+	// Validate a config is passed just like Envoy
 	haveConfig := false
-	for _, arg := range os.Args[1:] {
-		switch arg {
-		case "-c", "--haveConfig-path", "--haveConfig-yaml":
```
codefromthecrypt, line 33: here was the find/replace error that sat too long :p

codefromthecrypt (commented):
notes

codefromthecrypt:
cool got a flake similar to windows on go 1.17.. I will sort this out

```
--- FAIL: TestFuncERun (1.90s)
    run_test.go:63: 
        	Error Trace:	run_test.go:63
        	Error:      	Not equal: 
        	            	expected: "initializing epoch 0\nstarting main dispatch loop\n"
        	            	actual  : "initializing epoch 0\n"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,2 @@
        	            	 initializing epoch 0
        	            	-starting main dispatch loop
        	            	 
        	Test:       	TestFuncERun
invoking shutdown hooks with deadline [2021-08-22 09:42:49.301]
failed shutdown hook: unable to create file to write ps output to: open /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/2738671499/envoy_home/runs/1629625363974914000/node/ps.txt: no such file or directory
failed shutdown hook: unable to capture Envoy configuration and metrics: unable to read /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/2738671499/envoy_home/runs/1629625363974914000/admin-address.txt: open /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/2738671499/envoy_home/runs/1629625363974914000/admin-address.txt: no such file or directory
failed shutdown hook: open /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/2738671499/envoy_home/runs/1629625363974914000/node/network_interface.json: no such file or directory
failed shutdown hook: open /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/2738671499/envoy_home/runs/1629625363974914000/node/connections.json: no such file or directory
sending interrupt to envoy (pid=6547)
```

internal/envoy/runtime.go
```diff
@@ -57,6 +58,31 @@ type Runtime struct {
 	shutdownHooks []func(context.Context) error
 }
 
+// String is only used in tests. It is slow, but helps when debugging CI failures
```
codefromthecrypt, line 61: @mathetake let me know what you think about this.. this is only for tests (like TestFuncERun) which are in a different package or invoked via urfave which makes it impossible to print out the state here on fail. It is some extra code, but it is tested.

internal/test/envoy.go
```diff
@@ -42,19 +42,26 @@ type Runner interface {
...
 	reader := bufio.NewReader(stderr)
-	require.Eventually(t, func() bool {
-		b, err := reader.Peek(512)
-		return err != nil && strings.Contains(string(b), "initializing epoch 0")
-	}, 2*time.Second, 100*time.Millisecond, "never started process")
+	waitFor := "initializing epoch 0"
+	if !assert.Eventually(t, func() bool {
+		b, e := reader.Peek(512)
+		return e != nil && strings.Contains(string(b), waitFor)
+	}, 5*time.Second, 100*time.Millisecond) {
```
codefromthecrypt, line 61: I switched this after finding out some were legit timeouts and the process didn't finish, yet. Using an "if" allows us to format state before the error raises

internal/envoy/install_test.go
```diff
@@ -248,30 +244,20 @@ type installTest struct {
...
-			GlobalOpts: globals.GlobalOpts{
-				HomeDir:          tempDir,
-				EnvoyVersionsURL: versionsServer.URL + "/envoy-versions.json",
-				Out:              new(bytes.Buffer),
-				Platform:         globals.DefaultPlatform,
-				RunOpts: globals.RunOpts{
-					EnvoyPath: filepath.Join(tempDir, "versions", string(version.LastKnownEnvoy), binEnvoy),
-				},
+		ctx:        context.Background(),
+		tempDir:    t.TempDir(),
```
mathetake, line 251: Oh is this change intentional? - before we pass `tempDir` both for here and HomeDir below, but not we use different dir.

internal/envoy/runtime.go
```diff
@@ -57,6 +58,31 @@ type Runtime struct {
...
+	}
+
+	exitStatus := -1
+	if r.cmd != nil && r.cmd.ProcessState != nil {
+		if ws, ok := r.cmd.ProcessState.Sys().(syscall.WaitStatus); ok {
+			exitStatus = ws.ExitStatus()
+		}
+	}
+
+	return fmt.Sprintf("{stdout: %s, stderr: %s, exitStatus: %d}", stdout, stderr, exitStatus)
```
mathetake, line 83: how does actual output look like from this String? Is that human-readable? 😄

internal/envoy/install_test.go
codefromthecrypt: yeah it feels like an accident. I think before when we couldn't get on-demand directories without a cleanup function, probably led pressure to re-use directories. I undid that here to avoid that since it is easy now.

internal/envoy/runtime.go
codefromthecrypt: you tell me :) I think it is ok https://github.com/tetratelabs/func-e/pull/348/files#diff-0f025fafb42983e9b6848a9ba595f8f38e9cfaa80cd1a5b4c6bb8de7c82d1835R110

mathetake: no I mean what if stderr and stdout become too long by actualy Envoy? 😄

mathetake: oh now I see you comments ^^ this is only for tests!

codefromthecrypt: yeah I agree if real console out, this would be a bad idea

internal/envoy/runtime.go
mathetake: In order to achieve this we will have to define interface over Runtime vs the codebase complexity. I'm fine for now with this, we can come back to this in another PR if we want to move out this test only function😄

codefromthecrypt: yeah let's do a refactoring PR later if you are ok with it for an hour :D because there's some package complexity due to some things being in cmd_test vs cmd etc

internal/envoy/version_test.go
```diff
@@ -48,7 +47,9 @@ func TestGetHomeVersion_Empty(t *testing.T) {
 		{"missing home version file", func() {
 			require.NoError(t, os.Remove(homeVersionFile))
 		}},
-		{"missing home dir", removeHomeDir},
+		{"missing home dir", func() {
+			require.NoError(t, os.Remove(homeDir))
```
mathetake, line 51: nit do we need this? I believe homeDir is cleaned up internally by testing.T.

codefromthecrypt: good question. in this case it isn't about cleanup, rather that this is testing behavior when it isn't there. Since most tests need the directory, it is eagerly provisioned and only cleaned up for this case

mathetake (approved):

codefromthecrypt:
thanks for the help takeshi-san I will rebase the other in-flight PR then do the package refactoring as I agree there's too much weight in the main code solely for CI flakes that occur now with go 1.17

internal/envoy/runtime.go
codefromthecrypt: https://github.com/tetratelabs/func-e/pull/349

PR #350, opened 2021-08-24
Removes bingo for `go run`

### LINKED ISSUE #314: replace bingo with go 1.17 `go run` commands

codefromthecrypt:
This uses versioned syntax of `go run` brought to us in 1.17
https://golang.org/doc/go1.17

In doing so, we can remove the complexity of bingo and also simplify its
impact on windows.

Fixes #314


.github/workflows/msi.yaml
```diff
@@ -72,9 +72,8 @@ jobs:
         with:
           path: |  # TODO: go build cache if we care, noting it is OS-specific
```
codefromthecrypt, line 73: I will sort out the build cache in a different PR as it requires handling Os-specifics!

mathetake (approved):
nice :smile:

codefromthecrypt:
might have a glitch on goreleaser on windows. regardless will be green prior to merge!

```
± |no-bingo {1} ✓| → make clean msi
--- clean ---
--- win-bin ---
GOOS=windows GOARCH=amd64 go run github.com/goreleaser/goreleaser@v0.175.0 build --snapshot --single-target --rm-dist
fork/exec /var/folders/vd/1cf8zdb1721f4z5rjggy8bp40000gn/T/go-build2241750511/b001/exe/goreleaser.exe: exec format error
make: *** [dist/func-e_windows_amd64/func-e.exe] Error 1
```

codefromthecrypt:
note: this is slower than before because go rebuilds each time. However, this time spent is less important to me than lack of cruft, so think it is a better tradeoff. @mathetake wdyt?

Makefile
```diff
@@ -123,24 +128,20 @@ coverage:
...
-format: $(GOIMPORTS) ## Format all Go code
+format: ## Format all Go code
 	@echo "--- format ---"
-	@$(LICENSER) apply -r "Tetrate"
+	@go run $(licenser) apply -r "Tetrate"
 	@find . -type f -name '*.go' | xargs gofmt -s -w
-	@for f in `find . -name '*.go'`; do \
-	    awk '/^import \($$/,/^\)$$/{if($$0=="")next}{print}' $$f > /tmp/fmt; \
-	    mv /tmp/fmt $$f; \
-	    $(GOIMPORTS) -w -local github.com/tetratelabs/func-e $$f; \
```
codefromthecrypt, line 142: doing this in a loop is as time consuming as trying to parse wtf ^^ was trying to do

Makefile
```diff
@@ -43,9 +46,11 @@ MSI_VERSION ?= 0.0.1
 # Right now, the only arch we support is amd64 because Envoy doesn't yet support arm64 on Windows
 # https://github.com/envoyproxy/envoy/issues/17572
 # Once that occurs, we will need to set -arch arm64 and bundle accordingly.
-$(WIN_BIN_EXE): $(GORELEASER)
+$(WIN_BIN_EXE):
 	@echo "--- win-bin ---"
-	@GOOS=windows GOARCH=amd64 $(GORELEASER) build --snapshot --single-target --rm-dist
+    # We can't pass GOOS=windows GOARCH=amd64 as it ends up interpreted by go, not goreleaser.
```
codefromthecrypt, line 51: this is annoying to build all when you only need windows, but I think we will be killing goreleaser soon anyway

Makefile
```diff
@@ -43,9 +46,11 @@ MSI_VERSION ?= 0.0.1
...
+    # Hence, we build all targets even if we only need Windows.
+	@go run $(goreleaser) build --snapshot --rm-dist
```
mathetake, line 53: I like being better aligned 😄 

```suggestion
	# We can't pass GOOS=windows GOARCH=amd64 as it ends up interpreted by go, not goreleaser.
	# Hence, we build all targets even if we only need Windows.
	@go run $(goreleaser) build --snapshot --rm-dist
```

codefromthecrypt:
thanks takeshi-san!

Issue #352, opened 2021-08-24
Remove goreleaser for a make-centric release process


codefromthecrypt:
Right now, we use three tiers of tools to do a release (besides go!):
* make - orchestrate goreleaser and things it leaves out (like package signing)
* goreleaser - build the matrix of goos_gooarch
* github actions (GHA) - kick off release and patch it later

As goreleaser has limited scope, we have two strange things:
 * we have to run make inside as a hook because of issues like https://github.com/goreleaser/goreleaser/issues/459
 * we have to separately build and upload the msi it is unsupported https://github.com/goreleaser/goreleaser/issues/1295

The result of the combination of concerns is that critical steps are spread between the three toolchains: make, gorelease and GHA. This means copy/paste of comments, inability to really simulate a release in all that it needs before doing it, which implies bumpy releases. It is also more complicated and requires higher skillset as you have to know how all these things work.

I propose we unify to Make and untangle all of this. The reason is that we have to know make anyway, regardless of what we do. It is better to use a single make process than work too hard trying to get goreleaser to be used when it no longer serves us.

One con is that make can end up looking like a cat ran over the keyboard. We should be careful to use its power, but also cite the why as there are a mixed audience of maintainers here. We should also be conscious of platform dependencies due to our build being required to execute on linux, darwin and also .... windows!

Related work:

When looking around, I found some inspiration from @hairyhenderson in [gomplate](https://github.com/hairyhenderson/gomplate) I think this project is the most "on track" I've seen with regards to a healthy setup. While our solution will end up looking different than Dave's, he deserves a shout-out especially as I must have looked at 200 project makefiles before finding what I think is the clear winner! https://github.com/hairyhenderson/gomplate/blob/master/Makefile

codefromthecrypt:
https://github.com/tetratelabs/func-e/pull/361 finishes this

PR #353, opened 2021-08-25
Scaffold a new Makefile

codefromthecrypt:
Doing this as a separate file helps #352 not become massive.

Ex.

Screen Shot 2021-08-25 at 5 35 25 PM


mathetake (approved):
Adrian Cole - Principal Makefile Engineer 😄

PR #358, opened 2021-08-28
Generates release assets without goreleaser on 'make dist'

### LINKED ISSUE #352: Remove goreleaser for a make-centric release process

codefromthecrypt:
This removes the dependency on goreleaser by implementing what it does
in make.

This will fix #352 once the release process uses it, which will be a
different PR.

Screen Shot 2021-08-29 at 2 19 17 PM

```bash
± |make-dist {3} ✓| → file dist/*
dist/func-e_dev_checksums.txt:       ASCII text
dist/func-e_dev_darwin_amd64.tar.gz: gzip compressed data, last modified: Sun Aug 29 04:19:00 2021, from Unix, original size modulo 2^32 7503360
dist/func-e_dev_darwin_arm64.tar.gz: gzip compressed data, last modified: Sun Aug 29 04:19:01 2021, from Unix, original size modulo 2^32 7743488
dist/func-e_dev_linux_amd64.tar.gz:  gzip compressed data, last modified: Sun Aug 29 04:19:01 2021, from Unix, original size modulo 2^32 7005696
dist/func-e_dev_linux_arm64.tar.gz:  gzip compressed data, last modified: Sun Aug 29 04:19:02 2021, from Unix, original size modulo 2^32 6948352
dist/func-e_dev_windows_amd64.msi:   Composite Document File V2 Document, Little Endian, Os: Windows, Version 5.0, MSI Installer, Code page: 1252, Title: Installation Database, Subject: func-e installer, Author: Tetrate, Keywords: Installer, Comments: This installer database contains the logic and data required to install func-e., Template: x64;1033, Revision Number: {D3FEA555-CCBD-4DCF-9FE3-55E05B7C63ED}, Create Time/Date: Sun Aug 29 04:19:04 2021, Last Saved Time/Date: Sun Aug 29 04:19:04 2021, Number of Pages: 310, Number of Words: 2, Name of Creating Application: msitools 0.101, Security: 2
dist/func-e_dev_windows_amd64.zip:   Zip archive data, at least v2.0 to extract
```

### DIFF
--- a/Next.mk
+++ b/Next.mk
@@ -109,7 +109,7 @@ $(checksums): $(archives) $(packages)
 
 # dist generates the assets that attach to a release
 # Ex. https://github.com/tetratelabs/func-e/releases/tag/v$(VERSION)
-dist: $(archives) $(archives) $(packages) $(checksums) ## generates release assets
+dist: $(archives) $(packages) $(checksums) ## generates release assets
 
 # this makes a marker file ending in .signed to avoid repeatedly calling codesign
 %.signed: %
--- a/Next.mk
+++ b/Next.mk
@@ -1,18 +1,20 @@
 # Copyright 2021 Tetrate
 # Licensed under the Apache License, Version 2.0 (the "License")
-.PHONY: test build e2e
+#
+# This script uses automatic variables (ex $<) and substitution references $(<:.signed=)
+# Please see GNU make's documentation if unfamiliar: https://www.gnu.org/software/make/manual/html_node/
+.PHONY: test build e2e dist
 
 # This should be driven by automation and result in N.N.N, not vN.N.N
-VERSION ?= dev
+VERSION   ?= dev
+build_dir := build
+dist_dir  := dist
 
-# Temporarily match convention of where goreleaser put binaries
-build_dir         := dist
 # Build the path relating to the current runtime (goos,goarch)
-current_binary    := $(build_dir)/func-e_$(shell go env GOOS)_$(shell go env GOARCH)/func-e$(shell go env GOEXE)
-# Right now, the only arch we support is amd64 because Envoy doesn't yet support arm64 on Windows
-# https://github.com/envoyproxy/envoy/issues/17572
-# Once that occurs, we will need to set -arch arm64 and bundle accordingly.
-windows_binary    := $(build_dir)/func-e_windows_amd64/func-e.exe
+goos   := $(shell go env GOOS)
+goarch := $(shell go env GOARCH)
+goexe  := $(shell go env GOEXE)
+current_binary := $(build_dir)/func-e_$(goos)_$(goarch)/func-e$(goexe)
 
 # ANSI escape codes. f_ means foreground, b_ background.
 # See https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_(Select_Graphic_Rendition)_parameters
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
ps the `msi` action uses the work here, but it will be used in the `release` workflow also, later once we refactor out the old makefile.

codefromthecrypt:
ok so looks like the windows image maybe doesn't have zip.. or maybe it needs the paths different.. I will try later on my windows host.

```
process_begin: CreateProcess(NULL, zip -qj dist/func-e_dev_windows_amd64.zip build/func-e_windows_amd64/func-e.exe, ...) failed.
make (e=2): The system cannot find the file specified.
```

codefromthecrypt:
yeah looks like zip isn't installed.. maybe we ask people to install it. otherwise, it gets complicated https://github.com/bamler-lab/constriction/blob/831ac417526d14de34977aa2665ce752d3a5a5fc/.github/workflows/release.yaml#L199-L210

codefromthecrypt:
7zip might do the trick easier https://github.com/ducaale/xh/blob/8e90649a535e82c07b0ec380f7e93d019236c890/.github/workflows/release.yaml#L98-L104

I will work on this later

codefromthecrypt:
ok hopefully I got it. I also removed the ✔️ from the output as it wasn't formatting on windows

🤞

Next.mk
```diff
@@ -37,32 +39,97 @@ build: $(current_binary) ## Build the func-e binary
...
+	$(call go-build, $@, $^)
 
+$(build_dir)/func-e_%/func-e.exe: main.go
+	$(call go-build, $@, $^)
+
+$(dist_dir)/func-e_$(VERSION)_%.zip: $(build_dir)/func-e_%/func-e.exe.signed
+	@printf "$(ansi_format_dark)" zip "zipping $@"
+	@mkdir -p $(dir $@)
+ifeq ($(OS),Windows_NT)  # Windows 10 etc use 7zip:
+	@# single-quote wildcard to strip directory: https://sevenzip.osdn.jp/chm/cmdline/commands/add.htm
```
codefromthecrypt, line 64: 7zip is super bizzare

Next.mk
```diff
@@ -37,32 +39,97 @@ build: $(current_binary) ## Build the func-e binary
...
+platforms := darwin_amd64 darwin_arm64 linux_amd64 linux_arm64 windows_amd64
 
-$(windows_binary): override goos=windows
-$(windows_binary): override goarch=amd64
-$(windows_binary):
-	$(call go-build, $@)
+$(build_dir)/func-e_%/func-e: main.go
+	$(call go-build, $@, $^)
 
+$(build_dir)/func-e_%/func-e.exe: main.go
```
codefromthecrypt, line 57: in case you didn't know the nuance of make, you can only use a '%' once, that's why we have to do this twice

Next.mk
```diff
@@ -37,32 +39,97 @@ build: $(current_binary) ## Build the func-e binary
...
+	$(call go-build, $@, $^)
+
+$(dist_dir)/func-e_$(VERSION)_%.zip: $(build_dir)/func-e_%/func-e.exe.signed
+	@printf "$(ansi_format_dark)" zip "zipping $@"
+	@mkdir -p $(dir $@)
+ifeq ($(OS),Windows_NT)  # Windows 10 etc use 7zip:
+	@# single-quote wildcard to strip directory: https://sevenzip.osdn.jp/chm/cmdline/commands/add.htm
+	@7z -bso0 -bsp0 a $@ './$(dir $<)/*.exe'
```
codefromthecrypt, line 65: `-bso0 -bsp0` means quiet (in a very loud way!)

codefromthecrypt:
ok all tidied up.. if windows fails on github it is different than local and will report back before saying "review please"

codefromthecrypt:
heh next is shasum.. somehow it must be there on my machine but not on the github actions windows worker

```
/usr/bin/sh: shasum: command not found
mingw32-make: *** [Next.mk:104: dist/func-e_dev_checksums.txt] Error 127
rm build/func-e_linux_arm64/func-e build/func-e_darwin_arm64/func-e build/func-e_windows_amd64/func-e.exe.signed build/func-e_linux_amd64/func-e build/func-e_darwin_amd64/func-e
Error: Process completed with exit code 2.
```

codefromthecrypt:
ok all OS-specific things are sorted I think

Next.mk
```diff
@@ -37,32 +39,110 @@ build: $(current_binary) ## Build the func-e binary
...
+	$(call codesign, $@)
+	@printf "$(ansi_format_bright)" msi "ok"
+
+# Archives are tar.gz, except in the case of Windows, which uses zip.
+archives  := $(non_windows_platforms:%=$(dist_dir)/func-e_$(VERSION)_%.tar.gz) $(windows_platforms:%=$(dist_dir)/func-e_$(VERSION)_%.zip)
+packages  := $(windows_platforms:%=$(dist_dir)/func-e_$(VERSION)_%.msi)
+checksums := $(dist_dir)/func-e_$(VERSION)_checksums.txt
+
+# Darwin doesn't have sha256sum. See https://github.com/actions/virtual-environments/issues/90
+sha256sum := $(if $(findstring darwin,$(goos)),shasum -a 256,sha256sum)
```
codefromthecrypt, line 104: this really should be `(if (eq`, as it will be a keyword equal.  However, this works, and I'm not sure the function if any for inline equals https://www.gnu.org/software/make/manual/html_node/Text-Functions.html

Next.mk
```diff
@@ -37,32 +39,110 @@ build: $(current_binary) ## Build the func-e binary
...
-	@CGO_ENABLED=0 GOOS=$(goos) GOARCH=$(goarch) go build \
+	@CGO_ENABLED=0 GOOS=$(call go-os,$1) GOARCH=$(call go-arch,$1) go build \
 		-ldflags "-s -w -X main.version=$(VERSION)" \
-		-o $1 main.go
-	@printf "$(ansi_format_bright)" build "✔"
+		-o $1 $2
+	@printf "$(ansi_format_bright)" build "ok"
+endef
+
+# This requires osslsigncode package (apt or brew) or latest windows release from mtrojnar/osslsigncode
```
dio, line 130: Sorry, a question: do we generally assume these tools (including the one from Wixtools above) are installed, and it is not necessary to "assert" their existence?

codefromthecrypt: generally it will fail out, and one thing is if we did do an assert, we would want to scope it here because otherwise we end up needing osslsigncode in things that don't require it (ex easy to make e2e accidentally depend on things). So, the thinking is that breaking because file-not-found is as good as breaking explicitly on file-not-found and less work. does that make sense?

codefromthecrypt: in any case, if you know a portable search-path operation that will work on windows/darwin/linux and not hairy, we can add it.. just not sure it won't cause more distraction than clarity. in general I wish more things were written in go because we can build them on-demand :D

dio: Got it! Thanks for the explanation!

dio (approved):

mathetake (approved):

codefromthecrypt:
thanks for the help folks!

PR #361, opened 2021-08-30
Switches to goreleaser-free Makefile

### LINKED ISSUE #352: Remove goreleaser for a make-centric release process

codefromthecrypt:
This changes to the restructured Makefile. This completes the `dist`
target, which prepares release assets for upload.

```bash
$ make
func-e
help       Describe how to use each target
build      Build the func-e binary
test       Run all unit tests
coverage   Generate test coverage
e2e        Run all end-to-end tests
dist       Generate release assets
clean      Ensure a clean build
check      Verify contents of last commit
site       Serve website content
```

Fixes #352


.github/workflows/release.yaml
```diff
@@ -42,26 +42,23 @@ jobs:
...
+      # raise issues to ask about it.
+      - name: "Create draft release"
+        run: |
           tag="${GITHUB_REF#refs/tags/}"
-          version=${tag#v}
-          MSI_VERSION=${version} make msi
-          msi=dist/func-e_${version}_windows_amd64.msi
-          mv dist/func-e_windows_amd64/func-e.msi ${msi}
-          gh release upload "${tag}" "${msi}"
+          ./.github/release_notes.sh ${tag} > release-notes.txt
```
codefromthecrypt, line 56: this is where we don't use goreleaser anymore. we prep the release and create it in one step. Before, we had to update the release afterwards to do what goreleaser couldn't do (MSI)

mathetake (approved):

PR #363, opened 2021-08-30
Allow to "use" minor versions

### LINKED ISSUE #360: func-e use latest patch
## Current State
Right now, we do a lazy lookup of the most recent version of envoy, as determined by https://archive.tetratelabs.io/envoy/envoy-versions.json latestVersion Whatever the latest version is at the time is serialized to ~/.func-e/version to prevent excessive downloading and ensure the config is stable.

## Impact
This is great because the first time use of Envoy doesn't require knowing any version, much less the patch version. However, they also won't automatically get a new patch version, and usually patches are desirable. Moreover, there's no way to get the latest patch of an older minor version. For example, if 1.20 is the last minor, there's no way for a user to automatically get the last patch of 1.18.

## Current Workaround

Currently, if a user wants to get the latest patch, they can delete ~/.func-e/version and have it re-fetch. To get the latest patch of an older minor (ex 1.18), they need to parse `func-e versions -a` and `func-e use` that value (ex. 1.18.4). While this can be worked around manually, it is even more a pain in devops scenarios like build matrixes.

## Design idea

It is a better default to get the latest patch of the latest minor. For example, at the cost of re-hitting envoy-versions.json each time, we can take a value such as '1.19' and translate it into the concrete latest version, even after '1.20' comes out. Basically, if we see an expression missing the patch part, do a pre-emptive fetch of the version list and parse the last patch from that. After that point, the existing logic kicks in to see if it is already downloaded or not.

In the first implementation, it isn’t necessary for us to cache the envoy-versions.json to disk even if we do in process, but we should handle the offline case and fall-back to an existing patch when the versions list is unavailable. There should be a test for this edge case.

dio:
This changeset introduces a way to switch the Envoy version using compatible
Envoy version (for example 1.18) and automatically infer the available
patched version to be downloaded. For instance, if the currently
downloaded version is 1.18.3, and the remote one has 1.18.4, calling
"use" downloads and sets the current version to 1.18.4.

Fixes #360 

Signed-off-by: Dhi Aurrahman <dio@tetrate.io>

### DIFF
--- a/e2e/func-e_use_test.go
+++ b/e2e/func-e_use_test.go
@@ -65,3 +65,79 @@ func TestFuncEUse_UnknownVersion(t *testing.T) {
 	require.Equal(t, moreos.Sprintf(`error: couldn't find version "%s" for platform "%s/%s"
 `, v, runtime.GOOS, runtime.GOARCH), stderr)
 }
+
+func TestFuncEUse_MinorVersion(t *testing.T) {
+	homeDir := t.TempDir()
+
+	t.Run("install last known", func(t *testing.T) {
+		stdout, stderr, err := funcEExec("--home-dir", homeDir, "use", string(version.LastKnownEnvoy))
+
+		require.NoError(t, err)
+		require.Regexp(t, `^downloading https:.*tar.*z\r?\n$`, stdout)
+		require.Empty(t, stderr)
+
+		// The binary was installed.
+		envoyBin := filepath.Join(homeDir, "versions", string(version.LastKnownEnvoy), "bin", "envoy"+moreos.Exe)
+		require.FileExists(t, envoyBin)
+
+		// The current version was written.
+		f, err := os.ReadFile(filepath.Join(homeDir, "version"))
+		require.NoError(t, err)
+		require.Equal(t, version.LastKnownEnvoy, version.Version(f))
+	})
+
+	baseVersion := version.Version("1.12.0")
+	t.Run("install base version", func(t *testing.T) {
+		stdout, stderr, err := funcEExec("--home-dir", homeDir, "use", string(baseVersion))
+
+		require.NoError(t, err)
+		require.Regexp(t, `^downloading https:.*tar.*z\r?\n$`, stdout)
+		require.Empty(t, stderr)
+
+		// The binary was installed.
+		envoyBin := filepath.Join(homeDir, "versions", string(baseVersion), "bin", "envoy"+moreos.Exe)
+		require.FileExists(t, envoyBin)
+
+		// The base version was written.
+		f, err := os.ReadFile(filepath.Join(homeDir, "version"))
+		require.NoError(t, err)
+		require.Equal(t, baseVersion, version.Version(f))
+	})
+
+	// The intended minor version to be installed.
+	minorVersion := version.Version("1.12")
+
+	// This version is EOL'd based on: https://github.com/envoyproxy/envoy/blob/main/RELEASES.md#release-schedule.
+	upgradedVersion := version.Version("1.12.7")
+	t.Run("install upgraded version", func(t *testing.T) {
+		stdout, stderr, err := funcEExec("--home-dir", homeDir, "use", string(minorVersion))
+
+		require.NoError(t, err)
+		require.Regexp(t, `^downloading https:.*tar.*z\r?\n$`, stdout)
+		require.Empty(t, stderr)
+
+		// The binary was installed.
+		envoyBin := filepath.Join(homeDir, "versions", string(upgradedVersion), "bin", "envoy"+moreos.Exe)
+		require.FileExists(t, envoyBin)
+
+		// The upgraded version was written.
+		f, err := os.ReadFile(filepath.Join(homeDir, "version"))
+		require.NoError(t, err)
+		require.Equal(t, minorVersion, version.Version(f))
+	})
+
+	t.Run("use upgraded version after downloaded", func(t *testing.T) {
+		stdout, stderr, err := funcEExec("--home-dir", homeDir, "use", string(minorVersion))
+		require.NoError(t, err)
+		require.Equal(t, moreos.Sprintf("%s is already downloaded\n", upgradedVersion), stdout)
+		require.Empty(t, stderr)
+	})
+
+	t.Run("which upgraded version", func(t *testing.T) {
+		stdout, stderr, err := funcEExec("--home-dir", homeDir, "which")
+		relativeEnvoyBin := filepath.Join("versions", string(upgradedVersion), "bin", "envoy"+moreos.Exe)
+		require.Contains(t, stdout, moreos.Sprintf("%s\n", relativeEnvoyBin))
+		require.Empty(t, stderr)
+		require.NoError(t, err)
+	})
+}
--- a/e2e/func-e_use_test.go
+++ b/e2e/func-e_use_test.go
@@ -161,7 +161,7 @@ func getVersionsRange(stdout, minor string) (max, min string) {
 			rows = append(rows, row[:strings.Index(row, " ")])
 		}
 	}
-	min = rows[0]
-	max = rows[1]
+	min = rows[len(rows)-2]
+	max = rows[len(rows)-1]
 	return
 }
--- a/e2e/func-e_use_test.go
+++ b/e2e/func-e_use_test.go
@@ -82,8 +82,6 @@ func TestFuncEUse_MinorVersion(t *testing.T) {
 	// The upgraded version.
 	upgradedVersion := version.Version(upgraded)
 
-	fmt.Println()
-
 	homeDir := t.TempDir()
 
 	t.Run("install last known", func(t *testing.T) {
--- a/e2e/func-e_use_test.go
+++ b/e2e/func-e_use_test.go
@@ -69,6 +69,16 @@ func TestFuncEUse_UnknownVersion(t *testing.T) {
 `, v, runtime.GOOS, runtime.GOARCH), stderr)
 }
 
+func TestFuncEUse_UnknownMinorVersion(t *testing.T) {
+	v := "1.1"
+	stdout, stderr, err := funcEExec("use", v)
+
+	require.EqualError(t, err, "exit status 1")
+	require.Empty(t, stdout)
+	require.Equal(t, moreos.Sprintf(`error: couldn't find the latest patch for "%s" for platform "%s/%s"
+`, v, runtime.GOOS, runtime.GOARCH), stderr)
+}
+
 func TestFuncEUse_MinorVersion(t *testing.T) {
 	// The intended minor version to be installed. This version is known to have darwin, linux, and windows binaries.
 	minorVersion := version.Version("1.18")
@@ -151,8 +161,8 @@ func TestFuncEUse_MinorVersion(t *testing.T) {
 	})
 }
 
-// getVersionsRange returns the latest patch of a minor version and the one before it.
-func getVersionsRange(stdout, minor string) (max, min string) {
+// getVersionsRange returns the first and latest patch of a minor version.
+func getVersionsRange(stdout, minor string) (first, latest string) {
 	s := bufio.NewScanner(strings.NewReader(stdout))
 	rows := []string{}
 	for s.Scan() {
@@ -161,7 +171,8 @@ func getVersionsRange(stdout, minor string) (max, min string) {
 			rows = append(rows, row[:strings.Index(row, " ")])
 		}
 	}
-	min = rows[len(rows)-2]
-	max = rows[len(rows)-1]
+	// The rows is sorted in descending order.
+	first = rows[len(rows)-1]
+	latest = rows[0]
 	return
 }
--- a/e2e/func-e_use_test.go
+++ b/e2e/func-e_use_test.go
@@ -15,9 +15,12 @@
 package e2e
 
 import (
+	"bufio"
+	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -67,12 +70,19 @@ func TestFuncEUse_UnknownVersion(t *testing.T) {
 }
 
 func TestFuncEUse_MinorVersion(t *testing.T) {
-	// The initial version.
-	baseVersion := version.Version("1.18.3")
-	// The intended minor version to be installed.
+	// The intended minor version to be installed. This version is known to have darwin, linux, and windows binaries.
 	minorVersion := version.Version("1.18")
+
+	allVersions, _, err := funcEExec("versions", "-a")
+	require.NoError(t, err)
+
+	base, upgraded := getVersionsRange(allVersions, string(minorVersion))
+	// The initial version.
+	baseVersion := version.Version(base)
 	// The upgraded version.
-	upgradedVersion := version.Version("1.18.4")
+	upgradedVersion := version.Version(upgraded)
+
+	fmt.Println()
 
 	homeDir := t.TempDir()
 
@@ -93,7 +103,7 @@ func TestFuncEUse_MinorVersion(t *testing.T) {
 		require.Equal(t, version.LastKnownEnvoy, version.Version(f))
 	})
 
-	t.Run("install base version", func(t *testing.T) {
+	t.Run(fmt.Sprintf("install %s as base version", base), func(t *testing.T) {
 		stdout, stderr, err := funcEExec("--home-dir", homeDir, "use", string(baseVersion))
 
 		require.NoError(t, err)
@@ -110,7 +120,7 @@ func TestFuncEUse_MinorVersion(t *testing.T) {
 		require.Equal(t, baseVersion, version.Version(f))
 	})
 
-	t.Run("install upgraded version", func(t *testing.T) {
+	t.Run(fmt.Sprintf("install %s as upgraded version", upgraded), func(t *testing.T) {
 		stdout, stderr, err := funcEExec("--home-dir", homeDir, "use", string(minorVersion))
 
 		require.NoError(t, err)
@@ -141,4 +151,19 @@ func TestFuncEUse_MinorVersion(t *testing.T) {
 		require.Empty(t, stderr)
 		require.NoError(t, err)
 	})
+}
+
+// getVersionsRange returns the latest patch of a minor version and the one before it.
+func getVersionsRange(stdout, minor string) (max, min string) {
+	s := bufio.NewScanner(strings.NewReader(stdout))
+	rows := []string{}
+	for s.Scan() {
+		row := strings.TrimSpace(s.Text())
+		if strings.HasPrefix(row, minor+".") {
+			rows = append(rows, row[:strings.Index(row, " ")])
+		}
+	}
+	min = rows[0]
+	max = rows[1]
+	return
 }
--- a/e2e/func-e_use_test.go
+++ b/e2e/func-e_use_test.go
@@ -67,6 +67,13 @@ func TestFuncEUse_UnknownVersion(t *testing.T) {
 }
 
 func TestFuncEUse_MinorVersion(t *testing.T) {
+	// The initial version.
+	baseVersion := version.Version("1.18.3")
+	// The intended minor version to be installed.
+	minorVersion := version.Version("1.18")
+	// The upgraded version.
+	upgradedVersion := version.Version("1.18.4")
+
 	homeDir := t.TempDir()
 
 	t.Run("install last known", func(t *testing.T) {
@@ -86,7 +93,6 @@ func TestFuncEUse_MinorVersion(t *testing.T) {
 		require.Equal(t, version.LastKnownEnvoy, version.Version(f))
 	})
 
-	baseVersion := version.Version("1.12.0")
 	t.Run("install base version", func(t *testing.T) {
 		stdout, stderr, err := funcEExec("--home-dir", homeDir, "use", string(baseVersion))
 
@@ -104,11 +110,6 @@ func TestFuncEUse_MinorVersion(t *testing.T) {
 		require.Equal(t, baseVersion, version.Version(f))
 	})
 
-	// The intended minor version to be installed.
-	minorVersion := version.Version("1.12")
-
-	// This version is EOL'd based on: https://github.com/envoyproxy/envoy/blob/main/RELEASES.md#release-schedule.
-	upgradedVersion := version.Version("1.12.7")
 	t.Run("install upgraded version", func(t *testing.T) {
 		stdout, stderr, err := funcEExec("--home-dir", homeDir, "use", string(minorVersion))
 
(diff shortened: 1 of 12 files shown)


dio:
Still missing the test for the edge case, when `FuncEVersions` failed.

dio:
Kapture 2021-08-30 at 21 44 57

internal/version/last_known_compatible_envoy.txt
```diff
@@ -0,0 +1 @@
+1.19
```
codefromthecrypt, line 1: let's derive this from last_known so there's less maintenance. Remember, the last_known is only used in docs. It should never be thought of as the actual latestVersion. So, what I mean is if this is for purpose of docs, stripping the minor from last_known_envoy should do the trick!

dio: Yes, sure thing!

dio:
Oops. I have wrongly sent out the wrong GIF haha. Will update soon.

internal/envoy/install.go
```diff
@@ -75,9 +79,68 @@ func InstallIfNeeded(ctx context.Context, o *globals.GlobalOpts, v version.Versi
 		// TODO: figure out how to get a stat error that isn't file not exist so we can test this
 		return "", err
 	}
+
+	// Set the inferred Envoy version here, hence it can be used in the subsequent steps.
+	o.EnvoyVersion = v
```
codefromthecrypt, line 84: mention which subsequent steps need this? I hope we don't have any and you can delete this. The main reason for the version is just to get the install path, which this method returns. If you delete this, what breaks?

internal/envoy/install.go
```diff
@@ -75,9 +79,68 @@ func InstallIfNeeded(ctx context.Context, o *globals.GlobalOpts, v version.Versi
...
+	var err error
+
+	// When the given version has no patch component, we fetch all available versions and check if
+	// there is a compatible versions available.
+	if o.EnvoyVersionHasNoPatch {
+		ev, err = FuncEVersions(ctx, o.EnvoyVersionsURL, o.Platform, v)
+		if err != nil {
+			return "", ev, err
+		}
+		compatibleVersions := ev.CompatibleVersionStrings()
```
codefromthecrypt, line 100: both the logic here and CompatibleVersionStrings and its corresponding tests feel like too much code for the job. Also it is looser than it needs to be. There's only one call site as far as I can tell (here), so if you define logic here until there are multiple call sites. That way the function can be un exported and focus only on getting the thing.. iterative seems simplest. I think this can reduce the lines and things to explain, test, review during future refactorings as FuncEVersions can be left alone.

```
// findLatestVersion returns the latest available patch or empty if not found
findLatestVersion(ev version.ReleaseVersions, minor string) string {
var latestPatch
for _, patch ... {
  if patch has prefix minor && patch > latestPatch {
latestPath = patch
}
}
return latestPatch
```

codefromthecrypt (commented):
Thanks for the start on this! and you are fast! Appreciate the preview and I understand this was marked draft so not in final form. I went ahead and made some notes anyway :D

So, main thing to focus is on least impact to achieve the task. I made some suggestions to tighten up the change. Each function defined has to pull its weight once merged.

codefromthecrypt:
ps cool screen recording.. what do you use?

dio:
> ps cool screen recording.. what do you use?

@codefromthecrypt I use https://getkap.co/. I know cmd+shift+5 on macOS gives you screen recording and [kap](https://getkap.co/) is an electron-based app. I'm too lazy to convert the recorded .mov to GIF and put it inside the clipboard: [kap](https://getkap.co/) has this functionality with a single click.

image

e2e/func-e_use_test.go
```diff
@@ -28,32 +29,41 @@ import (
 
 // TestFuncEUse needs to always execute, so we run it in a separate home directory
 func TestFuncEUse(t *testing.T) {
```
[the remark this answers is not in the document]
dio: Added.

internal/envoy/install.go
```diff
@@ -78,6 +82,87 @@ func InstallIfNeeded(ctx context.Context, o *globals.GlobalOpts, v version.Versi
...
 
+func ensureEnvoyVersionHasPatch(ctx context.Context, o *globals.GlobalOpts, v version.Version) (version.Version, version.ReleaseVersions, error) {
+	var ev version.ReleaseVersions
+	var err error
+
+	if getPatchFromVersion(v) >= 0 {
+		return v, ev, nil
+	}
+
+	if ev, err = FuncEVersions(ctx, o.EnvoyVersionsURL, o.Platform, v); err != nil {
```
dio, line 93: @codefromthecrypt do we have a facility to inject FuncEVersions in a test?

dio:
re: Failing downloading versions after install. Seems like there is no way to overcome this? 

So the scenario of fallback to the "current version" is _only_ when the expected binary version is available at least one? For example: 

```
$ cat ~/.func-e/version
1.18.3
$ ls ~/.func-e/versions/1.18.3/
bin/envoy
$ # network failure happens after this line
$ func-e use "1.18"
1.18.3 is already downloaded
```

But what if:

```
$ cat ~/.func-e/version
1.18.3
$ ls ~/.func-e/versions/1.18.3/
bin/envoy
$ # network failure happens after this line
$ func-e use 1.19
```

Are we going to "use" 1.18.3 or throw an error?

codefromthecrypt:
on using a different minor when network isn't available, that should be a legit error

main thing is we are trying to allow offline when the minor is the same

internal/envoy/install.go
codefromthecrypt: To answer your question directly, the cmd package already has a mock server (in `setupTest()`), but we shouldn't need that in this package. This file should be left alone. InstallIfNeeded has no idea of the current version file or anything. Its only responsibility is installing a concrete version which it has been passed.

If new logic should be in this package, it would be version.go, and only add a new (single method not broken up into many as that makes it hard to understand). Iterating through a list of versions to get the patch is only several lines.


I think the key insight here will be that a compositional approach is simpler both to review and implement. IOTW, there's logic in places like `run.go` which would have a new compositional function call to resolve the version into a patch.

```go
		if err := setHomeEnvoyVersion(c.Context, o); err != nil {
			return err
		}
		v, _, err := envoy.CurrentVersion(o.HomeDir)
		if err != nil {
			return NewValidationError(err.Error())
		}
+		if minor pattern matches v {
+			ev, err = envoy.FuncEVersions(ctx, o.EnvoyVersionsURL, o.Platform, o.Version)
+			if err != nil {
+				return NewValidationError(err.Error())
+			}
+			latest = findLatestVersion(ev, v)
+			if latest == "" {
+				return NewValidationError("no version available")
+			}
+			o.EnvoyVersion = latest
+                 } else {
+			o.EnvoyVersion = v
+                 }
-		o.EnvoyVersion = v
```
Bear in mind the above is a sketch and this whole piece may be better off as a higher order function.

If it ends up needing to do be extracted into a higher order function, I personally would slightly adjust FuncEVersions so that it is an interface. That way you don't need to deal with re-testing it in cmd at least the http related stuff.

Ex FuncEVersions is then an interface with 2 functions, the existing as `Get() (version.ReleaseVersions, error)` and a new one to support this as `FindLatestVersion(v) (string, error)`

The struct implementing it can optionally cache between these calls but we should do that separately to keep review burden down. Restructuring to what I mentioned I would expect only 20 lines or so change excluding indents.

The diff changes slightly as you then have one less visible if statement at the call site
```go
		if err := setHomeEnvoyVersion(c.Context, o); err != nil {
			return err
		}
		v, _, err := envoy.CurrentVersion(o.HomeDir)
		if err != nil {
			return NewValidationError(err.Error())
		}
+		if minor pattern matches v {
+			// since this is lazy it probably will be initialized sooner and ideally plumbed through, but again that can be a separate change.
+			feV := envoy.NewFuncEVersions(ctx, o.EnvoyVersionsURL, o.Platform, o.Version)
+			latest, err = feV.FindLatestVersion(v)
+			if latest == "" {
+				return NewValidationError(err.Error())
+			}
+			o.EnvoyVersion = latest
+                 } else {
+			o.EnvoyVersion = v
+                 }
-		o.EnvoyVersion = v
```

Anyway, it pains me to ask you to change the code, but I think in any case code isn't something to hold tightly, I've deleted thousands of lines I wrote here as well tens of thousands of lines I didn't write :D Main thing is that if we focus on responsibilities, I think this change can be implemented changing less existing code and most of that will come from composition.

Hope this helps!

dio: I have added an implementation of FuncEVersions to have the `FindLatestVersion` method. Please take a look. If that makes sense I'll continue to put it in cmd.

internal/envoy/versions.go
```diff
@@ -20,30 +20,122 @@ import (
...
 
 	if err := json.Unmarshal(body, &result); err != nil {
 		return result, fmt.Errorf("error unmarshalling Envoy versions: %w", err)
 	}
+
 	return result, nil
 }
+
+// FindLatestVersions returns the latest patch version given an Envoy version.
+func (f funcEVersions) FindLatestVersions(envoyVersion version.Version) (string, error) {
```
codefromthecrypt, line 82: it makes me uncomfortable a little to pass in version.Version and pass out a string. I think we shouldn't enter this at all if we already know we have a concrete version. So, if anything it should pass out version.Version and in minorVersion

Also, we tend to use pointer style
```suggestion
// FindLatestPatch finds the latest patch version for the given minor or raises an error.
// minor is just like envoy.Version format, except missing the patch. Ex. 1.18 or 1.20_debug
func (f *funcEVersions) FindLatestPatch(minorVersion string) (version.Version, error) {
```

internal/envoy/versions.go
```diff
@@ -20,30 +20,122 @@ import (
...
+	if matched := globals.EnvoyVersionPattern.MatchString(string(envoyVersion)); matched {
```
codefromthecrypt, line 83: ex we shouldn't enter this case at all, there's no reason to assume our internal code will pass the wrong value, and if we did we'd have to test it, so it isn't really fun

This type shouldn't even need to be instantiated if someone passes a concrete version iotw.

internal/envoy/versions.go
```diff
@@ -20,30 +20,122 @@ import (
...
+
+// versionsGetter ensures the getter function is specified.
+func (f funcEVersions) versionsGetter() funcEVersionsGetter {
+	if f.getFunc != nil {
+		return f.getFunc
+	}
+	return f.Get
+}
+
+func checkLatestVersion(latest, v string) error {
```
codefromthecrypt, line 135: this isn't worth scrolling around for, can you inline the error check instead of making another function then having to name it and scroll to it?

internal/envoy/versions.go
```diff
@@ -20,30 +20,122 @@ import (
...
+				latestVersion = string(v)
+			}
+		}
+	}
+
+	return latestVersion, checkLatestVersion(latestVersion, string(envoyVersion))
+}
+
+// versionsGetter ensures the getter function is specified.
+func (f funcEVersions) versionsGetter() funcEVersionsGetter {
```
codefromthecrypt, line 128: I would rather spend the code on this being lazy vs it being a swappable implementation, especially if there's no user that swaps it. How about just initializing it so it is never null if using New?

codefromthecrypt (commented):
I think this is on the right track, just I think the new code shouldn't be called at all unless we actually need to (they specified a minor version)

ex getLatestPatch(alreadyAPatch) isn't sensible.

Otherwise, I think this is on the right track. thanks

internal/envoy/versions.go
dio: Thanks for this. Was not thinking thoroughly about this.

internal/envoy/versions.go
dio: Got it.

mathetake:
Let me know when this is ready for review! 👯

internal/envoy/versions.go
```diff
@@ -20,30 +20,121 @@ import (
...
+type funcEVersionsGetter func() (version.ReleaseVersions, error)
+
+// FuncEVersions fetches a version map form a remote URL and gets the latest patched version given
+// a version input. This input can be major.minor.patch (e.g. 1.1.1) or major.minor (1.1).
+type FuncEVersions interface {
+	// Get returns a version map from a remote URL. e.g. from globals.DefaultEnvoyVersionsURL.
+	Get() (version.ReleaseVersions, error)
+
+	// FindLatestVersions returns the latest patch version given an Envoy version.
+	FindLatestVersions(version.Version) (string, error)
```
codefromthecrypt, line 39: speaking aloud as going to be offline 🌴 one neat thing about this being an interface is that code to conditionally call FindLatestPatch is easy to guard in the cmd package.

Ex. you could have code that accepts this interface and the test cases will be easy to write as the impl can raise an error on Get() and do whatever you are testing on FindLatestVersions both without needing to worry about http calls and not requiring to bring most version parsing internal here.

dio:
@mathetake sure. Thank you! I'm looking at the e2e installing with the minor and offline cases.

dio:
@mathetake @codefromthecrypt I think I'll split the PR by first introducing FuncEVersions as an interface. After that, I'll update the "use" functionality. WDYT?

mathetake:
sg!

internal/cmd/use_test.go
```diff
@@ -73,3 +77,172 @@ func TestFuncEUse_InstallsAndWritesHomeVersion(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Equal(t, version.Version(tc.minorVersion), version.Version(f))
+
+			// Set o.EnvoyVersion to empty string so the logic for ensuring installed Envoy version works.
+			o.EnvoyVersion = ""
+			c, stdout, stderr = newApp(o)
+			require.NoError(t, c.Run([]string{"func-e", "which"}))
+			envoyPath = filepath.Join(o.HomeDir, "versions", tc.minorVersion+"."+tc.secondVersions.latestPatch, "bin", "envoy"+moreos.Exe)
+			require.Equal(t, moreos.Sprintf("%s\n", envoyPath), stdout.String())
+			require.Empty(t, stderr)
```
dio, line 155: Please let me know if you want me to wrap some of these. I don't know the preference for wrapping codes for tests. Some projects asked me to wrap, but others want it easy to read, hence copy-pasting is fine.

dio:
> Let me know when this is ready for review! 👯

When you have time, please take a look. Thanks!

codefromthecrypt:
I hereby relinquish review to the person who reviewed all my work so far: @mathetake

In other words, when Tasheshi and you feel ready, go ahead and merge. ciao until early next week 🌴 🏖️

internal/cmd/app.go
```diff
@@ -89,7 +89,9 @@ func NewApp(o *globals.GlobalOpts) *cli.App {
 		if err := setEnvoyVersionsURL(o, envoyVersionsURL); err != nil {
 			return err
 		}
-		o.FuncEVersions = envoy.NewFuncEVersions(o.EnvoyVersionsURL, o.Platform, o.Version)
+		if o.FuncEVersions == nil {
```
mathetake, line 92: I think we'd better put comment about this brunch! - this only happens in tests, right?

mathetake: sorry typo:  *branch 😄

internal/cmd/run.go
```diff
@@ -149,6 +149,17 @@ func ensureEnvoyVersion(c *cli.Context, o *globals.GlobalOpts) error {
...
 		}
 		o.EnvoyVersion = v
+		if matched := globals.EnvoyStrictMinorVersionPattern.MatchString(string(v)); matched {
+			var err error
+			var latest version.Version
+			if latest, err = o.FuncEVersions.FindLatestPatch(c.Context, v); err != nil {
+				if latest, err = getLatestInstalledPatch(o, v); err != nil {
+					return err
+				}
+				o.Logf("couldn't check the latest patch for %q for platform %q using %q instead\n", v, o.Platform, latest)
```
mathetake, line 159: maybe more friendly to mention the fact that "using the latest among installed ones"?

e2e/func-e_use_test.go
```diff
@@ -65,3 +68,100 @@ func TestFuncEUse_UnknownVersion(t *testing.T) {
...
+		stdout, stderr, err := funcEExec("--home-dir", homeDir, "which")
+		relativeEnvoyBin := filepath.Join("versions", string(upgradedVersion), "bin", "envoy"+moreos.Exe)
+		require.Contains(t, stdout, moreos.Sprintf("%s\n", relativeEnvoyBin))
+		require.Empty(t, stderr)
+		require.NoError(t, err)
+	})
+}
+
+// getVersionsRange returns the latest patch of a minor version and the one before it.
+func getVersionsRange(stdout, minor string) (max, min string) {
```
mathetake, line 155: nit: strictly speaking this is not "min", right? 😄

dio: Yeah, I'm changing this to `0` to make it more "versatile", but still set the purpose right.

internal/cmd/run.go
dio: Got it.

internal/cmd/app.go
dio: I got it @mathetake! I'm hungry already!

internal/cmd/run.go
dio: Updated. See what you think.

e2e/func-e_use_test.go
dio: Updated.

internal/cmd/use.go
```diff
@@ -60,8 +78,28 @@ func validateVersionArg(c *cli.Context) error {
...
+	}
+	// Sort so that new release dates appear first and on conflict choosing the higher version.
+	sort.Slice(rows, func(i, j int) bool {
+		if rows[i].releaseDate == rows[j].releaseDate {
+			return rows[i].version > rows[j].version
+		}
+		return rows[i].releaseDate > rows[j].releaseDate
+	})
+	for i := range rows {
+		if strings.HasPrefix(string(rows[i].version), string(minorVersion)+".") {
```
mathetake, line 100: May we could avoid allocating the string in every loop 😄 
```suggestion
	prefix := string(minorVersion)+"."
	for i := range rows {
		if strings.HasPrefix(string(rows[i].version), prefix) {
```

mathetake:
```suggestion
	for _, row := range rows {
		if strings.HasPrefix(string(row.version), string(minorVersion)+".") {
```

dio: Ah OK

internal/globals/globals.go
```diff
@@ -91,4 +93,7 @@ var (
 	EnvoyVersionPattern = regexp.MustCompile(`^[1-9][0-9]*\.[0-9]+\.[0-9]+(_debug)?$`)
 	// EnvoyMinorVersionPattern is EnvoyVersionPattern but with optional patch and _debug components.
 	EnvoyMinorVersionPattern = regexp.MustCompile(`^[1-9][0-9]*\.[0-9]+(\.[0-9]+)?(_debug)?$`)
+	// EnvoyStrictMinorVersionPattern is used to validated minor versions. A Minor version is just
+	// like envoy.Version format, except missing the patch. For example: 1.18 or 1.20_debug.
+	EnvoyStrictMinorVersionPattern = regexp.MustCompile(`^[1-9][0-9]*\.[0-9]+(_debug)?$`)
```
mathetake, line 98: I think we might want to tests to make sure this regex is actually correct to avoid future drifts?

mathetake: I should have commented this for other regexes to @codefromthecrypt 😄

dio: I can do that.

mathetake:
maybe we should modify this instruction as something like `should look like "1.19.1" or "1.19"`?
```
$ ./func-e use 1.19.
invalid [version] argument: "1.19." should look like "1.19.1"
show usage with: func-e help
```

internal/globals/globals.go
mathetake: thank you! 🎸

dio:
> maybe we should modify this instruction as something like `should look like "1.19.1" or "1.19"`?
> 
> ```
> $ ./func-e use 1.19.
> invalid [version] argument: "1.19." should look like "1.19.1"
> show usage with: func-e help
> ```

Yes, I missed the USAGE.md part. Will update it.

internal/cmd/run.go
mathetake: Brilliant!

internal/cmd/use_test.go
```diff
@@ -73,3 +77,172 @@ func TestFuncEUse_InstallsAndWritesHomeVersion(t *testing.T) {
...
+	o, cleanup := setupTest(t)
+	defer cleanup()
+
+	// The initial version to be installed.
+	initial := avaliableVersions{
+		latestPatch: "3",
+		versions:    []version.Version{"1.12.3"},
+	}
+	minorVersion := "1.12"
+	latestPatch := "3"
```
mathetake, line 170:
```suggestion
	// The initial version to be installed.
	minorVersion := "1.12"
	latestPatch := "3"
	initial := avaliableVersions{
		latestPatch:  latestPatch,
		versions:    []version.Version{minorVersion},
	}
```

internal/cmd/use_test.go
mathetake: I think this looks fine as-is! Thank you!

mathetake (approved):
Mostly LGTM now!

internal/cmd/use_test.go
dio: Love this. Thanks!

internal/cmd/use_test.go
mathetake: I think this looks fine as-is! Thank you!

internal/cmd/use_test.go
mathetake: oh my bad - nvm 😄

dio:
Thanks for your time, @mathetake! If you have other comments, please do so. Otherwise, I'll merge it in when the CI jobs are green 🙏🏽

internal/globals/globals_test.go
```diff
@@ -0,0 +1,98 @@
...
+		},
+		{
+			version:                "1.1.1_debug",
+			capturedPatchComponent: ".1",
+		},
+	}
+
+	for _, tc := range tests {
+		var matched [][]string
+		if matched = globals.EnvoyMinorVersionPattern.FindAllStringSubmatch(tc.version, -1); matched == nil {
```
mathetake, line 91:
```suggestion
		if matched = globals.EnvoyMinorVersionPattern.FindAllStringSubmatch(tc.version, -1); matched != nil {
```

dio: Oops.

mathetake (approved):
Awesome!

mathetake:
Good to go!

dio:
Thank you @mathetake @codefromthecrypt!

PR #365, opened 2021-08-31
build: Add *.go as prerequisite for building current binary

dio:
This adds sources (`*.go` except `*_test.go`) as prerequisites for building
the `current_binary` target. This forces rebuilding the `current_binary` 
target when any of `*.go` is changed.

Signed-off-by: Dhi Aurrahman <dio@tetrate.io>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,8 +59,16 @@ non_windows_platforms := darwin_amd64 darwin_arm64 linux_amd64 linux_arm64
 # TODO: arm64 on Windows https://github.com/envoyproxy/envoy/issues/17572
 windows_platforms := windows_amd64
 
-$(build_dir)/func-e_%/func-e: main.go
-	$(call go-build, $@, $^)
+# TODO: To have expression that understand globbing for any depth.
+gocodes := [!_test].go
+subdirs := $(wildcard */)
+sources := $(wildcard \
+	$(addsuffix *$(gocodes),$(subdirs)) \
+	$(addsuffix */*$(gocodes),$(subdirs)) \
+	$(addsuffix */*/*$(gocodes),$(subdirs)))
+
+$(build_dir)/func-e_%/func-e: main.go $(sources)
+	$(call go-build, $@, main.go)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -59,17 +59,25 @@ non_windows_platforms := darwin_amd64 darwin_arm64 linux_amd64 linux_arm64
...
+# TODO: To have expression that understand globbing for any depth.
+gocodes := [!_test].go
+subdirs := $(wildcard */)
+sources := $(wildcard \
+	$(addsuffix *$(gocodes),$(subdirs)) \
+	$(addsuffix */*$(gocodes),$(subdirs)) \
+	$(addsuffix */*/*$(gocodes),$(subdirs)))
+
+$(build_dir)/func-e_%/func-e: main.go $(sources)
+	$(call go-build, $@, main.go)
```
dio, line 71: I like the caret (hat?) but seems like we need to change it to `main.go`.

codefromthecrypt: According to my constant reference Managing Projects with GNU Make", `$<` should get the first dependency. Would that work?

codefromthecrypt (approved):
cool. also pragmatic to add a TODO vs fuss forever (as I often do ;))

Makefile
```diff
@@ -59,17 +59,25 @@ non_windows_platforms := darwin_amd64 darwin_arm64 linux_amd64 linux_arm64
 # TODO: arm64 on Windows https://github.com/envoyproxy/envoy/issues/17572
 windows_platforms := windows_amd64
 
-$(build_dir)/func-e_%/func-e: main.go
-	$(call go-build, $@, $^)
+# TODO: To have expression that understand globbing for any depth.
+gocodes := [!_test].go
```
codefromthecrypt, line 63: since our code is always in the internal subdir (except main.go) can we tighten this?

Makefile
dio: Ah, yeah! Thanks, @codefromthecrypt!

codefromthecrypt:
well done!

PR #366, opened 2021-09-01
Implements linux packaging with nfpm

### LINKED ISSUE #345: Publish Linux system packages (debian and RPM)
Right now, to install func-e on Linux requires homebrew (ubuntu-only and somewhat slow) or using curl. It is a more idiomatic experience to have system packages, at least debian and RPM.

create, sign and publish RPM and Debian packages for func-e

This implies work including researching where to publish them as well man page generation

not do anything

**Additional context**
This also helps users work around lack of [envoy system packages](https://github.com/envoyproxy/envoy/issues/16867). Basically, they can install func-e then choose the version they need, even run it without func-e using the `func-e which` command to locate the executable.

codefromthecrypt:
This creates Debian and RPM packages on `make dist` and tests them on each commit.

Upon release, these are signed with an org secret.

fixes #345

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -69,6 +69,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  # Note: We don't test arm64 on release as it is unlikely to fail and too much effort.
   e2e:
     needs: func-e
     name: Run e2e tests (${{ matrix.os }})
@@ -78,7 +79,7 @@ jobs:
       matrix:
         include:
           - os: ubuntu-20.04  # Hard-coding an LTS means maintenance, but only once each 2 years!
-            pattern: '*linux_amd64.tar.gz'
+            pattern: '*linux_amd64*'
           - os: macos-latest
             pattern: '*darwin_amd64.tar.gz'
           - os: windows-latest
@@ -112,3 +113,17 @@ jobs:
         shell: cmd
         env:  # use the stashed msi name instead of parsing it
           MSI_FILE: ${{ steps.download.outputs.msi }}
+
+      - name: "Test Debian package"
+        if: runner.os == 'Linux'
+        run: |
+          mkdir -p dist && mv *.deb dist/
+          packaging/nfpm/verify_deb.sh
+
+      - name: "Test RPM package (CentOS)"
+        if: runner.os == 'Linux'
+        run: |
+          mkdir -p dist && mv *.rpm dist/
+          docker run --rm -v $PWD:/work --entrypoint packaging/nfpm/verify_rpm.sh ${CENTOS_IMAGE}
+        env:  # CENTOS_IMAGE was built by internal-images.yaml
+          CENTOS_IMAGE: ghcr.io/tetratelabs/func-e-internal:centos-8
--- a/packaging/nfpm/verify_deb.sh
+++ b/packaging/nfpm/verify_deb.sh
@@ -26,7 +26,7 @@ case $(uname -m) in
     exit 1;
 esac
 
-deb_file=${deb_file:-$(ls dist/func-e_*_linux_${deb_arch}.deb)}
+deb_file=${1:-$(ls dist/func-e_*_linux_${deb_arch}.deb)}
 
 echo installing "${deb_file}"
 sudo dpkg -i "${deb_file}"
--- a/packaging/nfpm/verify_rpm.sh
+++ b/packaging/nfpm/verify_rpm.sh
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-rpm_file=${rpm_file:-$(ls dist/func-e_*_linux_$(uname -m).rpm)}
+rpm_file=${1:-$(ls dist/func-e_*_linux_$(uname -m).rpm)}
 
 echo "installing ${rpm_file}"
 sudo rpm -i "${rpm_file}"


codefromthecrypt:
@dio as I'll be offline at least another day, if you are interested in carrying this one on, it would be helpful. I mentioned in the PR description notes that basically we should make sure our GitHub actions can install and run the rpm and deb just like msi (and probably the more coherent way to do that is to refactor msi.yaml into a matrix test that also does rpm and deb). There may be some devil in details ex centos isn't an operating system choice here. You can see in one of the GitHub actions where we are using docker to run centos for this reason. If there isn't enough bread crumbs here to carry it on, feel free to let it sit, just might be nice to get into the next release and possibly not a huge reason for me to sit on it

dio:
@codefromthecrypt I'll try to work on it. Take care, @codefromthecrypt!

dio:
- [ ] verify script (install/uninstall)
- [ ] job using verify script to verify install/uninstall on x86 (`ubuntu` (`_amd64.deb`), `centos` (`_x86_64.rpm`))
- [ ] job using verify script to verify install/uninstall on non-x86/aarch64 (`ubuntu` (`_arm64.deb`), `fedora` (`_aarch64.rpm`))

.github/workflows/packaging.yaml
```diff
@@ -2,27 +2,30 @@
...
 
 env:  # As of 2021-08-05 images come with go 1.15 https://github.com/actions/virtual-environments/tree/main/images
   GO_VERSION: "1.17.0"
 
 on:
   push:  # We run tests on non-tagged pushes to master
-    tags: ''
-    branches: master
+    tags:
+      - ""
```
dio, line 13: Not sure if we really need this since this fails the validation with (https://json.schemastore.org/github-workflow.json): `String is shorter than the minimum length of 1.yaml-schema: https://json.schemastore.org/github-workflow.json`

.github/workflows/release.yaml
```diff
@@ -10,6 +10,10 @@ on:
```
dio, line 10: What is the best way to test this job? Manually?

codefromthecrypt: you can see at the bottom of this file "Test Windows Installer (Windows)" which is similar to what was in the msi.yaml. Assuming msi.yaml -> packaging.yaml works on commit, we should expect low risk doing similarly here for the other two

codefromthecrypt: once that's in, the next question is.. would it work on release? One key is that the release.yaml results in a *draft* release not a published one. What I've done in the past is merged things and cut a release tag for the next one. If there was a problem, push a fix to master, delete the release and the tag and push the tag again. Once the release.yaml e2e tests work (via a successful draft release), then it is safe to publish a real release.. either the one just tested or drop the release and push it again.  make sense?

.github/workflows/packaging.yaml
codefromthecrypt: I recall having to add this directly in Zipkin before in order to stop this workflow from running on a release tag. I'm not sure how authoritative that schema is or not. If this is authoritative, we should note to verify the schema using it as that would assume that the current validation hint at the top of the file isn't good enough.

In any case, we'd know it was needed if the explicit empty was deleted, you push a tag to master and this ran as a result. That's an experiment that could be done in another repo as well.

packaging/nfpm/verify.sh
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+# The packaging type. This can be .rpm or .deb.
+PACKAGING=${PACKAGING:-deb}
```
codefromthecrypt, line 19: I think this file is over-normalized. To figure out what's going on requires thinking too much, like which variables are passing to the almost identical two workflows and do they or do they not result in a rpm or deb run?

Let's keep it more straightforward.. how about verify_rpm verify_deb instead? the cost of duplicating `func-e -version || exit 1` a couple more times is worth buying the straight-forwardness

.github/workflows/packaging.yaml
```diff
@@ -94,3 +97,87 @@ jobs:
...
+  linux-x86:
+    name: "Test installing on (${{ matrix.image }}/${{ matrix.arch }})"
+    needs: dist
+    runs-on: "ubuntu-latest"
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - image: "ubuntu:latest"
+            arch: amd64
```
codefromthecrypt, line 117: let's use arch here to represent the docker arch, not a filename expression, as the mapping is confusing.

.github/workflows/packaging.yaml
```diff
@@ -94,3 +97,87 @@ jobs:
...
+            packaging: deb
+          - image: "ghcr.io/tetratelabs/func-e-internal:centos8"
```
codefromthecrypt, line 119: please use the built-in container feature not manually running docker as especially in normal case we don't need to. You can look at the file I meant to reference for this commit.yaml
 
```
      matrix:
        container: ["ghcr.io/tetratelabs/func-e-internal:centos8"]
```

.github/workflows/packaging.yaml
```diff
@@ -94,3 +97,87 @@ jobs:
...
+      - name: "Install .${{ matrix.packaging }} on ${{ matrix.image }} (${{ matrix.arch }})"
+        run: |
+          docker run -e ARCH=${{ matrix.arch }} \
+            -e PACKAGING=${{ matrix.packaging }} \
+            -v ${{ github.workspace }}:/src \
+            -w /src \
+            ${{ matrix.image }} \
+            packaging/nfpm/verify.sh
+
+  linux-non-x86:
```
codefromthecrypt, line 142: this is pretty nasty

We are currently using travis to test arm64, let's please not use docker explicitly and drift how things are tested feature by feature. Ack you probably haven't seen enough of what's going on, yet, but it is a lot easier to be consistent in approach vs have one change be different only for one thing, as we'd have to remember even more context when doing things.

codefromthecrypt (commented):
Thanks for giving this a try. I think my main feedback is around the inconsistency here, doing things differently to accomplish similar goals.  That's really not a good idea as it adds even more distraction if merged.

One thing I'll ask always is to not make one-off changes to build config or really anything that's otherwise done consistently. Ex formatting one file completely different to the others doesn't help maintenance, and instead it makes the project look and feel sloppy.  Please roll back the formatting or do it in another PR to all files.

Similarly it is unnecessary to deviate from how we test on centos depending on if we are using rpm vs just running func-e directly on e2e. Try to research how other things are done first for the same reason as formatting.. That will result in even more important consistency. Docker usually isn't a game we'd want to get into, and luckily between the `container` directive here and native runners on travis for arm, we really don't need to do most or any of explicit dockering.

I think the idea end result is a single linux matrix for both packaging and release. arm64 if tested only on commit with travis, even if only partially tested, is better than introducing a fairly gnarly GitHub action. If we did want to go the pure emulation route, and inherit the gnarly action we'd probably want to remove travis. Having one leg in each canoe is on the other hand worse.

dio:
I reverted my changes. And convert it to a draft. Obviously, I'm still far behind in understanding what consistency in code means.

.github/workflows/packaging.yaml
```diff
@@ -94,3 +94,27 @@ jobs:
...
+        if: runner.os == 'Linux'
+        uses: docker://ghcr.io/tetratelabs/func-e-internal:centos8
+        with:
+          # Since we use the root user from within the container process when doing this, we don't
+          # require "sudo".
+          args: packaging/nfpm/verify_rpm.sh
+
+      # TODO: Verification for installing arm64 Debian and aarch64 RPM packages. Curently it is
+      # not attempted since we need to pass the artifacts to Travis CI every time this job is
+      # finished.
```
dio, line 120: Since currently I don't want to be creative on this, and not sure if the creativity will be still in the scope of this change.

.github/workflows/packaging.yaml
```diff
@@ -94,3 +94,27 @@ jobs:
...
+        # Running dpkg and apt-get requires "sudo".
+        run: sudo packaging/nfpm/verify_deb.sh
+
+      # This only checks installing the RPM package when running this job on Linux, but using
+      # container (on CentOS 8 via ghcr.io/tetratelabs/func-e-internal:centos8, x86_64) as it is
+      # more efficient than having a new job definition (which requires to upload and download the
+      # built packages as artifacts) and do the verification there.
+      - name: "Test installing RPM package (on CentOS 8, x86_64 via ghcr.io/tetratelabs/func-e-internal:centos8)"
+        if: runner.os == 'Linux'
+        uses: docker://ghcr.io/tetratelabs/func-e-internal:centos8
```
dio, line 112: Please yell at me if the comment on this block above does not sufficiently justify this.

.github/workflows/packaging.yaml
```diff
@@ -94,3 +94,27 @@ jobs:
...
+        with:
+          # Since we use the root user from within the container process when doing this, we don't
+          # require "sudo".
+          args: packaging/nfpm/verify_rpm.sh
```
dio, line 116: Use different scripts for verifying rpm and deb to facilitate deb or rpm specific checks or tasks in the script later (if we want to do it later) vs. having to branch in a single script for deb or rpm cases.

.github/workflows/packaging.yaml
```diff
@@ -1,8 +1,8 @@
 # `name` value will appear "as is" in the badge.
 # See https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository
-# yamllint --format github .github/workflows/msi.yaml
+# yamllint --format github .github/workflows/packaging.yaml
```
dio, line 3: Hope this s/msi/packaging doesn't hurt anything.

.github/workflows/packaging.yaml
```diff
@@ -94,3 +94,27 @@ jobs:
         run: |
           ./packaging/msi/winget_manifest.sh > Tetrate.func-e.yaml
           yamllint -sd '{extends: default, rules: {line-length: disable}}' Tetrate.func-e.yaml
+
+      # This only checks installing the Debian package (amd64) when running this job step on Linux
+      # (Ubuntu, amd64) as it is more efficient than having a new job definition (which requires to
+      # upload and download the built packages as artifacts) and do the verification there.
+      - name: "Test installing Debian package (Ubuntu, amd64)"
```
dio, line 101: Following the footsteps of testing Windows .msi here. Not sure if the reasoning (hence the decision to add it here) is correct.

.github/workflows/packaging.yaml
```diff
@@ -78,7 +78,7 @@ jobs:
           key: test-${{ runner.os }}-${{ env.GO_VERSION }}-go-${{ hashFiles('go.sum', 'Makefile') }}
           restore-keys: test-${{ runner.os }}-${{ env.GO_VERSION }}-go-
 
-      - name: "Build Windows Installer (MSI)"
+      - name: "Build packages (.msi, .deb [amd64, arm64], .rpm [x86_64, aarch64])"
```
dio, line 81: If this is too much, please scream at me what the perfect description for this is. (needed to change since it is not all about "Windows Installer (MSI)" anymore).

.github/workflows/packaging.yaml
```diff
@@ -33,8 +33,8 @@ defaults:
     shell: bash
 
 jobs:
-  msi:
-    name: "Test Windows Installer build (${{ matrix.os }})"
+  packaging:
+    name: "Test packaging build (${{ matrix.os }})"
```
dio, line 37: Since it is not all about Windows Installer anymore. Is "packaging" here is the correct replacement?

.github/workflows/packaging.yaml
```diff
@@ -15,14 +15,14 @@ on:
       - 'packaging/msi/*'
       - 'packaging/icon@48w.ico'
       - '.github/workflows/msi.yaml'
-      - 'Next.mk'
+      - 'Makefile'
```
dio, line 18: Since `Next.mk` is removed. And, yes no formatting changes here.

.github/workflows/packaging.yaml
```diff
@@ -94,3 +94,28 @@ jobs:
         run: |
           ./packaging/msi/winget_manifest.sh > Tetrate.func-e.yaml
           yamllint -sd '{extends: default, rules: {line-length: disable}}' Tetrate.func-e.yaml
+
+      # This only checks installing and unistalling the Debian package (amd64) when running this job
+      # step on Linux (Ubuntu, amd64) as it is more efficient than having a new job definition
+      # (which requires to upload and download the built packages as artifacts) and do the
+      # verification there.
+      - name: "Test installing and unistalling Debian package (on Ubuntu, amd64)"
```
dio, line 102: Following the footsteps of `- name: "Test Windows Installer (Windows)"` step. Not sure if this has sufficient reasoning hence the decision to put it here is correct.

.github/workflows/packaging.yaml
```diff
@@ -94,3 +94,28 @@ jobs:
...
+        if: runner.os == 'Linux'
+        uses: docker://ghcr.io/tetratelabs/func-e-internal:centos8
+        with:
+          # Since we use the root user from within the container process when doing this, we don't
+          # require "sudo".
+          args: packaging/nfpm/verify_rpm.sh
+
+      # TODO: Verification for installing and unistalling arm64 Debian and aarch64 RPM packages.
+      # Curently it is not attempted since we need to pass the artifacts to Travis CI every time
+      # this job is finished.
```
dio, line 121: Since currently I don't want to be creative on this, and not sure if the creativity will be still in the scope of this change.

For example: running make dist on graviton.

dio: Do you have ideas on this @mathetake?

.github/workflows/release.yaml
```diff
@@ -43,8 +55,11 @@ jobs:
         run: sudo apt update -qq && sudo apt install -qq -y wixl osslsigncode
 
       - name: "Make release assets"
-        run: make release
+        run: make dist
```
dio, line 58: I hope this is correct.

.github/workflows/packaging.yaml
dio: I'm thinking to remove Travis and use `self-hosted` GHA instead. Probably in another PR.

.github/workflows/packaging.yaml
```diff
@@ -94,3 +94,28 @@ jobs:
         run: |
           ./packaging/msi/winget_manifest.sh > Tetrate.func-e.yaml
           yamllint -sd '{extends: default, rules: {line-length: disable}}' Tetrate.func-e.yaml
+
+      # This only checks installing and uninstalling the Debian package (amd64) when running this job
+      # step on Linux (Ubuntu, amd64) as it is more efficient than having a new job definition
+      # (which requires to upload and download the built packages as artifacts) and do the
+      # verification there.
+      - name: "Test installing and uninstalling Debian package (on Ubuntu, amd64)"
```
dio, line 102: Following the Windows MSI footsteps to install/uninstall the Debian package here.

.github/workflows/packaging.yaml
codefromthecrypt: please don't as this is worse and docker in general breaks in many releases as well emulation setup is quite more problematic than the risk that the RPM verification will fail. What travis is doing is infinitely more simple to troubleshoot than emulation. You can feel free to use emulation on your own, but please don't pull that in here. I won't be fixing it months later when it invariably breaks due to some strange relicensing of docker or arch mismatch problem. you can notice even zipkin's build recently broke due to a docker change in emu

codefromthecrypt:
one thing I can hope we can do is "be consistent and use the right tool for the job" I have made countless requests to default to practice that already exists here, as a lot of thought went into that. The second classification of requests were to find the simplest way out. Introducing more layers of technology (ex buildx, emu etc) is the opposite of this and generally how all projects end up a mess at some point and/or abandoned.

I'm not interested in stressing myself out repeating the same requests again and again, and hope that you can please focus on the smallest change first vs holding something hostage to a complex set of infrastructure changes. we have time to discuss changes but as I've noticed many times, the current status quo is not understood, and until it is, it isn't the right time to change it.

.github/workflows/packaging.yaml
codefromthecrypt: good that was drift worth fixing!

.github/workflows/packaging.yaml
codefromthecrypt: yep

codefromthecrypt:
UPDATE:

GHA and Travis as of #383 run e2e tests on the matrix [ubuntu,centos]*[amd64,arm64]

NEXT STEP:

Rebase this and use the same centos image to run the rpm tests

codefromthecrypt:
UPDATE:

In order to have coherent builds across platforms and CI, Make needs to own GOROOT selection and configuration. https://github.com/tetratelabs/func-e/pull/386 implements that.

NEXT ACTION:

Once the above is merged, complete GHA runner-style docker images, use them for e2e tests and ultimately here for RPM and Debian tests.

codefromthecrypt:
https://github.com/tetratelabs/func-e/pull/387 will push the docker images we need, once finished

codefromthecrypt:
thanks for the help on this @dio!

codefromthecrypt:
pulled the package signing because it never worked. the test scripts didn't verify, so they didn't notice. Make didn't export the variables, so signing was silently ignored until release which exported them https://github.com/tetratelabs/func-e/pull/395

PR #367, opened 2021-09-01
Add FuncEVersions interface and its default implementation

dio:
This patch introduces FuncEVersions as an interface for easier testing,
especially when simulating failure on getting the remote Envoy release
versions map.

This also provides the default implementation of FuncEVersions, which
converts the existing functionality via Get() method and also a way to
find a latest patch given Envoy release versions map and a minor
version.

### DIFF
--- a/internal/envoy/versions.go
+++ b/internal/envoy/versions.go
@@ -20,30 +20,94 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
+	"strings"
 
+	"github.com/tetratelabs/func-e/internal/globals"
 	"github.com/tetratelabs/func-e/internal/version"
 )
 
-// FuncEVersions returns a version map from a remote URL. eg globals.DefaultEnvoyVersionsURL.
-func FuncEVersions(ctx context.Context, envoyVersionsURL string, p version.Platform, v version.Version) (version.ReleaseVersions, error) {
+// NewFuncEVersions creates a new Envoy versions fetcher.
+func NewFuncEVersions(envoyVersionsURL string, p version.Platform, v version.Version) version.FuncEVersions {
+	feV := &funcEVersions{envoyVersionsURL: envoyVersionsURL, platform: p, version: v}
+	feV.getFunc = feV.Get
+	return feV
+}
+
+type funcEVersions struct {
+	envoyVersionsURL string
+	platform         version.Platform
+	version          version.Version
+
+	// getFunc allows to override the release versions getter implementation.
+	getFunc func(ctx context.Context) (version.ReleaseVersions, error)
+}
+
+var _ version.FuncEVersions = (*funcEVersions)(nil)
+
+// Get implements fetching the Envoy versions from the specified Envoy versions URL.
+func (f *funcEVersions) Get(ctx context.Context) (version.ReleaseVersions, error) {
 	result := version.ReleaseVersions{}
 	// #nosec => This is by design, users can call out to wherever they like!
-	resp, err := httpGet(ctx, envoyVersionsURL, p, v)
+	resp, err := httpGet(ctx, f.envoyVersionsURL, f.platform, f.version)
 	if err != nil {
 		return result, err
 	}
 	defer resp.Body.Close() //nolint
 
 	if resp.StatusCode != http.StatusOK {
-		return result, fmt.Errorf("received %v status code from %v", resp.StatusCode, envoyVersionsURL)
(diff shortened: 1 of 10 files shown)


internal/version/versions.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2021 Tetrate
```
dio, line 1: I'm not sure if this is the right package for this file. I've tried to introduce another package for this interface only: "remote", but felt not right. Please let me know the right location for this. Thank you!

internal/envoy/versions.go
```diff
@@ -20,30 +20,94 @@ import (
...
 	if err := json.Unmarshal(body, &result); err != nil {
 		return result, fmt.Errorf("error unmarshalling Envoy versions: %w", err)
 	}
 	return result, nil
 }
+
+// FindLatestPatch implements finding the latest patch version for the given minor version or raises
+// an error. The Envoy release versions fetching logic can be overridden by setting the getFunc with
+// different implementation.
+func (f *funcEVersions) FindLatestPatch(ctx context.Context, minorVersion version.Version) (version.Version, error) {
```
dio, line 75: This will be used later when we introduce `func-e use <minor version>`, as an example: `func-e use 1.18`

internal/version/versions.go
mathetake: I feel this is the right place to put! +1

internal/envoy/versions_test.go
```diff
@@ -0,0 +1,91 @@
...
+func newFuncEVersionsTester(versions map[version.Version]version.Release) funcEVersionsTester {
+	return funcEVersionsTester{
+		feV: funcEVersions{
+			// Override Envoy versions getter for testing purpose only.
+			getFunc: func(context.Context) (version.ReleaseVersions, error) {
+				return version.ReleaseVersions{Versions: versions}, nil
+			},
+		},
+	}
+}
```
mathetake, line 91: I think maybe we don't need getFunc by just embedding  funcEVersions into funcEVersionsTester and override Get directly like this? 😄 

```suggestion
type funcEVersionsTester struct {
	funcEVersions // embed!
	versions map[version.Version]version.Release
}

// Override version.FuncEVersions
func (f *funcEVersions) Get(ctx context.Context) (version.ReleaseVersions, error) {
	return version.ReleaseVersions{Versions: versions}, nil
}

func newFuncEVersionsTester(versions map[version.Version]version.Release) version.FuncEVersions {
	return funcEVersionsTester{
		feV: funcEVersions{}
		versions: versions,
	}
}
```

internal/envoy/versions_test.go
```diff
@@ -0,0 +1,91 @@
...
+		},
+	}
+
+	ctx := context.Background()
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			tester := newFuncEVersionsTester(tc.versions)
+			have, err := tester.feV.FindLatestPatch(ctx, tc.input)
+			if tc.want == "" {
+				require.Error(t, fmt.Errorf("couldn't find latest version for %s", tc.input))
```
mathetake, line 69: I guess we could do something like this!
```suggestion
				require.Errorf(t, "couldn't find latest version for %s", tc.input)
```

internal/envoy/versions_test.go
dio: I think I have tried this before. Calling `Get` inside "parent's" `FindLatestPatch` is not overridden, if we call it from the tester object. But not sure: https://play.golang.org/p/aXL3DdUxSAg.

internal/envoy/versions_test.go
dio: Cool. Thanks!

mathetake (approved):
LGTM!

internal/envoy/versions.go
```diff
@@ -20,30 +20,94 @@ import (
...
+type funcEVersions struct {
+	envoyVersionsURL string
+	platform         version.Platform
+	version          version.Version
+
+	// getFunc allows to override the release versions getter implementation.
+	getFunc func(ctx context.Context) (version.ReleaseVersions, error)
+}
+
+var _ version.FuncEVersions = (*funcEVersions)(nil)
```
codefromthecrypt, line 46: comment or delete please

codefromthecrypt (commented):
Thanks for putting this in. I'm also uncomfortable about having an interface, yet needing a function overrider only for tests. Seems something isn't quite right.

That said the bulk of the code is in the right spot, so makes for an easy refactor if we can do better. Thanks for putting this in!

PR #368, opened 2021-09-06
Parse date string using system local as its timezone

dio:
This fixes the timezone bug when doing `Chtimes`. By default `time.Parse`
uses `UTC` while we use `time.Modtime` which uses `time.Local` as its
location.

Fixes https://github.com/tetratelabs/func-e/issues/303.

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -105,10 +105,6 @@ jobs:
           # See: http://www.unicode.org/cldr/charts/latest/supplemental/zone_tzid.html.
           timezoneWindows: "US Mountain Standard Time"
 
-      - name: "Check to be sure"
-        if: runner.os == 'Linux'
-        run: timedatectl
-
       - name: "Run unit tests in a different timezone"
         run: make test
 
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -97,16 +97,11 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: bash <(curl -s https://codecov.io/bash)
 
-      - name: "Change timezone" # https://github.com/tetratelabs/func-e/issues/303.
-        uses: szenius/set-timezone@v1.0
-        with:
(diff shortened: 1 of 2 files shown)


dio:
The GHA used [szenius/set-timezone@v1.0](https://github.com/szenius/set-timezone) for changing TZ works:

image

dio:
Before this patch: https://github.com/tetratelabs/func-e/issues/303#issuecomment-913344140.

.github/workflows/commit.yaml
```diff
@@ -97,6 +97,17 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: bash <(curl -s https://codecov.io/bash)
 
+      - name: "Change timezone" # https://github.com/tetratelabs/func-e/issues/303.
+        uses: szenius/set-timezone@v1.0
```
mathetake, line 101: Is this something we can rely on? I mean I want to avoid introducing fragile third party actions dep vs manually setting timezone😄

dio: We can do Linux only on this if you prefer that one. This deps basically do exec-command:

https://github.com/szenius/set-timezone/blob/master/index.js#L13-L24

mathetake (approved):
Thank you💯🎉

.github/workflows/commit.yaml
mathetake: Thanks, only doing in Linux sounds reasonable!

dio:
Thanks for the review @mathetake! 🙇🏽

.github/workflows/commit.yaml
```diff
@@ -97,6 +97,12 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: bash <(curl -s https://codecov.io/bash)
 
+      - name: "Run unit tests in a different timezone" # https://github.com/tetratelabs/func-e/issues/303.
```
codefromthecrypt, line 100: 😍

PR #371, opened 2021-09-11
Add fakebinary for acommodating testing the child process management

dio:
This adds the `internal/test/fakebinary` package allowing us to test running a fake `func-e` process that spawns a fake `envoy`. This allows us to isolate the process control angle and can lead to quicker diagnosis of platform-specific behaviors, as well as show potential for future process abstraction. For example: to check the behavior of `func-e` when we send `SIGKILL` to it (in Linux, we need to make sure the child dies and receive whatever signal we have configured in `SysProcAttr.Pdeathsig`).

We changed the setting for `SysProcAttr.Pdeathsig` to `SIGKILL` since when anyone sends `SIGKILL` on Linux to `func-e`, we want the spawned envoy to receive `SIGKILL` as well. Reference: https://github.com/tetratelabs/func-e/pull/371#discussion_r707015768.

This also abstracts away the process for building any fake binary that optionally uses some of `func-e` internal packages. For example, in this change, we can isolate the build for fake func-e to use `internal/moreos` package (to check on `moreos.ProcessGroupAttr()`) only and no more than that what it requires.


internal/moreos/test/moreos_linux_test.go
```diff
@@ -0,0 +1,105 @@
...
+		if expectedErr == nil {
+			require.FailNow(t, "timeout waiting for finding process with no error")
+		}
+		require.FailNow(t, "timeout waiting for expected error %v", expectedErr)
+	}
+}
+
+func requireFindProcessNoError(t *testing.T, proc *os.Process) {
+	requireFindProcessError(t, proc, nil) // expect to find the process with no error.
+}
```
codefromthecrypt, line 105: This part while there will be some kinks is pretty on the money. I will have some comments to make, but nothing huge.

internal/test/fake_binary.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/tetratelabs/func-e/internal/moreos"
+)
+
+var (
+	_, currentFilePath, _, _ = runtime.Caller(0)
+
+	// funcEGoModDir points to the directory where the func-e's go.mod resides.
```
codefromthecrypt, line 32: this is neat, but it isn't as isolated as I was hoping we could be. the variant here accomplishes the minimum dependencies, and will help us avoid bloat. For example, we shouldn't be adding massive deps like gopsutil into moreos. isolating helps avoid that https://github.com/tetratelabs/func-e/compare/fake-deux

internal/test/testdata/fake_func-e/fake_func-e.go
```diff
@@ -0,0 +1,39 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/tetratelabs/func-e/internal/moreos"
+	"github.com/tetratelabs/func-e/internal/version"
+)
+
+func main() {
+	envoy := filepath.Join(os.Getenv("FUNC_E_HOME"), "versions", string(version.LastKnownEnvoy), "envoy")
```
codefromthecrypt, line 32: this source doesn't exercise enough of how the subprocess is controlled, though I appreciate the start first. Also, we should be eagle eye on any dev (ex. envoy. here). Finally, the main should be runnable normally.. one way to test is to use an actual path to envoy in it.. you'd then possibly notice the paths aren't right here. I made a more complete example here, and hopefully by the time we are done there is less copy/pasta involved https://github.com/tetratelabs/func-e/compare/fake-deux

codefromthecrypt (commented):
Thanks for narrowing the scope and giving another stab. The end of the test looks good.

I have a more preferred approach to isolation of the incidental binary, ex so that it doesn't end up depending on anything but moreos. I also prefer to be more complete in how we launch processes as there are a couple key utilities used by envoy.Run. I made a more comprehensive sketch here, which will replace most of the code except the bottom of the test you wrote. https://github.com/tetratelabs/func-e/compare/fake-deux

Note: It took a while for me to figure out why things were moving around. While your description says the test case desired in text, and that's good, we should try to make notes either in the description about other structural change and why or self-PR comments. This helps explain the incidental change and why, in this case around needing to get a relative path to the module (the reason is that the fake source is different than fake envoy as it is importing a fully-qualified go import). When change is explained, it can help others understand the why vs have them guessing or figuring it out from scratch in a review out of timezone.

internal/test/fake_binary.go
dio: Yes, I was either do the copying or this. Was optimizing for minimal diff. But the sketch looks good, I'll accommodate that. Thanks, @codefromthecrypt.

internal/test/func-e.go
```diff
@@ -0,0 +1,45 @@
...
+	"path/filepath"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	// fakeFuncESrcPath is a path to the test source file used to simulate func-e which runs envoy as its child.
+	fakeFuncESrcPath = filepath.Join(funcEGoModDir, "internal", "test", "testdata", "fake_func-e", "fake_func-e.go")
```
codefromthecrypt, line 28: done in the other branch, but noting for completion that go:embed is superior to this style if for no other reason IDEs can help make it work (ex refactoring and navigation). OTOH you have to play by the rules and only do it where the path you need is enclosed by the current directory downwards.

another nit you'll notice in the other change: we aren't ambiguous in the testdata about the main package, so we can use fake_func-e.go instead of making a directory and then a file named almost the same as the directory. If later the moreos testdata gets crowded, since we are using go:embed at least renaming or moving that with goland will fix the go:embed reference. IOTW we can push any tension about directory contents (ex don't need to pre-emptively make a dir with only one file in it), because refactoring later is easy with go:embed.

dio:
@codefromthecrypt do you want me to continue this or you to create a PR from your branch?

internal/moreos/test/moreos_linux_test.go
dio: What do you mean by: "This part while there will be some kinks is pretty on the money."? My English seems bad to capture the real meaning of it. But if you have comments, please add them. I'll be happy to learn from it.

codefromthecrypt: sorry by "on the money" I meant exactly right. I prefixed "pretty" because a couple nits, but I didn't want to nit while some other things are in play.

codefromthecrypt (commented):
main thing I meant was that retain almost all "moreos_linux_test.go" except the setup which is different per the other branch. Feel free to continue that branch and raise a PR out of it or force push this one and delete the other.

It would be nice to not increment notify watchers with new issue numbers as much as we have been, so I prefer to not raise yet another PR on the same topic. IOTW I prefer to rework this PR vs drop another PR and create another one on the same topic. That's why I didn't raise a PR on the other branch.

internal/moreos/test/moreos_linux_test.go
codefromthecrypt: so translating from Adrian into English "This part while there will be some kinks is pretty on the money."  "this part is almost exactly right, and we can save the minor details for later" :D

internal/moreos/moreos_func-e_linux_test.go
```diff
@@ -0,0 +1,114 @@
...
+	require.Equal(t, len(children), 1) // Should have only one child process i.e. the fake envoy process.
+	fakeEnvoyProcess := &os.Process{Pid: int(children[0].Pid)}
+	requireFindProcessNoError(t, fakeEnvoyProcess)
+
+	// Kill the fake func-e process.
+	// This works only for linux, sending kill -9 on darwin will not kill the process, we need to kill
+	// via pgid or kill the child first.
+	require.NoError(t, cmd.Process.Kill())
+
+	// The child process is expected to receive ENVOY_SIGTERM.
```
dio, line 84: This is in sync with the Pdeathsig configuration.

internal/moreos/moreos_func-e_linux_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	fakeFuncEProcess, err := process.NewProcess(int32(cmd.Process.Pid))
+	require.NoError(t, err)
+
+	// Get all fake func-e children processes.
+	children, err := fakeFuncEProcess.Children()
+	require.NoError(t, err)
+	require.Equal(t, len(children), 1) // Should have only one child process i.e. the fake envoy process.
+	fakeEnvoyProcess := &os.Process{Pid: int(children[0].Pid)}
+	requireFindProcessNoError(t, fakeEnvoyProcess)
```
dio, line 77: We can remove this since https://github.com/tetratelabs/func-e/pull/371/files#diff-ef580e2a8cde7ec5cdae2de31d239791be6dfb1bbc37c0ef9b51d6db5fa26659R85 implicitly tells that the fake envoy process is up and receiving SIGTERM.

dio:
@codefromthecrypt when you have time, please take a look. Mostly I migrated everything you have done to this branch, so mostly you have seen all of these before 😅.

internal/moreos/moreos_func-e_test.go
```diff
@@ -0,0 +1,84 @@
...
+	fakeFuncESrc []byte // Embedding the fakeFuncESrc is easier than file I/O and ensures it doesn't skew coverage
+
+	// Include the source imported by fakeFuncESrc directly and indirectly
+	//go:embed moreos.go
+	//go:embed proc_*.go
+	moreosSrcDir embed.FS
+)
+
+// Test_X tests multiple features include moreos.EnsureProcessDone and moreos.ProcessGroupAttr to ensure X occurs when Y
+func Test_X(t *testing.T) {
```
codefromthecrypt, line 42: Test_X was a placeholder to weave in TestProcessGroupAttr_Kill :D

dio: Oops

dio: re: `Test_X` do we want a separated test for when func-e successfully running envoy vs immediate exit?

internal/moreos/moreos_func-e_linux_test.go
codefromthecrypt: if you can remove these lines, do it! especially it simplifies by removing a function overload I think. thanks for the explanation.

internal/moreos/moreos_func-e_linux_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/tetratelabs/func-e/internal/test/fakebinary"
+	"github.com/tetratelabs/func-e/internal/version"
+)
+
+// TestProcessGroupAttr_Kill tests sending SIGKILL to fake func-e and makes sure fake envoy receives
+// SIGTERM (as defined in Pdeathsig).
+func TestProcessGroupAttr_Kill(t *testing.T) {
```
codefromthecrypt, line 37: I prefer to use a condition on `runtime.GOOS` vs `_linux` because myself and others might refactor something and it will break compilation only on CI.

So, basically handle this in a `t.Skip()` instead of build condition please? We can do this because there are no linux-specific code called here. When doing the skip, do note why it won't work in darwin or windows in case we can follow-up.

internal/moreos/moreos_func-e_linux_test.go
codefromthecrypt: wow thanks for calling attention to this!

I think that there wasn't much thought into setting Pdeathsig as SIGTERM instead of `Pdeathsig: syscall.SIGKILL`. We inherited that from getenvoy, and it seems nonsense that a user sending kill -9 to func-e wants to not kill -9 envoy.

We can choose from one of...
* comment in the code here `TODO: moreos.ProcessGroupAttr propagates SIGTERM on SIGKILL.. should it?* comment in `moreos.ProcessGroupAttr` why in the world we are propagating term on kill
* propagate SIGKILL on SIGKILL!

internal/moreos/moreos_func-e_linux_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Get all fake func-e children processes.
+	children, err := fakeFuncEProcess.Children()
+	require.NoError(t, err)
+	require.Equal(t, len(children), 1) // Should have only one child process i.e. the fake envoy process.
+	fakeEnvoyProcess := &os.Process{Pid: int(children[0].Pid)}
+	requireFindProcessNoError(t, fakeEnvoyProcess)
+
+	// Kill the fake func-e process.
+	// This works only for linux, sending kill -9 on darwin will not kill the process, we need to kill
```
codefromthecrypt, line 82: this is the sort of note that should move to the top of this function when calling t.Skip conditionally

codefromthecrypt (approved):
Looking good. some minor concerns when addressed, please merge!

A follow-up would be to matrix this test and show the behavior on TERM vs KILL. That can be a follow-up PR actually.

internal/moreos/moreos_func-e_test.go
codefromthecrypt: we can add this stuff back later, if we need it. For the sake of this change, we can focus on the most primary case of interest, which is signal propagation!

codefromthecrypt:
@mathetake if you have some other notes of interest, feel free to review also. Otherwise, with things already said LGTM!

internal/moreos/moreos_func-e_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	// fakeFuncESrc is a test source file used to simulate how func-e manages its child process
+	//go:embed testdata/fake_func-e.go
+	fakeFuncESrc []byte // Embedding the fakeFuncESrc is easier than file I/O and ensures it doesn't skew coverage
+
+	// Include the source imported by fakeFuncESrc directly and indirectly
+	//go:embed moreos.go
```
codefromthecrypt, line 36:
```suggestion
	//  TODO: replace wildcard with {{goos}} after https://github.com/golang/go/issues/48348
	//go:embed moreos.go
```

internal/moreos/moreos_func-e_linux_test.go
dio: OK, I chose to do the SIGKILL propagation.

internal/moreos/moreos_func-e_test.go
```diff
@@ -0,0 +1,143 @@
...
+	cmd := exec.Command(fakeFuncE, "run", arg, "-c")
+	cmd.Stdout = stdout
+	cmd.Stderr = stderr
+
+	require.NoError(t, cmd.Start())
+
+	// Block until we reach an expected line or timeout.
+	reader := bufio.NewReader(stderr)
+	waitFor := "initializing epoch 0"
+	if !assert.Eventually(t, func() bool {
```
mathetake, line 82: any reason why this is using assert instead of require?
```suggestion
	if !require.Eventually(t, func() bool {
```

internal/moreos/testdata/fake_func-e.go
```diff
@@ -0,0 +1,73 @@
+package main
```
mathetake, line 1: nit I would like to name this to `fake_func_e` following the Go convention 😄

internal/moreos/moreos_func-e_test.go
```diff
@@ -0,0 +1,143 @@
...
+		require.NoError(t, os.WriteFile(filepath.Join(moreosDir, src.Name()), data, 0600))
+	}
+
+	fakeFuncEBin := fakebinary.RequireBuildFakeBinary(t, workDir, "func-e", fakeFuncESrc)
+	require.NoError(t, os.WriteFile(path, fakeFuncEBin, 0700)) //nolint:gosec
+}
+
+func requireFindProcessError(t *testing.T, proc *os.Process, expectedErr error) {
+	// Wait until the operating system removes or adds the scheduled process.
+	if !assert.Eventually(t, func() bool {
```
mathetake, line 137: same here

mathetake (approved):

internal/moreos/testdata/fake_func-e.go
dio: I have a mixed feeling about this since `func-e` is a name. I will defer this to @codefromthecrypt 🙏🏽.

internal/moreos/moreos_func-e_test.go
dio: Ah yeah, missed this one. Seems like a copy-and-paste error. Thanks!

internal/moreos/moreos_func-e_test.go
dio: Updated.

internal/moreos/moreos_func-e_test.go
dio: Updated. Thanks!

mathetake (approved):

PR #372, opened 2021-09-13
Add tests calling terminating signals to func-e

dio:
This is a follow-up for https://github.com/tetratelabs/func-e/pull/371#pullrequestreview-752322738 to test a set of possible signals (`SIGINT`, `SIGTERM`, `SIGKILL`) to terminate func-e.

A minor change on checking the stderr: this PR uses a scanner instead of peeking on the buffer since for some reason redirecting `cmd.Stderr` to a buffer doesn't give complete stderr after the process is terminated. Hence here we use a scanner to `io.ReadCloser` returned by `cmd.StderrPipe()` instead.

### DIFF
--- a/internal/moreos/moreos_func-e_test.go
+++ b/internal/moreos/moreos_func-e_test.go
@@ -91,7 +91,7 @@ func Test_CallSignals(t *testing.T) {
 
 			tempDir := t.TempDir()
 
-			// Build a fake envoy and pass the ENV hint so that fake func-e uses it
+			// Build a fake envoy and pass the path via via ENVOY_PATH so that fake func-e uses it.
 			fakeEnvoy := filepath.Join(tempDir, "envoy"+Exe)
 			fakebinary.RequireFakeEnvoy(t, fakeEnvoy)
 			t.Setenv("ENVOY_PATH", fakeEnvoy)
@@ -128,6 +128,10 @@ func Test_CallSignals(t *testing.T) {
 
 			tc.signal(cmd.Process)
 			if tc.waitForExiting {
+				// When we decide to wait for the fake envoy for exiting, we wait for "exiting" message
+				// from envoy (see: internal/test/fakebinary/testdata/fake_envoy.go#82) after receiving
+				// interrupt from func-e (the fake func-e forwards fake envoy's stderr.
+				// See: internal/moreos/testdata/fake_func-e.go#38).
 				requireScannedWaitFor(t, stderrScanner, "exiting")
 			}
 
--- a/internal/moreos/moreos_func-e_test.go
+++ b/internal/moreos/moreos_func-e_test.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


dio:
Seems like adding more parallel tests give Windows a hard time?

internal/moreos/moreos_func-e_test.go
```diff
@@ -45,68 +46,106 @@ var (
...
-
-	// Wait and check if fake func-e and envoy processes are killed.
-	requireFindProcessError(t, cmd.Process, process.ErrorProcessNotRunning)
-	requireFindProcessError(t, fakeEnvoyProcess, process.ErrorProcessNotRunning)
+	tests := []testCase{
+		{
+			name:           "interrupt",
+			signal:         Interrupt,
+			waitForExiting: true,
+			skip:           runtime.GOOS == OSWindows,
```
dio, line 63: Seems like calling interrupt on CI to a running process gives Windows CI on GHA a hard time. exit code -1073741510. I'll spin up my local windows and see what I can do.

internal/moreos/testdata/fake_func-e.go
```diff
@@ -56,6 +56,10 @@ func main() {
 
 	// Block until we receive SIGINT or are canceled because Envoy has died.
 	<-sigCtx.Done()
+	stop()
+
+	// Simulate handleShutdown like in envoy.Run.
+	_ = moreos.Interrupt(cmd.Process)
```
dio, line 62: Probably this one causing -1073741510 on Windows? If we don't have this, we waste 5 secs on waiting the child process to terminate.

internal/moreos/moreos_func-e_test.go
```diff
@@ -45,68 +46,106 @@ var (
...
+			require.Equal(t, len(children), 1) // Should have only one child process i.e. the fake envoy process.
+			fakeEnvoyProcess := &os.Process{Pid: int(children[0].Pid)}
+
+			tc.signal(cmd.Process)
+			if tc.waitForExiting {
+				requireScannedWaitFor(t, stderrScanner, "exiting")
+			}
+
+			// Wait for the process to die; this could error due to the kill signal.
+			err = cmd.Wait()
```
dio, line 134: On windows, this `cmd.Wait` will exit when the test runner is finished. You can kill the child alone (cmd) here.

dio: Resolved. That was my fault for not setting the creation flag.

internal/moreos/testdata/fake_func-e.go
dio: This is because I didn't set the `cmd.SysProcAttr` for windows.

internal/moreos/testdata/fake_func-e.go
```diff
@@ -56,6 +56,10 @@ func main() {
 
 	// Block until we receive SIGINT or are canceled because Envoy has died.
 	<-sigCtx.Done()
+	stop()
```
mathetake, line 59: shouldn't this be called as `defer stop()` right after `signal.NotifyContext`?

internal/moreos/testdata/fake_func-e.go
```diff
@@ -39,7 +39,7 @@ func main() {
 
 	// Like envoy.Run.
 	waitCtx, waitCancel := context.WithCancel(context.Background())
-	sigCtx, _ := signal.NotifyContext(waitCtx, syscall.SIGINT, syscall.SIGTERM)
+	sigCtx, stop := signal.NotifyContext(waitCtx, os.Interrupt, syscall.SIGTERM)
 	defer waitCancel()
```
mathetake, line 43: nit I think we should call `defer waitCancel()` _before_ signal.NotifyContext :D

internal/moreos/moreos_func-e_test.go
```diff
@@ -45,68 +46,107 @@ var (
...
 
-	// With an arg so fakeFuncE runs fakeEnvoy as its child and doesn't exit.
-	arg := string(version.LastKnownEnvoy)
-	cmd := exec.Command(fakeFuncE, "run", arg, "-c")
-	cmd.Stdout = stdout
-	cmd.Stderr = stderr
+			tempDir := t.TempDir()
 
-	require.NoError(t, cmd.Start())
+			// Build a fake envoy and pass the ENV hint so that fake func-e uses it
```
mathetake, line 94: the more explicit the better 👯 
```suggestion
			// Build a fake envoy and pass it via ENVOY_PATH so that fake func-e uses it
```

mathetake (commented):
overall looks great!

internal/moreos/moreos_func-e_test.go
```diff
@@ -45,68 +46,107 @@ var (
...
+			require.NoError(t, err)
+			require.Equal(t, len(children), 1) // Should have only one child process i.e. the fake envoy process.
+			fakeEnvoyProcess := &os.Process{Pid: int(children[0].Pid)}
 
-	// Wait and check if fake func-e and envoy processes are killed.
-	requireFindProcessError(t, cmd.Process, process.ErrorProcessNotRunning)
-	requireFindProcessError(t, fakeEnvoyProcess, process.ErrorProcessNotRunning)
+			tc.signal(cmd.Process)
+			if tc.waitForExiting {
+				requireScannedWaitFor(t, stderrScanner, "exiting")
```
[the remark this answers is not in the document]
dio: Thanks. A good call. I added it.

internal/moreos/testdata/fake_func-e.go
dio: Thanks!

internal/moreos/moreos_func-e_test.go
dio: Sure thing. Updated.

internal/moreos/testdata/fake_func-e.go
dio: 👍🏽

mathetake (approved):
LGTM!

internal/moreos/moreos_func-e_test.go
```diff
@@ -136,3 +180,14 @@ func requireFindProcessError(t *testing.T, proc *os.Process, expectedErr error)
 		return err == expectedErr
 	}, 100*time.Millisecond, 5*time.Millisecond, "timeout waiting for expected error %v", expectedErr)
 }
+
+func requireScannedWaitFor(t *testing.T, s *bufio.Scanner, waitFor string) {
```
codefromthecrypt, line 184: later I will make a refactoring PR as we use scanner in the e2e for the same reasons, even if it could be documented better. Most notably we also avoid `require.Eventually` as it makes crap error messages when it fails. In other words, we used to use `require.Eventually`, that caused very difficult troubleshooting in CI, then we switched off of it.

Even if we fail to do refactoring for whatever reason, when there is copy/paste or something very similar, it is worth commenting the other function or file this is almost the same as. That way we can know to look at it in worst case. Folks who read those comments can help prevent adding a third similar thing later. It also helps reviewers who may have forgotten the other thing.

PR #373, opened 2021-09-19
WIP: Tag arm64 images

dio:
This builds and pushes arm64 flavor for internal images. This is to support the idea to run the following step when testing installing/uninstalling packages on `linux/arm64` (both for deb and rpm) in packaging workflow (please refer to: https://github.com/tetratelabs/func-e/pull/366).

```yaml
- uses: docker/setup-qemu-action@v1
- uses: docker://ghcr.io/tetratelabs/func-e-internal:ubuntu20.04-arm64
  with:
    args: packaging/nfpm/verify_deb.sh arm64
- uses: docker://ghcr.io/tetratelabs/func-e-internal:centos8-arm64
  with:
    args: packaging/nfpm/verify_rpm.sh aarch64

```

### DIFF
--- a/.github/workflows/internal-images.yml
+++ b/.github/workflows/internal-images.yml
@@ -66,8 +66,8 @@ jobs:
       # arm64 image by directly specifiying the tag without the need of supplying
       # `--platform linux/arm64`. Using `uses: docker://` only allows to override `args`, but not
       # the `--platform` flag.
-      # Reference: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-the-github-packages-container-registry.
-      # Related: https://github.com/actions/virtual-environments/issues/1368.
+      # Reference: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-the-github-packages-container-registry
+      # Related: https://github.com/actions/virtual-environments/issues/1368
       - name : Re-tag for arm64 only
         uses: docker/build-push-action@v2
         with:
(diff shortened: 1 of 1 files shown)


dio:
A running action as a test is on my repo here: https://github.com/dio/gha-test/runs/3644585431. Not sure how to test this in this repo with this workflow file: https://github.com/dio/gha-test/blob/random-images/.github/workflows/test.yaml.

codefromthecrypt (commented):
we don't need to test via emulation. you'll notice many comments to please use natural architectures in testing. testing rpm via travis is fine and I prefer to not have arm64 split across multiple build systems. Doing this here is confusing further the point and probably forgetting that we have a .travis.yaml for this. I would recommend citing this when publishing two platforms, that one consumer is GH actions and the other is travis (for arm64 linux which is the only consumer of that)

As this comment about using travis for arm64 has been repeated many times, I'd prefer to take over vs saying it again.

codefromthecrypt:
https://github.com/tetratelabs/func-e/pull/376 will close this

.github/workflows/internal-images.yml
```diff
@@ -43,10 +43,34 @@ jobs:
...
           EOF
-          docker build -t ghcr.io/tetratelabs/func-e-internal:${{ matrix.target_tag }} .
-          docker push ghcr.io/tetratelabs/func-e-internal:${{ matrix.target_tag }}
+
+      - name : Build and push
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64 # Build and push for amd64 and arm64 architectures.
+          tags:  ghcr.io/${{ github.repository_owner }}/func-e-internal:${{ matrix.target-tag }}
```
codefromthecrypt, line 63: fyi typo: target_tag not target-tag

my guess is that if "push: true" was here it would have failed for you.

codefromthecrypt:
obviated by #376

PR #375, opened 2021-09-24
Makes fake_func-e more consistent with real func-e

codefromthecrypt:
This addresses some fuzz to make sure the comments don't lie. Notably,
this defers contexts the same way between fake and real. This also pulls
out handleShutdown to accent notes from #372

### DIFF
--- a/internal/envoy/run.go
+++ b/internal/envoy/run.go
@@ -17,7 +17,6 @@ package envoy
 import (
 	"context"
 	"fmt"
-	"io"
 	"os"
 	"os/exec"
 	"os/signal"
@@ -50,8 +49,7 @@ func (r *Runtime) Run(ctx context.Context, args []string) error {
 	}
 
 	// Warn, but don't fail if we can't write the pid file for some reason
-	err := os.WriteFile(r.pidPath, []byte(strconv.Itoa(cmd.Process.Pid)), 0600)
-	maybeWarn(r.Out, err)
+	r.maybeWarn(os.WriteFile(r.pidPath, []byte(strconv.Itoa(cmd.Process.Pid)), 0600))
 
 	waitCtx, waitCancel := context.WithCancel(ctx)
 	defer waitCancel()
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
In some ways this is ugly.. that fake func-e is longer. however, this is more honest. It helps showcase some potential to refactor the process state model, which afterwards could pay down the visible debt here.

internal/envoy/run.go
```diff
@@ -64,8 +64,10 @@ func (r *Runtime) Run(ctx context.Context, args []string) (err error) { //nolint
 	}
 
 	waitCtx, waitCancel := context.WithCancel(ctx)
-	sigCtx, sigCancel := signal.NotifyContext(waitCtx, syscall.SIGINT, syscall.SIGTERM)
 	defer waitCancel()
+
+	sigCtx, sigCancel := signal.NotifyContext(waitCtx, syscall.SIGINT, syscall.SIGTERM)
+	defer sigCancel()
```
mathetake, line 70: I thought calling CancelFunc multiple times would result in panics but I confirmed it is not. Thanks

> // After the first call, subsequent calls to a CancelFunc do nothing.

internal/moreos/testdata/fake_func-e.go
```diff
@@ -31,23 +32,41 @@ func main() {
...
+	cmd := exec.Command(os.Getenv("ENVOY_PATH"), args...)
 	cmd.SysProcAttr = moreos.ProcessGroupAttr()
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
-	// Like envoy.Run.
-	waitCtx, waitCancel := context.WithCancel(context.Background())
+	// Suppress any error and replace it with the envoy exit status when > 1
+	defer func() {
+		if cmd.ProcessState != nil && cmd.ProcessState.ExitCode() > 0 {
```
mathetake, line 53: ExitCode() is safe to call even when ProcessState == nil and in that case it returns `-1`
```suggestion
		if cmd.ProcessState.ExitCode() > 0 {
```

internal/moreos/testdata/fake_func-e.go
```diff
@@ -31,23 +32,41 @@ func main() {
...
+			if err != nil {
```
mathetake, line 54: it seems really difficult to me to figure out where this "err" becomes not nil - I can't find `err = *` anywhere except two lines below.

codefromthecrypt: indeed and the good thing is now you know the opportunity to improve real func-e :D it is the same logic from there.

internal/envoy/run.go
codefromthecrypt: thanks for researching this!

internal/moreos/testdata/fake_func-e.go
codefromthecrypt: wow sneaky nil goodness!

```Golang
// ExitCode returns the exit code of the exited process, or -1
// if the process hasn't exited or was terminated by a signal.
func (p *ProcessState) ExitCode() int {
	// return -1 if the process hasn't started.
	if p == nil {
		return -1
	}
	return p.status.ExitStatus()
}
```

codefromthecrypt: fixed this in read func-e also

internal/moreos/testdata/fake_func-e.go
codefromthecrypt: PTAL I refactored out this part, and luckily we no longer have to skip lint due to cyclomatic complexity

codefromthecrypt:
Thanks for the help @mathetake. in the near future I hope to reduce the amount of code here, too.

PR #376, opened 2021-09-24
Pushes internal arm64 CentOS image and adds RATIONALE.md about Travis

codefromthecrypt:
A lot of thought went into the decision of whether or not to use
emulation for arm64. However, this was spread out across discussions and
PRs, but not in the codebase. This adds a RATIONALE section with the
background, which is that this is temporary until GHA supports arm64.

Note, the same would apply if we start supporting s390x, except we don't
at the moment, which is why that isn't mentioned. s390x support would
need to happen first in Envoy.

Obviates #373

### DIFF
--- a/.github/workflows/internal-images.yml
+++ b/.github/workflows/internal-images.yml
@@ -30,6 +30,19 @@ jobs:
             target_tag: centos8
     runs-on: ubuntu-latest
     steps:
+      # Same as doing this locally: echo "${GHCR_TOKEN}" | docker login ghcr.io -u "${GHCR_TOKEN}" --password-stdin
+      - name: Login into GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          # GHCR_TOKEN=<hex token value>
+          #   - pushes Docker images to ghcr.io
+          #   - create via https://github.com/settings/tokens
+          #   - assign via https://github.com/organizations/tetratelabs/settings/secrets/actions
+          #   - needs repo:status, public_repo, write:packages, delete:packages
+          password: ${{ secrets.GHCR_TOKEN }}
+
       # We need QEMU and Buildx for multi-platform (amd64+arm64) image push. See RATIONALE.md
       - name: Setup QEMU
         uses: docker/setup-qemu-action@v1
@@ -50,11 +63,3 @@ jobs:
           context: .
           platforms: linux/amd64,linux/arm64  # arm64 is run only by Travis. See RATIONALE.md
           tags: ghcr.io/${{ github.repository_owner }}/func-e-internal:${{ matrix.target_tag }}
-          # Same as doing this locally: echo "${GHCR_TOKEN}" | docker login ghcr.io -u "${GHCR_TOKEN}" --password-stdin
-          # GHCR_TOKEN=<hex token value>
-          #   - pushes Docker images to ghcr.io
-          #   - create via https://github.com/settings/tokens
-          #   - assign via https://github.com/organizations/tetratelabs/settings/secrets/actions
-          #   - needs repo:status, public_repo, write:packages, delete:packages
-          secrets: |
-            GIT_AUTH_TOKEN=${{ secrets.GHCR_TOKEN }}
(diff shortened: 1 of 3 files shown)


RATIONALE.md
```diff
@@ -0,0 +1,25 @@
+# Notable rationale of func-e
+
+## Why do we use Travis and GitHub Actions instead of only GitHub Actions?
+We use Travis to run CentOS integration tests on arm64 until [GitHub Actions supports it](https://github.com/actions/virtual-environments/issues/2552).
+
+This is an alternative to using emulation instead. Using emulation (ex via `setup-qemu-action` and
+`setup-buildx-action`) re-introduces problems we eliminated with Travis. For example, not would
```
codefromthecrypt, line 7:
```suggestion
`setup-buildx-action`) re-introduces problems we eliminated with Travis. For example, not only would
```

RATIONALE.md
```diff
@@ -0,0 +1,25 @@
...
+we broke our [Makefile](Makefile) or that there's an unknown new dependency of Envoy® (such as a
+change to the floor version of glibc). While these are unlikely to occur, running tests are still
+important.
+
+The only place we use emulation is [publishing internal images](.github/workflows/internal-images.yml).
+This is done for convenience and occurs once per day, so duration and rate limiting effects are
+not important. If something breaks here, the existing images would become stale, so it isn't as
+much an emergency as if we put emulation in the critical path (ex in PR tests.)
+
+At the point when GitHub Actions supports free arm64 images, we can simplify by removing Travis.
```
codefromthecrypt, line 24:
```suggestion
At the point when GitHub Actions supports free arm64 runners, we can simplify by removing Travis.
```

codefromthecrypt:
tested with another github repo, pushing and consuming the image in the same way as we do here. Actually having travis use the other image would be a separate change.

codefromthecrypt:
Note: while I could list more reasons to not default to emulation, such as having to explain convoluted workarounds to actions that don't support it, I think there are enough reasons listed in the rationale for now. If we need more, definitely there's an advantage to being able to *not* do or explain this:

````yaml
      # We do re-tagging for easier usage when testing using emulation, so we can forcefuly pull
      # arm64 image by directly specifiying the tag without the need of supplying
      # `--platform linux/arm64`. Using `uses: docker://` only allows to override `args`, but not
      # the `--platform` flag.
      # Reference: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-the-github-packages-container-registry
      # Related: https://github.com/actions/virtual-environments/issues/1368
      - name : Re-tag for arm64 only
        uses: docker/build-push-action@v2
        with:
          context: .
          platforms: linux/arm64 # Build and push for arm64 architecture only.
          tags:  ghcr.io/${{ github.repository_owner }}/func-e-internal:${{ matrix.target-tag }}-arm64
````

**Comment by codefromthecrypt** [comment]
Another reason I didn't mention explicitly is the doubt emulation causes about the *real* reason why something fails. For example, in zipkin emulation bugs resulted in s390x to not work regardless of if it did in native arch. The complexity of layering several tools to achieve an arch means you have to consider in the case of func-e.. is this failing because envoy? or envoy over docker? or envoy over docker with emulation? Reducing the amount of things to wonder about when things fail is what you get when you move from complex -> complicated. I'll choose complicated (2 builds, one you barely touch) vs complex (config compat bomb waiting and only testable in a github actions run)

File: .github/workflows/internal-images.yml [hunk 3]
```diff
@@ -43,10 +43,23 @@ jobs:
...
+      - name: Write Dockerfile
+        run: |
           cat > Dockerfile <<'EOF'
           ${{ matrix.dockerfile }}
           EOF
-          docker build -t ghcr.io/tetratelabs/func-e-internal:${{ matrix.target_tag }} .
-          docker push ghcr.io/tetratelabs/func-e-internal:${{ matrix.target_tag }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v2
```
codefromthecrypt, line 60: ps this is more opaque than I'd like as it is really hard to test what this might do locally (presumably buildx commands). Meanwhile we can use the action until it proves itself a nuisance (which might not happen ever if we are lucky!)

mathetake (approved):

codefromthecrypt:
I will kick off the workflow manually and deal with any fallout

PR #379, opened 2021-09-27
Removes go version maintenance

codefromthecrypt:
This removes the hard requirement to build with exactly the same version
of go as indicated in the go.mod file. This allows more flexibility when
new versions of go (such as 1.18) are released.

This also removes the requirement to track down to the patch of go.
Instead, this builds with the latest patch of go that's available at the
time.

Go versions aren't often released, but this removes maintenance each
time that happens. It also allows us to more easily cache golang
installs in our Docker image, and without tight coupling to what GitHub
actions uses.

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -4,8 +4,8 @@
 ---
 name: "build"
 
-env:  # As of 2021-08-05 images come with go 1.15 https://github.com/actions/virtual-environments/tree/main/images
-  GO_VERSION: "1.17.0"
+env:  # Update this prior to requiring a higher minor version in go.mod
+  GO_VERSION: "1.17"  # Latest patch
 
 on:
   push:  # We run tests on non-tagged pushes to master
@@ -123,7 +123,7 @@ jobs:
 
       - name: "Install Go"
         uses: actions/setup-go@v2
-        with:
+        with:  # TODO: refactor this as the image will already have go installed
           go-version: ${{ env.GO_VERSION }}
 
       - name: "Cache golang"
--- a/.github/workflows/internal-images.yml
+++ b/.github/workflows/internal-images.yml
@@ -28,6 +28,10 @@ jobs:
         include:
           # This installs the dependencies needed to build func-e including the latest patch of go
           # CentOS' root user shell is /bin/bash so go env is set via ~/.bashrc
+          #
+          # Note: This uses gimme so that we can install the latest patch of go without knowing what
+          # that is, or parsing it on our own. This is more flexible and removes update maintenance
+          # we'd have if we hard coded a specific patch. See https://github.com/travis-ci/gimme
           - dockerfile: |
               FROM centos:8
               RUN yum install -y --quiet make which git gcc && yum clean all && \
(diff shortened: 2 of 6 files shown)


.github/workflows/commit.yaml
```diff
@@ -123,7 +123,7 @@ jobs:
 
       - name: "Install Go"
         uses: actions/setup-go@v2
-        with:
+        with:  # TODO: refactor this as the image will already have go installed
```
codefromthecrypt, line 126: This has to happen after the docker image is republished to include go

.github/workflows/internal-images.yml
```diff
@@ -24,9 +26,13 @@ jobs:
...
         # Be precise in tag versions to improve reproducibility
         include:
-          - dockerfile: |  # Don't install golang as the version often changes
+          # This installs the dependencies needed to build func-e including the latest patch of go
+          - dockerfile: |
               FROM centos:8
-              RUN yum install -y --quiet make which git gcc && yum clean all
+              RUN yum install -y --quiet make which git gcc && yum clean all && \
+                  curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | GIMME_GO_VERSION=${{ env.GO_VERSION }}.x bash && \
+                  source ~/.gimme/envs/latest.env && \
```
codefromthecrypt, line 34: I think this only applies to the current RUN directive and won't effect the path of the next, so maybe this isn't ready for merge, yet!

.github/workflows/internal-images.yml
```diff
@@ -24,9 +26,13 @@ jobs:
       matrix:
         # Be precise in tag versions to improve reproducibility
         include:
-          - dockerfile: |  # Don't install golang as the version often changes
+          # This installs the dependencies needed to build func-e including the latest patch of go
+          - dockerfile: |
               FROM centos:8
-              RUN yum install -y --quiet make which git gcc && yum clean all
+              RUN yum install -y --quiet make which git gcc && yum clean all && \
+                  curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | GIMME_GO_VERSION=${{ env.GO_VERSION }}.x bash && \
```
mathetake, line 33: oh why do we need to travis-ci/gimme and not follow the official installation? https://golang.org/doc/install#install

mathetake: then we don't need to install these deps ^^ I guess.

codefromthecrypt: gimme automates the go install which allows you to use latest patch. usually the rpm repos are out of date and not latest patch. if we download directly from go we have to know precise patch version again..

mathetake: hmm ic. I think we should at least put the comment about that in yaml!

codefromthecrypt: good idea

codefromthecrypt:
Thanks for the help making this better

.github/workflows/internal-images.yml
codefromthecrypt: https://github.com/tetratelabs/func-e/pull/382 found a way to not need to externally lookup the version, as we can look at the runner and re-use that!

PR #382, opened 2021-09-30
Matches minor revision of go of CentOS image with GHA runner

codefromthecrypt:
This matches the minor revision of go with the current GitHub Actions
runner instead of looking up latest using gimme. Doing so removes
complexity around what gimme is, and why figuring out the latest go
revision for a minor isn't straight-forward [1].

[1] Unlike a lot of projects, go does not attach binaries to releases.
This makes a potential race between a tag being pushed and binaries
being available. Moreover, there are release tags for release
canditates, so filtering these out requires logic. The way to get the
best known latest revision is to parse https://golang.org/dl, as that
is the most official source of what is the latest minor. However,
parsing that is tricky, so if we did, we might as well re-use gimme.

Ex.
```yaml
        run: |
          GIMME_SCRIPT=https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
          GO_MINOR_REVISION=$(curl -sSL ${GIMME_SCRIPT} | GIMME_GO_VERSION=${{ env.GO_VERSION }}.x bash -s -- -r)
```

The above is for reference only because after this change, we don't
lookup a version of go anymore.

### DIFF
--- a/.github/workflows/internal-images.yml
+++ b/.github/workflows/internal-images.yml
@@ -8,8 +8,11 @@ on:
     - cron: "23 3 * * *"
   workflow_dispatch:  # Allows manual refresh
 
-env:  # Update this prior to requiring a higher minor version in go.mod
-  GO_VERSION: "1.17"  # Latest patch
+# GOROOT_NAME resolves a GOROOT path corresponding to a major Golang release on the current runner.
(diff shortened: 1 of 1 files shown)


mathetake (approved):
Brilliant! 💯

PR #383, opened 2021-10-01
Avoids repetition by testing CentOS after Ubuntu

codefromthecrypt:
This is less copy/pasta

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -87,6 +87,13 @@ jobs:
       - name: "Run e2e tests using the `func-e` binary"
         run: make e2e
 
+      - name: "Run e2e tests using the `func-e` binary (CentOS)"
+        if: runner.os == 'Linux'
+        run: docker run -ti -v $PWD:/func-e --rm ${CENTOS_IMAGE} -c "cd /func-e; make -o ${E2E_FUNC_E_PATH} e2e"
+        env:  # CENTOS_IMAGE was built by internal-images.yaml; E2E_FUNC_E_PATH was built via `make build`
+          CENTOS_IMAGE: ghcr.io/tetratelabs/func-e-internal:centos8
+          E2E_FUNC_E_PATH: build/func-e_linux_amd64/func-e
+
       - name: "Generate coverage report"  # only once (not per OS)
         if: runner.os == 'Linux'
         run: make coverage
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
hmm 45s I could go either way on this. The pro is less config replicated, the con is the long pole is longer (3m until PR feedback completes)

.github/workflows/commit.yaml
```diff
@@ -87,6 +87,13 @@ jobs:
       - name: "Run e2e tests using the `func-e` binary"
         run: make e2e
 
+      - name: "Run e2e tests using the `func-e` binary (CentOS)"
+        if: runner.os == 'Linux'
+        run: docker run -v $PWD:/func-e --rm ${CENTOS_IMAGE} -c "cd /func-e; make -o ${E2E_FUNC_E_PATH} e2e"
```
codefromthecrypt, line 92: maybe I can volume mount ~/.func-e/versions to reduce execution time...

codefromthecrypt:
38s longer.. 15s is downloading the centos image. not sure we can make this much better.

@mathetake WDYT? should we do this and be 38s longer or leave it and be 35 lines longer?

codefromthecrypt:
actually total duration isn't longer as windows is slower than the now slower ubuntu :D

.github/workflows/commit.yaml
```diff
@@ -87,6 +87,13 @@ jobs:
       - name: "Run e2e tests using the `func-e` binary"
         run: make e2e
 
+      - name: "Run e2e tests using the `func-e` binary (CentOS)"
+        if: runner.os == 'Linux'
+        run: docker run -v $HOME/.func-e:/root/.func-e -v $PWD:/func-e --rm ${CENTOS_IMAGE} -c "cd /func-e; make -o ${E2E_FUNC_E_PATH} e2e"
```
mathetake, line 92: I thought we could `use: docker://ghcr.io/...` to avoid `docker run -v manual mounts etc...`?

mathetake (commented):
generally I like this better  than tens of lines longer :D

.github/workflows/commit.yaml
mathetake: oh that would result in the same maybe so nvm 😄

mathetake (approved):

.github/workflows/commit.yaml
codefromthecrypt: yep plus, you can paste this into your laptop and it probably works :D A follow-up change will use the same to test centos+arm64

codefromthecrypt:
thanks for thinking about it @mathetake

codefromthecrypt:
master now does the same in travis, running e2e back to back ubuntu then centos, using arm64graviton and done in <2.5minutes

```diff
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ cache:
   directories: # ~/.func-e/versions is cached so that we only re-download once: for TestFuncEInstall
     - $HOME/.func-e/versions
     - $HOME/go/pkg/mod
-    - $HOME/go/bin/*-v*
 
 git:
   depth: false  # TRAVIS_COMMIT_RANGE requires full commit history.
@@ -28,4 +27,13 @@ before_install: |  # Prevent test build of a documentation or GitHub Actions onl
   fi
   make check || travis_terminate 1
 
-script: make e2e
+env:  # CENTOS_IMAGE was built by internal-images.yaml; E2E_FUNC_E_PATH was built via `make e2e`
+  global:
+    - CENTOS_IMAGE=ghcr.io/tetratelabs/func-e-internal:centos8
+    - E2E_FUNC_E_PATH=build/func-e_linux_arm64/func-e
+
+script:
+  # Run e2e tests using the `func-e` binary (Ubuntu)
+  - make e2e
+  # Run e2e tests using the `func-e` binary (CentOS)
+  - docker run -v $HOME/.func-e:/root/.func-e -v $PWD:/func-e --rm ${CENTOS_IMAGE} -c "cd /func-e; make -o ${E2E_FUNC_E_PATH} e2e"
```

PR #385, opened 2021-10-01
Ensures Ubuntu runs the same version of go as CentOS

codefromthecrypt:
This avoids any version differences between running e2e tests via Docker
(ex for CentOS) vs a normal run (Ubuntu, MacOS, etc). It accomplishes
this by using the built-in GOROOT variable.

go.mod controls which GOROOT is used, and in doing so, this removes
maintenance.

This includes a temporary workaround because MacOS does not have the
same variable as Windows and Ubuntu. Once the below is solved, we can
simply delete the workaround.

https://github.com/actions/virtual-environments/issues/4156

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -4,12 +4,6 @@
 ---
 name: "build"
 
-# GOROOT_NAME resolves a GOROOT path corresponding to a major Golang release on the current runner.
-# Ex. Given GOROOT_NAME=GOROOT_1_17_X64, ${!GOROOT_NAME}=/opt/hostedtoolcache/go/1.17.1/x64
-# See https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#environment-variables-2
-env:
-  GOROOT_NAME: GOROOT_1_17_X64
-
 on:
   push:  # We run tests on non-tagged pushes to master
     tags: ''
@@ -57,32 +51,17 @@ jobs:
       - name: "Checkout"
         uses: actions/checkout@v2
 
-      # This sets the version-specific GOROOT env on macOS until actions/virtual-environments#4156
-      #
-      # This must happen before interpolating ${!GOROOT_NAME} on macOS
-      - name: "Set GOROOT_X64 env (macOS)"
-        if: runner.os == 'macOS'
-        run: |  # ex GOROOT_1_17_X64 -> 1.17
-          go_version=$(echo ${GOROOT_NAME//_/.}| cut -d. -f2,3)
-          go_root=$(ls -d /Users/runner/hostedtoolcache/go/${go_version}*/x64|sort -n|tail -1)
-          echo "${GOROOT_NAME}=${go_root}" >> $GITHUB_ENV
-
       - name: "Configure Go"
-        run: |  # This exports variables needed to run and cache a GOROOT set with ${!GOROOT_NAME}.
-          go_path=${!GOROOT_NAME}/bin
-          for var in GOROOT GOCACHE GOMODCACHE; do
-            echo ${var}=$(${go_path}/go env ${var}) >> $GITHUB_ENV
-          done
-          echo ${go_path} >> $GITHUB_PATH
+        run: .github/workflows/configure_go.sh
 
       - name: "Cache Go"
         uses: actions/cache@v2
         with:
-          path: |  # Cache downloaded go modules and the build cache (Makefile involves `go run`)
+          path: |  # Cache downloaded Go modules and the GOCACHE (to cache `go run` builds invoked by Makefile)
+            ~/go/pkg/mod
             ${{ env.GOCACHE }}
-            ${{ env.GOMODCACHE }}
-          key: test-${{ runner.os }}-${{ env.GOROOT_NAME }}-go-${{ hashFiles('go.sum', 'Makefile') }}
-          restore-keys: test-${{ runner.os }}-${{ env.GOROOT_NAME }}-go-
+          key: test-${{ runner.os }}-${{ env.GO_VERSION }}-go-${{ hashFiles('go.sum', 'Makefile') }}
+          restore-keys: test-${{ runner.os }}-${{ env.GO_VERSION }}-go-
 
       - name: "Cache Envoy binaries"
         uses: actions/cache@v2
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -4,8 +4,11 @@
 ---
 name: "build"
 
-env:  # Update this prior to requiring a higher minor version in go.mod
-  GO_VERSION: "1.17"  # Latest patch
+# GOROOT_NAME resolves a GOROOT path corresponding to a major Golang release on the current runner.
+# Ex. Given GOROOT_NAME=GOROOT_1_17_X64, ${!GOROOT_NAME}=/opt/hostedtoolcache/go/1.17.1/x64
+# See https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#environment-variables-2
+env:
+  GOROOT_NAME: GOROOT_1_17_X64
 
 on:
   push:  # We run tests on non-tagged pushes to master
@@ -54,19 +57,32 @@ jobs:
       - name: "Checkout"
         uses: actions/checkout@v2
 
-      - name: "Install Go"
-        uses: actions/setup-go@v2
-        with:
-          go-version: ${{ env.GO_VERSION }}
+      # This sets the version-specific GOROOT env on macOS until actions/virtual-environments#4156
+      #
+      # This must happen before interpolating ${!GOROOT_NAME} on macOS
+      - name: "Set GOROOT_X64 env (macOS)"
+        if: runner.os == 'macOS'
+        run: |  # ex GOROOT_1_17_X64 -> 1.17
+          go_version=$(echo ${GOROOT_NAME//_/.}| cut -d. -f2,3)
+          go_root=$(ls -d /Users/runner/hostedtoolcache/go/${go_version}*/x64|sort -n|tail -1)
+          echo "${GOROOT_NAME}=${go_root}" >> $GITHUB_ENV
 
-      - name: "Cache golang"
+      - name: "Configure Go"
+        run: |  # This exports variables needed to run and cache a GOROOT set with ${!GOROOT_NAME}.
+          go_path=${!GOROOT_NAME}/bin
+          for var in GOROOT GOCACHE GOMODCACHE; do
+            echo ${var}=$(${go_path}/go env ${var}) >> $GITHUB_ENV
+          done
+          echo ${go_path} >> $GITHUB_PATH
+
+      - name: "Cache Go"
         uses: actions/cache@v2
         with:
-          path: |  # TODO: go build cache if we care, noting it is OS-specific
-            ~/go/pkg/mod
-          # Makefile contains go run statements which affect the build cache
-          key: test-${{ runner.os }}-${{ env.GO_VERSION }}-go-${{ hashFiles('go.sum', 'Makefile') }}
-          restore-keys: test-${{ runner.os }}-${{ env.GO_VERSION }}-go-
+          path: |  # Cache downloaded go modules and the build cache (Makefile involves `go run`)
+            ${{ env.GOCACHE }}
+            ${{ env.GOMODCACHE }}
+          key: test-${{ runner.os }}-${{ env.GOROOT_NAME }}-go-${{ hashFiles('go.sum', 'Makefile') }}
+          restore-keys: test-${{ runner.os }}-${{ env.GOROOT_NAME }}-go-
 
       - name: "Cache Envoy binaries"
         uses: actions/cache@v2
@@ -104,7 +120,7 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: bash <(curl -s https://codecov.io/bash)
 
-      - name: "Run unit tests in a different timezone" # https://github.com/tetratelabs/func-e/issues/303.
+      - name: "Run unit tests in a different timezone"  # https://github.com/tetratelabs/func-e/issues/303.
         if: runner.os == 'Linux'
         run: |
           sudo timedatectl set-timezone America/Phoenix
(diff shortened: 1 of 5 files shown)


.github/workflows/commit.yaml
```diff
@@ -4,8 +4,11 @@
 ---
 name: "build"
 
-env:  # Update this prior to requiring a higher minor version in go.mod
-  GO_VERSION: "1.17"  # Latest patch
+# GOROOT_NAME resolves a GOROOT path corresponding to a major Golang release on the current runner.
+# Ex. Given GOROOT_NAME=GOROOT_1_17_X64, ${!GOROOT_NAME}=/opt/hostedtoolcache/go/1.17.1/x64
+# See https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#environment-variables-2
+env:
+  GOROOT_NAME: GOROOT_1_17_X64
```
codefromthecrypt, line 11: so now, everything that talks about go, talks about it in terms of GitHub Actions ENV variables, regardless of if it is building a docker image (prior image) or running go normally.

While not obvious, runners default to different go.. so we still have to do this because runners default to 1.15 even if 1.17 (specifically currently 1.17.1) is installed across the board.

.github/workflows/commit.yaml
```diff
@@ -54,19 +57,32 @@ jobs:
...
+          go_path=${!GOROOT_NAME}/bin
+          for var in GOROOT GOCACHE GOMODCACHE; do
+            echo ${var}=$(${go_path}/go env ${var}) >> $GITHUB_ENV
+          done
+          echo ${go_path} >> $GITHUB_PATH
+
+      - name: "Cache Go"
         uses: actions/cache@v2
         with:
-          path: |  # TODO: go build cache if we care, noting it is OS-specific
```
codefromthecrypt, line 65: look mom, no TODO!

.github/workflows/commit.yaml
```diff
@@ -4,8 +4,11 @@
 ---
 name: "build"
 
-env:  # Update this prior to requiring a higher minor version in go.mod
-  GO_VERSION: "1.17"  # Latest patch
+# GOROOT_NAME resolves a GOROOT path corresponding to a major Golang release on the current runner.
+# Ex. Given GOROOT_NAME=GOROOT_1_17_X64, ${!GOROOT_NAME}=/opt/hostedtoolcache/go/1.17.1/x64
+# See https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#environment-variables-2
+env:
```
codefromthecrypt, line 10: It occurs to me that now that we use shell to setup go, we can use that to parse go.mod to get the correct version (as @mathetake suggested in a different PR). This means we choose the runtime conflated with the minimum go version, but IMHO that's ok. The other thing this means is we must checkout the source prior to setting up go, but there's only one case where we don't already do that.

Basically I think we can remove the need to set GOROOT_NAME=GOROOT_1_17_X64 as long as we are ok conflating the runtime version with the version declared in go.mod

codefromthecrypt: making a general PR comment about this problem and a way out

codefromthecrypt:
# Lockup issues when using go.mod to drive the go tooling version
The main issues in locking the go tooling version exactly with go.mod are:
  * if eventually we are consumed as an api, we could be "too current" for a potential consumer
  * if releasing from a branch, you have to consider if the potentially older version of go in go.mod will be available.
  * pull requests to update go.mod assume the go minor version requested is already there

## consumers not wanting very latest go
Currently, we aren't being used as an api, and also AFAIK go doesn't enforce the go.mod version (ex at least you used to be able to build go 1.16 go.mod with go 1.15 unless you used library changes that depended on 1.16). So, locking might be ok.

## branch release
We currently aren't releasing from branches, so we don't have this problem. If we started to do that, likely it would only effect the centos runner (e2e) as that's the only thing not using the GHA runner image (which has several versions of go on it).

## pull request to upgrade go problem
Similar to branch release problem, a pull request to change go.mod becomes more complex if it is implicitly the same version of go used to build the project.

GHA runners are ok, the only issue is that we may have to wait 1-2 weeks top upgrade to latest go minor. For example, the GHA runners last update was about 8 days later than when go was published. That's easy to reason with.

The centos runner is currently more simple, as it only has 1 goroot. When the GOROOT is implicit and changes in a PR, that image won't have it yet, and the build will fail. Since publishing the image uses privileges this would be tricky to unlock

# Options to mitigate go tooling lockup
Assuming we are not going to worry about being "too new" for consumers (ex go.mod's version of go is later than theirs), we are mainly left with lockup caused by our centos image not having as many go roots as GHA runners. There are a couple ways out of that
 
* go back to dynamically installing go
* add 2 goroots to the centos image

## dynamically installing go
dynamically installing go adds back the issue of skew between when GHA update their virtual environments and latest release of go. Basically, the build will take longer for 8 days or so while actions/setup-go has to install a goroot because GHA's is slightly out of date (ex 1.17.2 is out, but 1.17.1 is on the host). This also reintroduces a chance of skew between the centos image and the GHA runner, as we'd have to heuristically figure out what actions/setup-go does to get the latest version, or run that action prior to building the centos image. In any case, there is more labor and build time involved as we move back from matching the virtual environment version of go to matching the output of actions/setup-go. For example, this would move us back to running gimme on-demand on centos just to make sure PRs can update go.

## Change our CentOS image to install the last 2 goroots from the GHA runner
Another way out is to not have only one goroot on CentOS. Rather, have two. In GHA runners, last two would be `GOROOT_1_17_X64` and `GOROOT_1_16_X64`. In our current setup, the latter would be redundant. However, this helps when both bump a week or two after 1.18 is released.

To emulate that in CentOS, we probably don't want to carry the `_X64` suffix because that's not multi-architecture friendly, and we wouldn't intentionally install two archs on a image that can only run as one. Rather, we could do `GOROOT_1_17` and `GOROOT_1_16`. We could change our dynamic go selection accordingly and choose the variable that has a prefix `GOROOT_${go_mod_version//./_}`

After this occurs, updating go because easier to reason with. The day *after* https://github.com/actions/virtual-environments adds a minor version of go, we can make it default. The day *after* part allows for the scheduled build to kick off. We could tighten this further if GHA ever lets you trap on changes to a file in another repo, but day after should be fine meanwhile.

# Recommendation
Personally, I don't mind the update several `GOROOT_NAME` variables problem as exists today. However, it does require one to remember to update these *before* updating go.mod to use a newer version of go. The two steps would be annoying although infrequent as go doesn't release that often.

OTOH, I think that it is overall more automated and less to think about if we automatically sync last 2 goroots from the current GHA runner when making our centos image. The centos image would become a little more complex, but that complexity would only be visible when someone chooses to look at how it is built.

So, I'm personally OK either way, but feel like it is less bombs for future maintainers to go ahead and install 2 goroots on the centos image and select them with variables similar to how GHA runners do.

@mathetake wdyt?

codefromthecrypt:
regardless of what we decide here, I'll update RATIONALE.md as it takes too much thinking (and sometimes a shower) to remember all the pros and cons here :D

.github/workflows/commit.yaml
```diff
@@ -54,19 +57,32 @@ jobs:
...
+          for var in GOROOT GOCACHE GOMODCACHE; do
+            echo ${var}=$(${go_path}/go env ${var}) >> $GITHUB_ENV
+          done
+          echo ${go_path} >> $GITHUB_PATH
+
+      - name: "Cache Go"
         uses: actions/cache@v2
         with:
-          path: |  # TODO: go build cache if we care, noting it is OS-specific
-            ~/go/pkg/mod
```
codefromthecrypt, line 66: TODO: It is really only GOCACHE that is OS-dependent. GHA runners don't set any GO related env variables.

So the main thing is that GOCACHE is sadly different on literally OS we use (as defined by go itself)...
* linux - $HOME/.cache/go-build
* macOS - $HOME/Library/Caches/go-build
* windows - %HOME%\AppData\Local\go-build

while GOMODCACHE is luckily always the same ~/go/pkg/mod

codefromthecrypt:
OK I rewrote this to choose the version based on go.mod. Another change would be needed to unlock this, allowing PRs to upgrade go without breaking. We can do that in a follow-up PR, if the current impl looks good. I felt it was probably easier to show vs tell, on how lookup via go.mod would work, so here it is!

.github/workflows/internal-images.yml
```diff
@@ -48,7 +42,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       # Same as doing this locally: echo "${GHCR_TOKEN}" | docker login ghcr.io -u "${GHCR_TOKEN}" --password-stdin
-      - name: Login into GitHub Container Registry
+      - name: "Login into GitHub Container Registry"
```
codefromthecrypt, line 45: this was the only file not quoting the name, so made it like the others.

.github/workflows/internal-images.yml
```diff
@@ -61,28 +55,30 @@ jobs:
...
-        run: |
-          rev=$(echo ${!GOROOT_NAME}|cut -d/ -f5)
+      - name: "Checkout"
+        uses: actions/checkout@v2
+
+      - name: "Configure Go"
+        run: .github/workflows/configure_go.sh
+
+      - name: "Export Go minor revision"
+        run: |  # Ex. GOROOT=/opt/hostedtoolcache/go/1.17.1/x64 -> GO_MINOR_REVISION=1.17.1
```
codefromthecrypt, line 71: so this is the part that syncs the docker build with the runner's version of go. In a later PR we would refactor this to address the "upgrade go in one PR" problem. However, doing that now would make this change too complex.

.github/workflows/configure_go.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
codefromthecrypt, line 1: net 20 new lines are mostly comments in this file, but also this does more (ex fixing the go build cache), so hopefully ok

codefromthecrypt:
So, to recap, the main goal is docker consistency, especially go env, between [GitHub actions runner](https://github.com/actions/runner/releases/tag/v2.283.2) + [GitHub Ubuntu virtual environment](https://github.com/actions/virtual-environments/). In other words, we mostly just think about GitHub ubuntu-latest, even when using CentOS or arm64 which GHA doesn't yet support.

So, the strategy is sync the same environment variable conventions, and exactly the same goroot targets when building docker. For efficiency reasons, we only get the last two GOROOTs, and also we get two to allow seamless migration to latest.

ex
GOROOT_1_17_X64 -> /opt/hostedtoolcache/go/1.17.1/x64
GOROOT_1_17_ARM64 -> /opt/hostedtoolcache/go/1.17.1/arm64

Note X64 doesn't match docker's TARGETARCH variable, which would evaluate to arm64. Again, even if that's more work on us docker-side, it is more consistent with what we are optimizing for. Also, note ARM64 doesn't exist in free runners. However, self-hosted runners, this is the [correct constant](https://github.com/actions/runner/blob/a592b14ae30cd9d3fccdd8cc87dcdbb471a83a00/src/Runner.Common/Constants.cs#L81).  When eventually GitHub free runners support arm64, we are totally ready for that.

So, the plan is to break up change into two parts:
1. make the centos internal image look more like the docker runner+virtual env on disk. This includes placing files in the right place, and we might as well also use the same runtime user (`runner`)
2. Re-arrange the makefile to look for these special variables ex `GOROOT_1_17_X64`. When there is no `GOROOT` exported, select one of these depending on go.mod and set it up instead. 

Using conventional variables inside the Makefile means that in normal course, we don't need to run actions/setup-go or any other equiv. Preferring GOROOT allows us to override the default image (ex via actions/setup-go) if for some reason we want a patch faster than the virtual environment (<2weeks), and it also allows local env to choose any GOROOT the dev desires.

The end result is coherence with GitHub Actions runner with no chance of runtime installation skew. The end result is also less duplicated config within GitHub Actions and between that and Travis, as Makefile sets up GOROOT consistently regardless of env.

.github/workflows/internal-images.yml
```diff
@@ -61,28 +55,30 @@ jobs:
...
+          rev=$(echo ${GOROOT}|cut -d/ -f5)
           echo "GO_MINOR_REVISION=${rev}" >> $GITHUB_ENV
```
mathetake, line 73: I think `GO_MINOR_REVISION` is the same as `GO_VERSION` exported in .github/workflows/configure_go.sh? So I think we can delete this "Export Go minor revision" step

mathetake (approved):

.github/workflows/configure_go.sh
```diff
@@ -0,0 +1,41 @@
...
+# Notes:
+# * In GitHub, these evaluate to ${RUNNER_TOOL_CACHE}/go/${GO_VERSION}*/x64
+#   * RUNNER_TOOL_CACHE lags Go releases by 1-2 weeks https://github.com/actions/virtual-environments
+# * To simulate GitHub for testing, set GITHUB_ENV and GITHUB_PATH to temporary files
+#   * Ex. `GITHUB_ENV=/tmp/test-env GITHUB_PATH=/tmp/test-path .github/workflows/configure_go.sh
+# * This uses bash because we need indirect variable expansion and GHA runners all have bash.
+set -uex pipefail
+
+go_version=$(sed -n 's/^go //gp' go.mod)
+echo GO_VERSION="${go_version}" >> "${GITHUB_ENV}"
```
mathetake, line 18: I would prefer rename this to `GO_MINOR_VERSION` for clarity per https://github.com/tetratelabs/func-e/pull/385#discussion_r720942288

.github/workflows/internal-images.yml
codefromthecrypt: it isn't really the same GO_VERSION is 1.17 and REVISION is 1.17.1 I was using go vocabulary from https://golang.org/doc/devel/release though confusingly go calls versions like 1.17 "major Go release" even if it is the minor version component updated!

another way could be GO_RELEASE and GO_REVISION?

mathetake: oh yeah sorry about that. as-is make sense!

mathetake (approved):
:100

codefromthecrypt:
ok I will merge this incrementally as it works fine until the major update of go happens. I'll do the decoupling I mentioned in a follow-up PR (or PRs)

Thanks for the review, @mathetake!

codefromthecrypt:
ps @mathetake thanks for staining my brain with thinking how to drive everything based on go.mod. While it took me several days to figure out how to accomplish that, I didn't really consider doing this as even being viable. That you spoke up and asked to make it simple really impacted my brain in the right way, even if not immediately, and I appreciate it.

codefromthecrypt:
WIP here will try to stitch it together tomorrow https://github.com/tetratelabs/func-e/compare/new-dockerfile

PR #386, opened 2021-10-04
Detects GOROOT for go.mod using make

codefromthecrypt:
This switches from external (GitHub Actions workflow) to internal
(Makefile) for configuring GOROOT. Doing so reduces the amount of
copy/paste.

This also drops build cache, and moves tools to `Tools.mk` to
make the go cache config simpler and more precise.

A future change will remove configure_go.sh entirely, by rewriting how
internal images are produced.

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -56,7 +56,7 @@ jobs:
         with:
           path: ~/go/pkg/mod
           # go.mod for go release version, go.sum for modules used, and Tools.mk for 'go run' tools
-          key: test-${{ runner.os }}-go-${{ hashFiles('go.mod', go.sum', 'Tools.mk') }}
+          key: test-${{ runner.os }}-go-${{ hashFiles('go.mod', 'go.sum', 'Tools.mk') }}
           restore-keys: test-${{ runner.os }}-go-
 
       - name: "Cache Envoy binaries"
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -51,17 +51,15 @@ jobs:
       - name: "Checkout"
         uses: actions/checkout@v2
 
-      - name: "Configure Go"
-        run: .github/workflows/configure_go.sh
+      - name: "Detect Go Release"  # only used for "Cache Go"
+        run: echo GO_RELEASE=$(sed -n 's/^go //gp' go.mod) >> "${GITHUB_ENV}"
 
       - name: "Cache Go"
         uses: actions/cache@v2
         with:
-          path: |  # Cache downloaded Go modules and the GOCACHE (to cache `go run` builds invoked by Makefile)
-            ~/go/pkg/mod
-            ${{ env.GOCACHE }}
-          key: test-${{ runner.os }}-${{ env.GO_VERSION }}-go-${{ hashFiles('go.sum', 'Makefile') }}
-          restore-keys: test-${{ runner.os }}-${{ env.GO_VERSION }}-go-
+          path: ~/go/pkg/mod
+          key: test-${{ runner.os }}-${{ env.GO_RELEASE }}-go-${{ hashFiles('go.sum', 'Tools.mk') }}
+          restore-keys: test-${{ runner.os }}-${{ env.GO_RELEASE }}-go-
 
       - name: "Cache Envoy binaries"
         uses: actions/cache@v2
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -51,15 +51,13 @@ jobs:
       - name: "Checkout"
         uses: actions/checkout@v2
 
-      - name: "Detect Go Release"  # only used for "Cache Go"
-        run: echo GO_RELEASE=$(sed -n 's/^go //gp' go.mod) >> "${GITHUB_ENV}"
-
       - name: "Cache Go"
         uses: actions/cache@v2
         with:
           path: ~/go/pkg/mod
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -5,15 +5,51 @@
...
+# Ensure POSIX-style GOROOT even in Windows, to support PATH updates in bash.
+ifdef COMSPEC
+goroot := $(shell cygpath $(goroot))
+endif
+
+# We must ensure `go` executes with GOROOT and PATH variables exported:
+# * GOROOT ensures versions don't conflict with /usr/local/go or c:\Go
+# * PATH ensures tools like golint can fork and execute the correct go binary.
+#
+# We may be using a very old version of Make (ex. 3.81 on macOS). This means we
```
codefromthecrypt, line 42: this took longer because I had to learn about floor features of make, and how things like exporting variables are very hard given the floor version installed on OS/x. Rather than force all devs to upgrade Make, this bears with the glitches and hopefully documents why.

Tools.mk
```diff
@@ -0,0 +1,7 @@
+# Copyright 2021 Tetrate
```
codefromthecrypt, line 1: this file is a much more appropriate way to taint the go mod cache

Makefile
```diff
@@ -39,20 +75,21 @@ build: $(current_binary) ## Build the func-e binary
 test_packages := . ./internal/...
 test: ## Run all unit tests
 	@printf "$(ansi_format_dark)" test "running unit tests"
-	@go test . $(test_packages)
```
codefromthecrypt, line 42: extra dot

Makefile
```diff
@@ -5,15 +5,51 @@
...
+goroot               := $(firstword $(GOROOT) $(goroot_github_env) $(goroot_github_cache) $(goroot_path))
+
+# Ensure POSIX-style GOROOT even in Windows, to support PATH updates in bash.
+ifdef COMSPEC
+goroot := $(shell cygpath $(goroot))
+endif
+
+# We must ensure `go` executes with GOROOT and PATH variables exported:
+# * GOROOT ensures versions don't conflict with /usr/local/go or c:\Go
+# * PATH ensures tools like golint can fork and execute the correct go binary.
```
codefromthecrypt, line 40: this may seem crazy, but it is really required and literally because golint-ci forking go. Took forever to figure this out.

.github/workflows/commit.yaml
```diff
@@ -51,17 +51,15 @@ jobs:
...
       - name: "Cache Go"
         uses: actions/cache@v2
         with:
-          path: |  # Cache downloaded Go modules and the GOCACHE (to cache `go run` builds invoked by Makefile)
-            ~/go/pkg/mod
-            ${{ env.GOCACHE }}
-          key: test-${{ runner.os }}-${{ env.GO_VERSION }}-go-${{ hashFiles('go.sum', 'Makefile') }}
-          restore-keys: test-${{ runner.os }}-${{ env.GO_VERSION }}-go-
+          path: ~/go/pkg/mod
+          key: test-${{ runner.os }}-${{ env.GO_RELEASE }}-go-${{ hashFiles('go.sum', 'Tools.mk') }}
```
codefromthecrypt, line 61: I will add a commit to remove the need to derive and add `${{ env.GO_RELEASE }}` to the cache key. Since we now set this with go.mod, we can more simply add go.mod to the key. That's less lines and less maintenance!

.github/workflows/commit.yaml
```diff
@@ -51,17 +51,13 @@ jobs:
...
         with:
-          path: |  # Cache downloaded Go modules and the GOCACHE (to cache `go run` builds invoked by Makefile)
-            ~/go/pkg/mod
-            ${{ env.GOCACHE }}
-          key: test-${{ runner.os }}-${{ env.GO_VERSION }}-go-${{ hashFiles('go.sum', 'Makefile') }}
-          restore-keys: test-${{ runner.os }}-${{ env.GO_VERSION }}-go-
+          path: ~/go/pkg/mod
+          # go.mod for go release version, go.sum for modules used, and Tools.mk for 'go run' tools
+          key: test-${{ runner.os }}-go-${{ hashFiles('go.mod', 'go.sum', 'Tools.mk') }}
+          restore-keys: test-${{ runner.os }}-go-
```
codefromthecrypt, line 60: taking the go version out of the restore key is ok as we aren't building with multiple versions of go intentionally. Basically, when we upgrade go it will dump the old cache, but so would it have before this simplification.. just it would have with more lines!

mathetake (approved):

Makefile
```diff
@@ -5,15 +5,56 @@
...
+github_goroot_name   := GOROOT_$(subst .,_,$(go_release))_$(github_runner_arch)
+github_goroot_val    := $(value $(github_goroot_name))
+# This works around missing variables on macOS via naming convention.
+# Ex. /Users/runner/hostedtoolcache/go/1.17.1/x64
+# Remove this after actions/virtual-environments#4156 is solved.
+github_goroot_cache  := $(lastword $(shell ls -d $(RUNNER_TOOL_CACHE)/go/$(go_release)*/$(github_runner_arch) 2>/dev/null))
+goroot_path          := $(shell go env GOROOT 2>/dev/null)
+goroot               := $(firstword $(GOROOT) $(github_goroot_val) $(github_goroot_cache) $(goroot_path))
+
+ifndef goroot
```
codefromthecrypt, line 34: added this because I was blind due to a typo in my Dockerfile until I added it :D

codefromthecrypt:
thanks for the look @mathetake!

PR #387, opened 2021-10-06
WIP: rewrite GitHub Actions runner Dockerfile

codefromthecrypt:


codefromthecrypt:
I need to find a way to get msitools on centos, and it isn't available there, and neither are all the deps needed to build it (not even rocky has all of them, notably libgcab1-devel).

However, msitools is available on fedora. Monday, I'll try running fedora msitools binaries on centos to see if they are compatible enough. If that doesn't work I might consider dropping code signing from the centos builder.

https://pkgs.org/search/?q=msitools

codefromthecrypt:
it isn't likely go 1.18 will be out before February based on release cadence, but I'll do this anyway as otherwise I'll forget all the context. https://golang.org/doc/devel/release

PR #388, opened 2021-10-06
Polishes makefile

codefromthecrypt:
* tidies how we recursively detect sources
* makes lint and formatting conditional on source change
* stops requiring gcc for e2e tests


codefromthecrypt (commented):
notes

PR #395, opened 2021-10-10
Removes broken and untested Linux package signing

codefromthecrypt:
This didn't work at all. It was hidden because variables weren't exported, so they didn't reach nfpm.

```diff
-       @$(go) run $(nfpm) pkg -f $1 --packager $2 --target $3
+       @# $(go:go=) removes the trailing 'go', so we can insert package signing variables
+       @$(go:go=) NFPM_KEYFILE=$(NFPM_KEYFILE) NFPM_KEYID=$(NFPM_KEYID) NFPM_PASSPHRASE=$(NFPM_PASSPHRASE) \
+       go run $(nfpm) pkg -f $1 --packager $2 --target $3
```

We also could have noticed this if our test scripts verified signatures, which they didn't!


codefromthecrypt:
https://github.com/tetratelabs/func-e/issues/396 will eventually re-add this

Issue #394, opened 2021-10-10
Add man pages to Linux packages


codefromthecrypt:
Currently, we generate USAGE.md using `urfave/cli` apis. We don't yet generate man pages, and we likely should. That way, when someone does `man func-e` after installing our packages, they'll see an expected result.

tip: packages are managed by nfpm, but it will be urfave/cli code that generates the man pages. Something similar to https://github.com/tetratelabs/func-e/blob/master/internal/cmd/usage_md_test.go will do the trick.

ginglis13:
Hi, can I take a crack at this? looks like `urfave/cli` include a `.ToMan` method OOTB

codefromthecrypt:
go for it, @ginglis13, thanks!

codefromthecrypt:
so the USAGE.md is shared by the website so needs to be synced always. For this one, we keep it running in unit tests and make sure it is always fresh. We also only access the code needed in the test source, which prevents any accidental inclusion in the normal binaries.

We don't currently have any utility commands, but it sounds reasonable to add one in such a way it doesn't

codefromthecrypt:
(accidentally clicked before finishing typing, /me is done now!)

codefromthecrypt:
so, yeah adding some main.go for generating the man makes sense, just I wouldn't include it in the commands func-e has available. IOTW, a separate main, and that doesn't need to be fancy since there will be no options except where to write the result.

I'm also cool with doing the same thing as we do with USAGE as in worst case hosting a static file with man page content isn't a big deal, and easier to troubleshoot.

ginglis13:
Ok, understood, so `func-e genman` is not an expected command - make a separate main or maybe folder like `man/` with generator for man page . can include the static man file from output in that folder as well. I'll open PR soon so you can see, we can discuss in that thread too

codefromthecrypt:
SG. As we may be in different timezones, if you end up going the route of verifying a file in source control, might as well add it as `packaging/nfpm/func-e.8` or similar, as that's more intuitive from the perspective of its consumer (nfpm). Regardless, indeed let's carry on in the PR. looking forward to it!

PR #398, opened 2021-10-11
add man page and short script for generating man page from app

ginglis13:
Addressing #394 

Adding roff man page generated using the urfave/cli method `ToMan`.

Adding short script to generate man page with option to write to specified path using `ToMan`. Script can be run by navigating to `packaging/nfpm` and running `go run main.go`.


ginglis13:
Got it. I didn't quite understand at first how you had meant to automatically install the man page via package manager / how it would be continously tested, but I see now 😄 I'll work towards adding that test in then and updating the nfpm config

ginglis13:
> no pressure, but I personally will be difficult to be online two days from now, for a week. a colleague can help next week. If somehow your updates happen and you don't receive feedback quickly, it isn't because you are ignored, it would be over this. Meanwhile, there's no rush on our side!

Got it, thanks so much for your help and quick responses in working on this :) Quick question for either you or colleague who might read this - what action is  needed on my side? get unit tests to pass ?

ginglis13:
Hi, thanks for the helpful hints. I believe  my branch was out of date, which caused unit tests to fail (got them passing locally, but ofc taking that with a grain of salt ...)

> in the shell scripts under packaging/nfpm, they should check in some way that the man pages work (ex man func-e

Testing with `man func-e` in the install verification scripts would require a non-minimal container image for centos and  ubuntu - it seems that `ghcr.io/tetratelabs/func-e-internal:ubuntu-20.04` and `ghcr.io/tetratelabs/func-e-internal:centos-8` both are minimized and do not contain things like documentation or man pages (and do not include the `man` command) OOTB.

ended up just checking if the func-e man page is copied to the correct directory, `/usr/local/share/man/man8`, which in my local testing it was - this should be enough verification, since a machine with `man` installed will be able to find this  man page with `man`.

> Finally, it is probably nice to note how the man page is generated into a new README.md in the packaging/nfpm directory. Ex a relative link to the source that keeps the file up-to-date.

By this do you mean add a README to packaging/nfpm that links to `internal/cmd/man_test.go` ?

Issue #393, opened 2021-10-09
Use last available version when "latestVersion" doesn't exist for a platform


codefromthecrypt:
**Is your feature request related to a problem? Please describe.**
Right now, there are two scenarios which hurt usability unnecessarily

* homebrew-core isn't up-to-date, so there's a version available for linux, but not darwin
* someone tries to use `1.19_debug` and the message is imprecise

## homebrew-core out of date

Right now, the envoyproxy project doesn't sync homebrew as a part of their release. This means windows and linux are available before darwin. Currently, we delay [archiving envoy for all platforms](https://github.com/tetratelabs/archive-envoy/) until this darwin is available. Otherwise, people freshly using func-e will not be able to because when `latestVersion` won't be available for `darwin/*` platform.

## trying 1.19_debug on something besides linux

Currently, [archive-envoy](https://github.com/tetratelabs/archive-envoy/) only has a source of debug releases for linux. If you try on darwin you'll get an imprecise message.. that `lastestVersion` isn't available instead of `no versions are available`

```bash
ENVOY_VERSIONS_URL=https://archive.tetratelabs.io/envoy/envoy-versions_debug.json ./build/func-e_darwin_amd64/func-e run -version
error: couldn't find version "1.19.1_debug" for platform "darwin/amd64"
```

**Describe the solution you'd like**
In both cases, we would be better off looking for the last available version when `latestVersion` is not available. This allows decoupled platform availability. For example, 1.20.0 can release for `linux/*` and `darwin/*` later. By searching, we can also know that a platform isn't available at all. For example, `windows/arm64` wouldn't be in any of the published lists. https://archive.tetratelabs.io/envoy

Care should be taken to not spam the user. We should still alert the user that we are looking up the latest version, just that the implementation of this prioritizes `latestVersion` from the json schema, but silently looks up the last available one if that fails. Upon failure, if no versions at all are available, we should have a better message.

Watch the edge cases in impl:
* Major updates: Ex. `latestVersion` is `2.0.0`, but the last version available for `darwin/amd64` is `1.19.1`
* Minor updates: Ex. `latestVersion` is `1.20.0`, but the last version available for `darwin/amd64` is `1.19.1`
* Patch updates: Ex. `latestVersion` is `1.20.10`, but the last version available for `darwin/amd64` is `1.20.9`
* No versions available Ex. `latestVersion` is `1.19.1_debug` and there are no versions available for that platform.

**Describe alternatives you've considered**
[...]

PR #403, opened 2021-10-29
Cleans up version selection logic

codefromthecrypt:
This removes an api for acknowledgement of the initial state. Doing so
also reduces file IO when environment variables select the version, as
in this case no default version file will be created.

This also fixes a bug where we had a false-positive on version selection.
Versions are platform-specific so we need to only be ok when the
current platform's tarball is available.

Next, this alerts the user when a remote lookup is happening. This was
missed before as we didn't let users know we are remotely resolving the
latest version.

Finally, this ensures there's enough information in warnings and errors
to troubleshoot a missing release. Particularly, there are cases where a
version is available, but not across the board. For example, linux exists,
but darwin does not.

### DIFF
--- a/e2e/func-e_use_test.go
+++ b/e2e/func-e_use_test.go
@@ -73,10 +73,14 @@ func TestFuncEUse_UnknownMinorVersion(t *testing.T) {
 	stdout, stderr, err := funcEExec("use", v)
 
 	require.EqualError(t, err, "exit status 1")
-	require.Empty(t, stdout)
-	require.Equal(t, moreos.Sprintf("error: couldn't find the latest patch for version %s\n", v), stderr)
+	require.Regexp(t, `^looking up the latest patch for Envoy version 1.1\r?\n$`, stdout)
+	stderrPattern := fmt.Sprintf("^error: https://.*json does not contain an Envoy release for version 1.1 on platform %s/%s\r?\n$", runtime.GOOS, runtime.GOARCH)
+	require.Regexp(t, stderrPattern, stderr)
 }
 
+// TODO: this test overuses bandwidth, making it tedious especially on slow networks.
+// Parse the envoy-versions.json to dynamically select the version before LastKnownEnvoyMinor if it isn't consistent.
+// That or don't update LastKnownEnvoy until it is consistent.
 func TestFuncEUse_MinorVersion(t *testing.T) {
 	// The intended minor version to be installed. This version is known to have darwin, linux, and windows binaries.
 	minorVersion := "1.18"
@@ -126,7 +130,7 @@ func TestFuncEUse_MinorVersion(t *testing.T) {
 		stdout, stderr, err := funcEExec("--home-dir", homeDir, "use", minorVersion)
 
 		require.NoError(t, err)
-		require.Regexp(t, `^downloading https:.*tar.*z\r?\n$`, stdout)
+		require.Regexp(t, `^looking up the latest patch for Envoy version 1.18\r?\ndownloading https:.*tar.*z\r?\n$`, stdout)
 		require.Empty(t, stderr)
 
 		// The binary was installed.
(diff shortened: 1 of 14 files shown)


e2e/func-e_use_test.go
```diff
@@ -73,8 +73,9 @@ func TestFuncEUse_UnknownMinorVersion(t *testing.T) {
 	stdout, stderr, err := funcEExec("use", v)
 
 	require.EqualError(t, err, "exit status 1")
-	require.Empty(t, stdout)
-	require.Equal(t, moreos.Sprintf("error: couldn't find the latest patch for version %s\n", v), stderr)
+	require.Regexp(t, `^looking up the latest patch for Envoy version 1.1\r?\n$`, stdout)
+	stderrPattern := fmt.Sprintf("^error: https://.*json does not contain an Envoy release for version 1.1 on platform %s/%s\n$", runtime.GOOS, runtime.GOARCH)
```
codefromthecrypt, line 77: much more helpful in alerting as you know why and how to check yourself!

internal/cmd/run.go
```diff
@@ -116,66 +116,65 @@ func initializeRunOpts(ctx context.Context, o *globals.GlobalOpts) error {
...
+				if _, ok = v.Tarballs[o.Platform]; ok {
+					patchVersions = append(patchVersions, k)
+				}
 			}
 			if pv := version.FindLatestPatchVersion(patchVersions, mv); pv != "" {
 				return pv, nil
 			}
-			return "", fmt.Errorf("couldn't find the latest patch for version %s", mv)
-		} else if rows, e := getInstalledVersions(o.HomeDir); e == nil {
+			err = fmt.Errorf("%s does not contain an Envoy release for version %s on platform %s", o.EnvoyVersionsURL, mv, o.Platform)
```
codefromthecrypt, line 168: before, we didn't fallback when there were no available versions. If the json is mistakenly empty, we should still check the local install for an existing patch!

codefromthecrypt (commented):
I think we're ready on this round. I backfilled tests for the new logic and also some nearby tests which weren't directly affected by this.

e2e/func-e_use_test.go
```diff
@@ -73,10 +73,14 @@ func TestFuncEUse_UnknownMinorVersion(t *testing.T) {
...
+	require.Regexp(t, stderrPattern, stderr)
 }
 
+// TODO: this test overuses bandwidth, making it tedious especially on slow networks.
+// Parse the envoy-versions.json to dynamically select the version before LastKnownEnvoyMinor if it isn't consistent.
```
codefromthecrypt, line 82: this is super annoying and I noticed it earlier, just it didn't really burn me as much until in a hotel. #393 discusses this more broadly and the solution would be change `make check` to ensure LastKnownEnvoy is consistent across all supported platforms (and have at least two patches, so min patch .1). That way, we don't need to redundantly download a second copy of an arbitrary version here. Since LastKnownEnvoy is advisory only (ex only in help docs and e2e tests) limiting this as I described shouldn't really have any impact.

internal/cmd/run_test.go
```diff
@@ -15,52 +15,180 @@
...
+	"github.com/urfave/cli/v2"
+
+	"github.com/tetratelabs/func-e/internal/moreos"
+
 	"github.com/stretchr/testify/require"
 
 	"github.com/tetratelabs/func-e/internal/globals"
 
 	"github.com/tetratelabs/func-e/internal/version"
 )
```
mathetake, line 35: I just realized that this messy import statement order hasn't been detected?

mathetake: https://github.com/tetratelabs/func-e/pull/350/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L139-L142 is the culprit 🗡️

codefromthecrypt: wonder if there's a cleaner way to accomplish that. meanwhile I can put the awk back in and say this groups imports neatly. /me grumbles why GOIMPORTS can't just format properly

mathetake: yeah I also wonder and would like to have better ways 😄

mathetake (approved):

internal/cmd/run_test.go
codefromthecrypt: ok I've restored the logic (in new style) PTAL!

mathetake (approved):

codefromthecrypt:
thanks for noticing the import glitch @mathetake!

PR #405, opened 2021-11-01
Ensures last_known_envoy.txt doesn't reference an inconsistent version

codefromthecrypt:
This changes the logic that verfies last_known_envoy.txt such that it
only expects updates when all supported platforms are available.

Before, this would update when the first platform was available, which
would break development (ex linux out before macOS).

Note: this doesn't address the similar problem at runtime, and a variant
of the problem in the e2e tests. This PR is standalone to highlight the
switch from using `jq` to go for linting.

See #393

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -51,7 +51,7 @@ const debugSuffix = "_debug"
 type Version interface {
 	// String allows access to the underlying representation. Ex. "1.18", "1.18_debug", "1.19.3_debug"
 	String() string
-	// ToMinor returns a variant used to look up the latest patch.
+	// ToMinor returns a variant used to look up the latest Patch.
 	ToMinor() MinorVersion
 }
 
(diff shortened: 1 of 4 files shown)


internal/version/version.go
```diff
@@ -121,9 +121,9 @@ func (v PatchVersion) ToMinor() MinorVersion {
 	return MinorVersion(latestPatchFormat)
 }
 
-// patch attempts to parse a patch number from the Version.String.
+// Patch attempts to parse a Patch number from the Version.String.
 // This will always succeed when created via NewVersion or NewPatchVersion
-func (v PatchVersion) patch() int {
+func (v PatchVersion) Patch() int {
```
codefromthecrypt, line 126: exporting this to help in the linting is fine imho as it helps decouple online vs offline checks (ex we don't want unit tests to have to call out to func-e.io!)

mathetake (approved):

PR #407, opened 2021-11-02
Ignores latestRelease in envoy-versions.json

### LINKED ISSUE #393: Use last available version when "latestVersion" doesn't exist for a platform

codefromthecrypt:
This stops reading latestRelease in the envoy-versions.json as it is not
platform-specific and causes usability and test glitches. Instead, the
last available version is looked up based on the current platform.
Notably, this allows linux to release ahead of macOS and without causing
problems.

Fixes #393

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -114,7 +114,7 @@ func (v PatchVersion) ToMinor() MinorVersion {
 	if matched == nil {
 		return "" // impossible if created via NewVersion or NewPatchVersion
 	}
-	return MinorVersion(matched[1] + matched[3])
+	return MinorVersion(matched[1] + matched[3]) // ex. "1.18" + ""  or "1.18" + "_debug"
 }
 
 // Patch attempts to parse a Patch number from the Version.String.
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -40,7 +40,7 @@ var LastKnownEnvoy = NewPatchVersion(lastKnownEnvoy)
 var (
 	// LastKnownEnvoyMinor is a convenience constant
 	LastKnownEnvoyMinor = LastKnownEnvoy.ToMinor()
-	versionPattern      = regexp.MustCompile(`^[1-9][0-9]*\.[0-9]+(\.[0-9]+)?(` + debugSuffix + `)?$`)
+	versionPattern      = regexp.MustCompile(`^([1-9][0-9]*\.[0-9]+)(\.[0-9]+)?(` + debugSuffix + `)?$`)
 )
 
 // debugSuffix is used to implement PatchVersion.ToMinor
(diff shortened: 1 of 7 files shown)


internal/version/version.go
```diff
@@ -110,25 +110,21 @@ func (v PatchVersion) String() string {
 
 // ToMinor satisfies Version.ToMinor
 func (v PatchVersion) ToMinor() MinorVersion {
-	splitDebug := strings.Split(string(v), debugSuffix)
```
codefromthecrypt, line 113: this is unrelated, but I fixed the only place in this file we didn't use the regex

internal/version/version.go
```diff
@@ -110,25 +110,21 @@ func (v PatchVersion) String() string {
...
-
-	if len(splitDebug) == 2 {
-		latestPatchFormat += debugSuffix
+	matched := versionPattern.FindStringSubmatch(v.String())
+	if matched == nil {
+		return "" // impossible if created via NewVersion or NewPatchVersion
 	}
-
-	return MinorVersion(latestPatchFormat)
+	return MinorVersion(matched[1] + matched[3])
```
codefromthecrypt, line 117: I chose to concat always even if only in debug matched[3] is non-empty. It was less than adding lines to avoid the allocation.. not sure if go would optimize concatenation against an empty string anyway, but max releases of envoy would ever be hundreds.

mathetake (approved):
LGTM :100:

PR #408, opened 2021-11-15
Fixes patterns used to weed out test sources

codefromthecrypt:
Previously, we used an invalid pattern style as regex isn't supported.
It was interpreted as a character class and resulted in invalid
filtering. Specifically, this prevented the "moreos" package from
showing up in coverage results. To fix this portably, we enumerate known
depths of what we don't want. Doing this looks explicit, but it
shouldn't result in high maintenance as this project isn't likely to
have deep package hierarchies.

Note: moreos is in the output and we also include globals even though it has no tests (only constants)
```bash
$ make clean test
clean      deleting temporary files
clean      ok
test       running unit tests
ok  	github.com/tetratelabs/func-e	1.065s
ok  	github.com/tetratelabs/func-e/internal/cmd	9.297s
ok  	github.com/tetratelabs/func-e/internal/envoy	4.970s
ok  	github.com/tetratelabs/func-e/internal/envoy/shutdown	7.072s
?   	github.com/tetratelabs/func-e/internal/globals	[no test files]
ok  	github.com/tetratelabs/func-e/internal/moreos	4.096s
ok  	github.com/tetratelabs/func-e/internal/tar	0.596s
ok  	github.com/tetratelabs/func-e/internal/version	0.258s
test       ok
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ non_windows_platforms := darwin_amd64 darwin_arm64 $(linux_platforms)
 windows_platforms     := windows_amd64
 
 # Make 3.81 doesn't support '**' globbing: Set explicitly instead of recursion.
-all_sources   := $(wildcard $*.go */*.go */*/*.go */*/*/*.go */*/*/*.go */*/*/*/*.go)
+all_sources   := $(wildcard *.go */*.go */*/*.go */*/*/*.go */*/*/*.go */*/*/*/*.go)
 all_testdata  := $(wildcard testdata/* */testdata/* */*/testdata/* */*/*/testdata/*)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -79,14 +79,10 @@ build: $(current_binary) ## Build the func-e binary
 
 test: ## Run all unit tests
 	@printf "$(ansi_format_dark)" test "running unit tests"
-	@$(go) test . ./internal/...
```
codefromthecrypt, line 82: this was a coherency problem between `make test` and `make coverage`!

mathetake (approved):

PR #409, opened 2021-11-16
Use t.Cleanup function to clean stuff after tests

tranngoclam:
Reference: https://golang.org/doc/go1.14#testing

Since the `t *testing.T` is being passed through many functions, we should use t.Cleanup(func() { }) to clean stuff instead of letting each individual test cases do this.

### DIFF
--- a/e2e/func-e_run_test.go
+++ b/e2e/func-e_run_test.go
@@ -52,8 +52,7 @@ var (
 func TestFuncERun(t *testing.T) {
 	t.Parallel() // uses random ports so safe to run parallel
 
-	cmd, cleanup := envoyRunTest(t, nil, minRunArgs...)
-	defer cleanup()
+	cmd := envoyRunTest(t, nil, minRunArgs...)
 
 	verifyRunArchive(t, cmd)
 }
@@ -68,7 +67,7 @@ func TestFuncERun_StaticFilesystem(t *testing.T) {
 	responseFromRunDirectory := []byte("foo")
 	require.NoError(t, os.WriteFile("response.txt", responseFromRunDirectory, 0600))
 
-	_, cleanup := envoyRunTest(t, func(ctx context.Context, c *funcE, a *adminClient) {
+	_ = envoyRunTest(t, func(ctx context.Context, c *funcE, a *adminClient) {
 		mainURL, err := a.getMainListenerURL(ctx)
 		require.NoError(t, err, "couldn't read mainURL after running [%v]", c)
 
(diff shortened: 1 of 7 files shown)


e2e/func-e_run_test.go
```diff
@@ -145,10 +142,12 @@ func envoyRunTest(t *testing.T, test func(context.Context, *funcE, *adminClient)
 	_, err = process.NewProcessWithContext(ctx, c.envoyPid) // because os.FindProcess is no-op in Linux!
 	require.Error(t, err, "expected func-e to terminate Envoy after running [%v]", c)
 
-	return c, func() {
+	t.Cleanup(func() {
 		// this may not be present if the process was kill -9'd so don't error
-		os.Remove(c.runDir + ".tar.gz") //nolint
-	}
+		require.NoError(t, os.Remove(c.runDir+".tar.gz"))
```
codefromthecrypt, line 147: this isn't a test condition, rather a cleanup.. if we did require here and it failed.. wouldn't we mask another assertion failure?

internal/cmd/app_test.go
```diff
@@ -256,9 +256,10 @@ func setupTest(t *testing.T) (*globals.GlobalOpts, func()) {
 
 	result.GetEnvoyVersions = envoy.NewGetVersions(result.EnvoyVersionsURL, result.Platform, result.Version)
 
-	return &result, func() {
+	t.Cleanup(func() {
```
codefromthecrypt, line 259: if we are using `t.Cleanup` wouldn't we no longer use `tearDown[n]`?

internal/envoy/install_test.go
```diff
@@ -267,5 +262,8 @@ func setupInstallTest(t *testing.T) (*installTest, func()) {
 		},
 	}
 	setup.GetEnvoyVersions = NewGetVersions(setup.EnvoyVersionsURL, setup.Platform, setup.Version)
-	return setup, versionsServer.Close
+	t.Cleanup(func() {
```
codefromthecrypt, line 265: please use `versionsServer.Close` instead of an anonymous function to do the same

codefromthecrypt (commented):
Thanks for the cleanup here.. I think this is a big win. There may be places where we make a slice of `cleanUp[]` and can also convert.. can you sweep for these so that we can avoid a follow-up PR?

e2e/func-e_run_test.go
codefromthecrypt: ex. the comment `// this may not be present if the process was kill -9'd so don't error` was saying why we didn't enforce errors here, so if we do enforce them now, we should delete that comment. I think it is better to leave it as it wasn't that simple to figure this out the first time 😅

tranngoclam:
@codefromthecrypt I simplified the code as your suggestion 😅

tranngoclam:
@codefromthecrypt I think we should consume the err to pass the errcheck of golangco-lint, and remove //nolint. Something like this:

```go
_, _ = moreos.Fprintf(stderr, "%s\n", err)
```
instead of
```go
moreos.Fprintf(stderr, "%s\n", err) //nolint
```

What do you think ?

codefromthecrypt:
@tranngoclam I agree on the signature change of `moreos.Fprintf(stderr, "%s\n", err)` as I don't think we have anything consuming the return values. it only returns to match `fmt`, which isn't a great reason to have to `//nolint` everywhere. Wanna do that in another PR?

codefromthecrypt (approved):
this PR looks lovely in red. Thanks!

PR #416, opened 2021-12-01
Uses latest runners when running tests

codefromthecrypt:
func-e is an introductory tool, so we can optimize to test the latest
available OS vs what GitHub Actions marks as latest (often years old).

Note: MacOS 12 is not supported on GHA, still!

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -44,7 +44,7 @@ jobs:
     timeout-minutes: 90  # instead of 360 by default
     strategy:
       fail-fast: false  # don't fail fast as sometimes failures are operating system specific
-      matrix:
+      matrix:  # use latest available versions and be consistent on all workflows!
         os: [ubuntu-20.04, macos-11, windows-2022]
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -41,8 +41,6 @@ jobs:
           WINDOWS_CODESIGN_P12: ${{ steps.p12.outputs.filePath }}
           WINDOWS_CODESIGN_PASSWORD: ${{ secrets.WINDOWS_CODESIGN_PASSWORD }}
 
-      # The release notable has a discussion channel to avoid people having to
```
codefromthecrypt, line 44: this was a typo and a lie :D

.github/workflows/commit.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
     strategy:
       fail-fast: false  # don't fail fast as sometimes failures are operating system specific
       matrix:
```
codefromthecrypt, line 47:
```suggestion
      matrix:  # use latest available versions and be consistent on all workflows!
```

mathetake (approved):

PR #419, opened 2021-12-08
Move packages from /internal to /pkg

arkodg:
* Allows external repositories to import func-e packages so that developers can leverage `func-e` to offload envoy binary arch type, version, download repository etc.
* This allows the developer to import this go package to mimic `func-e run` w/o downloading
`envoy` or `func-e` for quick prototyping and end to end testing.
* More reuse, less code duplication :) 
Signed-off-by: Arko Dasgupta <arko@tetrate.io>


codefromthecrypt:
if we had an api, it would basically only expose the features the CLI can do and no more, and with the minimal types needed to do that.

ex. an interface FuncE with functions defined for Run, Use, etc with the same allowed arguments as the CLI. This would then also be consumed by the CLI and be available for embedding. I'm not sure the package named `pkg` is modern nowadays, and it is certainly vague. This interface could be in its own package like api.

arkodg:
thanks for the advice, will take another pass at only moving the minimum packages needed to call the `func-e` CLI in a programmatic way

codefromthecrypt:
I presume any change would be a different PR so I'll go ahead and close this to keep the queue clean. Take care!

PR #420, opened 2022-01-05
Bumps versions to latest

codefromthecrypt:


codefromthecrypt:
FYI last release compiled with go 1.17.3. Current (in use for CI) is 1.17.5, so next release will include any changes from that also.

codefromthecrypt:
I checked https://deploy-preview-420--func-e.netlify.app/ since I updated hugo and the corresponding theme to latest

PR #426, opened 2022-03-23
Make the example on func-e.io pastable into a terminal

morepork:
Currently it includes the leading $, and includes the whole block, this
change removes the leading $ and splits it into two blocks, one for the
install command, and the other to run func-e

### DIFF
--- a/site/content/_index.md
+++ b/site/content/_index.md
@@ -9,10 +9,10 @@ your behalf. This makes knowledge sharing and troubleshooting easier, especially
 
 ```sh
 curl https://func-e.io/install.sh | bash -s -- -b /usr/local/bin
+func-e run -c /path/to/envoy.yaml
 ```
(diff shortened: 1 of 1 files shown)


codefromthecrypt (approved):
I moved the more advanced example to the bottom as if the goal is copy/pasting what's in the box, this allows the first box to be coherently copy/pasta'd!


See https://deploy-preview-426--func-e.netlify.app/

morepork:
Hey @codefromthecrypt, how do I get the unit tests to run?

codefromthecrypt:
unit tests intentionally don't run on website commit as it is wasteful as there are no unit tests for static content. same thing with README update. If the deploy failed, you would get a fail notice. GH is unsophisticated about this nuance so when there's a change like this an admin has to merge it.

PR #427, opened 2022-05-08
Update dependencies, notably Go 1.18

codefromthecrypt:
This updates so func-e releases with Go 1.18 https://go.dev/blog/go1.18

* This also removes some workarounds no longer needed due to urfave release.
* This fixes docker drift caused by CentOS 8 going EOL

Note: This doesn't move up to latest Ubuntu or CentOS intentionally. End users
of Envoy are still asking even for CentOS 7 support! Also, Ubuntu 22.04 isn't yet
working with the qemu action.

Note: This changed some code around end-to-end (e2e) testing, to explicitly
install the "last known" envoy vs assuming that's the latest.

Envoy maintainers are not paying attention to homebrew, forcing routine skew 
on "last known version". Hence, this changes the e2e to tolerate more at
the cost of redundant downloads for Linux or Windows when they work.
See https://github.com/Homebrew/homebrew-core/pull/100574


codefromthecrypt:
here's the deploy preview (showing hugo and the theme update didn't break anything) https://627735d53b8ea07a166de777--func-e.netlify.app/

mathetake (approved):

codefromthecrypt:
thanks for the look @mathetake!

PR #429, opened 2022-05-09
Removes old winget_manifest script

codefromthecrypt:
Since func-e is upstream we don't need this anymore. It should have been
removed during 3db5c5738a2c52bd62befd06830ec744361b6a77

this repo can be nuked for the same reason https://github.com/tetratelabs/winget-pkgs


codefromthecrypt:
heh whoops this is still needed, though https://github.com/tetratelabs/winget-pkgs isn't used

PR #430, opened 2022-05-28
Updates to use GitHub Action for codecov

codefromthecrypt:
The codecov GitHub Action replaces the uploader script for projects that
are either mostly amd64 or can consolidate uploads into one stage that
runs on amd64. func-e has no significant arm64 source files, so can use
this action instead of the uploader script.

This also bumps versions which should get rid of the dependabot warning.

Thanks to @anuraaga for noticing the uploader script is deprecated.


codefromthecrypt:
note: I switched the tests to use os/x 12 because recently the envoy build no longer works with multiple OS/x versions cc @cpakulski

mathetake (approved):

.github/workflows/commit.yaml
```diff
@@ -91,6 +91,7 @@ jobs:
 
       - name: "Upload coverage report"  # only on master push and only once (not per OS)
         if: github.event_name == 'push' && github.ref == 'refs/heads/master' && runner.os == 'Linux'
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-        run: bash <(curl -s https://codecov.io/bash)
+        uses: codecov/codecov-action@v2
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
anuraaga, line 96: BTW for OSS projects token shouldn't be needed when using the action I think. One less secret to manage

codefromthecrypt: will verify here then propagate. https://github.com/tetratelabs/car/pull/19

codefromthecrypt: verified in car and https://github.com/tetratelabs/func-e/pull/431 will do it here

PR #434, opened 2022-10-28
Adds api for library usage

codefromthecrypt:
This is a sketch.  A real impl will need to test things, similar to our e2e test. Particularly there could be some tricky things for example, how to signal close since the envoy run will block the calling thread. This is mainly to show that the api approach should be precise and only expose top-level concepts that relevant to import.

Fixes: https://github.com/tetratelabs/func-e/issues/433


api/run.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/tetratelabs/func-e/internal/globals"
+	"github.com/tetratelabs/func-e/internal/version"
+	"io"
+	"os"
+	"runtime"
+)
+
+// HomeDir is an absolute path which most importantly contains "versions"
+// installed from EnvoyVersionsURL. Defaults to "${HOME}/.func-e"
+func HomeDir(homeDir string) RunOption {
```
codefromthecrypt, line 15: ex unit tests can set this to t.TempDir

api/run.go
```diff
@@ -0,0 +1,81 @@
...
+	return func(o *runOpts) {
+		o.homeDir = homeDir
+	}
+}
+
+// EnvoyVersionsURL is the path to the envoy-versions.json.
+// Defaults to "https://archive.tetratelabs.io/envoy/envoy-versions.json"
+func EnvoyVersionsURL(envoyVersionsURL string) RunOption {
```
codefromthecrypt, line 23: ex unit tests can set this to httptest server

api/run.go
```diff
@@ -0,0 +1,81 @@
...
+// EnvoyVersionsURL. Its value can be in full version major.minor.patch format,
+// e.g. 1.18.1 or without patch component, major.minor, e.g. 1.18.
+func EnvoyVersion(envoyVersion string) RunOption {
+	return func(o *runOpts) {
+		o.envoyVersion = envoyVersion
+	}
+}
+
+// Out is where status messages are written. Defaults to os.Stdout
+func Out(out io.Writer) RunOption {
```
codefromthecrypt, line 42: ex unit tests can set this to bytes.Buffer and readback

api/run.go
```diff
@@ -0,0 +1,81 @@
...
+type RunOption func(*runOpts)
+
+type runOpts struct {
+	homeDir          string
+	envoyVersion     string
+	envoyVersionsURL string
+	out              io.Writer
+}
+
+func Run(args []string, options ...RunOption) error {
```
codefromthecrypt, line 58: probably missing an option because in library mode, you likely want a cancelable context as first param. that way it doesn't block forever.

codefromthecrypt (commented):
added notes on how to proceed

api/run.go
arkodg: not sure what you mean here

codefromthecrypt:
protip. we have some infrastructure that allows unit testing w/o remote download deps

something like..
```go
homeDir := t.TempDir()
envoyVersion := version.LastKnownEnvoy
versionsServer := test.RequireEnvoyVersionsTestServer(t, envoyVersion)
defer versionsServer.Close()
envoyVersionsURL := versionsServer.URL + "/envoy-versions.json"
out := new(bytes.Buffer)

... now use em
```

api/run.go
arkodg: cool makes sense with https://github.com/tetratelabs/func-e/pull/434#issuecomment-1295587586

arkodg:
looks like the value in https://github.com/tetratelabs/func-e/blob/master/internal/version/last_known_envoy.txt is causing the CI to fail

codefromthecrypt:
seems thing haven't been maintained in a while. I ran the 1.23.1 tag on archive-envoy and then raised this to move to it https://github.com/tetratelabs/func-e/pull/435

arkodg:
getting this error on windows
```
--- FAIL: TestRun (5.79s)
[162](https://github.com/tetratelabs/func-e/actions/runs/3364779609/jobs/5579526006#step:6:163)
    testing.go:1090: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestRun607085353\001\runs\1667250752759535200\stderr.log: The process cannot access the file because it is being used by another process.
```
@codefromthecrypt any idea why that might be happening ?

api/run.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	funcECmd := cmd.NewApp(&o)
+
+	funcERunArgs := []string{"func-e", "--platform", runtime.GOOS + "/" + runtime.GOARCH, "run"}
+	funcERunArgs = append(funcERunArgs, args...)
+
+	errChan := make(chan error)
+	go func() {
+		errChan <- funcECmd.Run(funcERunArgs)
```
codefromthecrypt, line 101:
```suggestion
		errChan <- funcECmd.RunContext(ctx, funcERunArgs)
```

api/run_test.go
```diff
@@ -0,0 +1,63 @@
...
+	require.Equal(t, 0, b.Len())
+
+	// 1. Pass a context with a timeout to Run to start running Envoy
+	// 2. Wait until the context is done
+	// 3. Ensure that the error is nil
+	var err error
+	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
+	defer cancel()
+	go func() {
```
codefromthecrypt, line 54: I don't think you need to use a goroutine here. If the goal is to make sure envoy can start maybe use `--version` then block until done? as long as the context timeout is used properly it should return when complete.

codefromthecrypt (commented):
added some thoughts, noting that windows concerns are a big reason why the e2e directory has some custom runner code, and also that we have things like moreos

api/run_test.go
arkodg: ha can use `--version`

arkodg:
Hey @mathetake can you ptal ( since @codefromthecrypt is a co-author ) ?

codefromthecrypt:
Since it isn't obvious to everyone why I closed this. Here's why.

This repo has no stale bot. Things dead without impact for several months should be closed. Anyone who is motivated and can do a decent job in execution can carry it forward. However, I'm not going to hold that up and have closed all PRs on all repos that have been dead for periods of 6 months or longer.

PR #437, opened 2023-01-05
Builds with Go 1.19

codefromthecrypt:
This updates the build to use Go 1.9 and addresses drift on the most recent Envoy version.

This also addresses some build implications that came up recently, for example osslsigncode doesn't work at the moment on macos (at least not in homebrew). Also, Envoy no longer works on CentOS 8.


codefromthecrypt:
netlify failed as it was configured on the website (not in anything here) to use old ubuntu which they deprecated. I kicked it manually as it seems not possible to refresh via actions. https://deploy-preview-437--func-e.netlify.app/

mathetake (approved):

codefromthecrypt:
attempting a revert of the nfpm version as otherwise this hints something botched our org secrets

codefromthecrypt:
so centos 8 is a dead duck

```bash
/home/runner/.func-e/versions/1.24.1/bin/envoy: /lib64/libm.so.6: version `GLIBC_2.29' not found (required by /home/runner/.func-e/versions/1.24.1/bin/envoy)
```

codefromthecrypt:
kicked off action to build an internal image of centos:9 since :8 is busted with recent envoy https://github.com/tetratelabs/func-e/actions/runs/3851413444

codefromthecrypt:
what's left is a signature issue on the windows installer. It says wrong password, which could either be a problem in the binary signing, or actually somehow our password is messed up.

```
Failed to parse PKCS#12 file: packaging/msi/func-e.p12 (Wrong password?)
00C6D30501000000:error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:crypto/evp/evp_fetch.c:373:Global default library context, Algorithm (RC2-40-CBC : 0), Properties ()
Failed
make: *** [build/func-e_windows_amd64/func-e.exe.signed] Error 255
rm build/func-e_darwin_arm64/func-e build/func-e_windows_amd64/func-e.exe build/func-e_linux_arm64/func-e build/func-e_linux_amd64/func-e
```

codefromthecrypt:
ok all things work now except package signing on macos-12 which is busted for reasons including homebrew removing the ability to use old versions of tools. This workflow is optional because we don't actually execute signing commands with macos. However, still hope it works again, but anyway I can't personally track it.

PR #442, opened 2023-03-07
build: updates to latests tools and envoy version

codefromthecrypt:


codefromthecrypt:
looks like we need to lift the windows stuff from wazero back here, huh :D

evacchi:
you mean the env vars/ifs right? 🤔

codefromthecrypt:
> you mean the env vars/ifs right? 🤔

yeah all that. basically, the most recent working version of the packaging flow needs to be backported here. if you are game for pushing commits millions of good karma will rain from above :D

evacchi:
🎶 *I miss the karma rains in packaging* 🎶

evacchi (approved):

codefromthecrypt:
ps this PR isn't high priority so I will let it sit while we finish up wazero

codefromthecrypt:
disregard last comment, as @evacchi decided to karmafy

codefromthecrypt:
hmm still packaging glitch decrypting the password.

evacchi:
> that fails for me too on my machine 🤔
> 
> EDIT: I think `func-e.p12` got corrupted or replaced, by running the lines here https://github.com/tetratelabs/func-e/blob/master/Makefile#L279-L280 fixes it for me
> 
> ```shell
> keytool -genkey -alias func-e -storetype PKCS12 -keyalg RSA -keysize 2048 -storepass func-e-bunch \
>    -keystore func-e.p12 -dname "O=func-e,CN=func-e.io" -validity 3650
> ```

I remember I had this issue at some point! I addressed this in #444 by committing the result of this command

PR #444, opened 2023-04-21
packaging: use native WiX toolchain on Windows

evacchi:
Cross-platform msitools do not support the `Environment` directive,
which should be the proper way to set up the PATH environment variable,
so we adapt the Makefile to assume a Windows environment when building.

`make dist` now runs on all platforms, but on Windows, assuming
`candle.exe`, `light.exe` and GNU `tar` are installed, it builds
all tarballs, zip and msi packages.

Notice this was already supported, but it still relied on `SETX`
to support building cross-platform.

On other platforms it will only build the tarballs.

Signed-off-by: Edoardo Vacchi <evacchi@users.noreply.github.com>


Makefile
```diff
@@ -118,18 +118,14 @@ dist/func-e_$(VERSION)_%.tar.gz: build/func-e_%/func-e
...
 	$(call go-build,$@,$<)
+endif
 
 dist/func-e_$(VERSION)_%.zip: build/func-e_%/func-e.exe.signed
 	@printf "$(ansi_format_dark)" zip "zipping $@"
 	@mkdir -p $(@D)
-ifeq ($(goos),windows)  # Windows 10 etc use 7zip
-	@# '-bso0 -bsp0' makes it quiet, except errors
-	@# Wildcards in 7zip will skip the directory. We single-quote to ensure they aren't interpreted by the shell.
-	@7z -bso0 -bsp0 a $@ './$(dir $<)/*.exe'
```
evacchi, line 129: no need, we install infozip with choice

Makefile
```diff
@@ -167,33 +163,32 @@ msi_version := $(VERSION:dev=0.0.1)
 
 # This builds the Windows installer (MSI) using platform-dependent WIX commands.
 dist/func-e_$(VERSION)_%.msi: build/func-e_%/func-e.exe.signed
+ifeq ($(OS),Windows_NT)  # Windows 10 etc use https://wixtoolset.org
 	@printf "$(ansi_format_dark)" msi "building $@"
 	@mkdir -p $(@D)
-ifeq ($(goos),windows)  # Windows 10 etc use https://wixtoolset.org
 	@candle -nologo -arch $(call msi-arch,$@) -dVersion=$(msi_version) -dBin=$(<:.signed=) packaging/msi/func-e.wxs
 	@light -nologo func-e.wixobj -o $@ -spdb
 	@rm func-e.wixobj
```
evacchi, line 171: funnily enough, this was already possible for func-e.

Makefile
```diff
@@ -167,33 +163,32 @@ msi_version := $(VERSION:dev=0.0.1)
...
+non_windows_archives  := $(non_windows_platforms:%=dist/func-e_$(VERSION)_%.tar.gz)
+windows_archives      := $(windows_platforms:%=dist/func-e_$(VERSION)_%.zip) $(windows_platforms:%=dist/func-e_$(VERSION)_%.msi)
 archives  := $(non_windows_platforms:%=dist/func-e_$(VERSION)_%.tar.gz) $(windows_platforms:%=dist/func-e_$(VERSION)_%.zip)
-packages  := $(windows_platforms:%=dist/func-e_$(VERSION)_%.msi) $(linux_platforms:%=dist/func-e_$(VERSION)_%.deb) $(rpms)
 checksums := dist/func-e_$(VERSION)_checksums.txt
 
 # Darwin doesn't have sha256sum. See https://github.com/actions/virtual-environments/issues/90
 sha256sum := $(if $(findstring darwin,$(goos)),shasum -a 256,sha256sum)
-$(checksums): $(archives) $(packages)
+$(checksums): $(non_windows_archives) $(if $(findstring Windows_NT,$(OS)),$(windows_archives),)
```
evacchi, line 184: aligning to the wazero strategy

evacchi:
uhm,

```
Failed to parse PKCS#12 file: packaging/msi/func-e.p12 (Wrong password?)
D4080000:error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:crypto\evp\evp_fetch.c:349:Global default library context, Algorithm (RC2-40-CBC : 0), Properties ()
```

I think the env var may be empty as in https://github.com/tetratelabs/wazero/pull/1189

Makefile
evacchi, line 1: for some reasons on my Windows+Cygwin machine this breaks the PATH in funky ways, but it works if I just comment it out...

```
# Ensure POSIX-style GOROOT even in Windows, to support PATH updates in bash.
ifdef COMSPEC
goroot := $(shell cygpath $(goroot))
endif
```

Makefile
codefromthecrypt: oops..

codefromthecrypt (approved):
Thanks for the tenacity to follow-up here!

PR #446, opened 2023-08-09
Allow downstream projects to import func-e

arkodg:
* Define an api that allows go projects to use func-e as a library

Takes https://github.com/tetratelabs/func-e/pull/434 forward

Fixes: https://github.com/tetratelabs/func-e/issues/433

### DIFF
--- a/api/run.go
+++ b/api/run.go
@@ -1,0 +1,111 @@
+// Copyright 2022 Tetrate
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package api allows go projects to use func-e as a library.
+package api
+
+import (
+	"context"
+	"io"
+	"os"
+	"runtime"
+
+	"github.com/tetratelabs/func-e/internal/cmd"
+	"github.com/tetratelabs/func-e/internal/globals"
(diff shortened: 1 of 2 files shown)


arkodg:
CI is failing due to an unrelated test failing https://github.com/tetratelabs/func-e/blob/9c44e5090080ecb742db5c386fd30256a0448979/lint/last_known_envoy_test.go#L37

api/run.go
```diff
@@ -0,0 +1,111 @@
...
+	go func() {
+		errChan <- funcECmd.RunContext(ctx, funcERunArgs)
+	}()
+
+	// Wait for run to exit or an explicit stop.
+	select {
+	case <-ctx.Done():
+		return nil
+	case err := <-errChan:
+		return err
```
mathetake, line 109: I think this hasn't been covered in tests?

mathetake: ack it may be difficult to cover

arkodg: added another test case to account for both  paths

mathetake: good!

api/run.go
```diff
@@ -0,0 +1,111 @@
...
+
+	errChan := make(chan error)
+	go func() {
+		errChan <- funcECmd.RunContext(ctx, funcERunArgs)
+	}()
+
+	// Wait for run to exit or an explicit stop.
+	select {
+	case <-ctx.Done():
+		return nil
```
mathetake, line 107: shouldn't we kill the envoy process in this case? (not sure if there's such function in func-e now)

arkodg: isn't this done in the called func libs https://github.com/tetratelabs/func-e/blob/82577c96a87fc3a5c5d7639add588cadb3b5fe9c/internal/envoy/run.go#L78 ?

mathetake: ok cool but it seems to me that the code block only reaches when the OS signal was sent. Is there any way to verify that it works as intended? e.g. passing the fake config that runs indefinitely in tests?

arkodg: can you point me to some tests in the code base that verify this, that I can peruse here ?

mathetake (approved):
YOLO

PR #456, opened 2025-03-18
ExitChannel to allow to wait for envoy process to exit

mathetake:

PR #458, opened 2025-03-19
fixes race condition in Run by decoupling singal handling from the impl

mathetake:
Redo of #456. Basically, the original impl had a race condition and cannot wait for the Envoy processes to actually exit plus there was a race condition when accessing exec.Process objects. This fixes it altogether by decoupling internal/envoy/run implementation from the actual signal handling and making it focus on the simple context cancelation.

### DIFF
--- a/api/run.go
+++ b/api/run.go
@@ -73,6 +73,9 @@ type runOpts struct {
 
 // Run downloads Envoy and runs it as a process with the arguments
 // passed to it. Use RunOption for configuration options.
+//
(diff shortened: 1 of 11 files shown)


codefromthecrypt (approved):
right, this wasn't originally designed to be called as a library, so moving the signal handling to main makes sense.

mathetake:
The tests running both on EG (https://github.com/envoyproxy/gateway/pull/5527) and AIGW (https://github.com/envoyproxy/ai-gateway/pull/458) using this branch. good to go

PR #460, opened 2025-03-20
Suppress noisy logging

mathetake:
func-e runs Envoy for given args, so there's no reason to log all arguments as that should be trivial to users. The only portion that is not trivial during debugging is the binary location managed by func-e, so this switches to print it only. This will suppress crazy large output by a library user such as Envoy Gateway that passes a giant literal yaml.


mathetake:
oh man

arkodg (approved):

PR #466, opened 2025-07-04
Fix run default which made a directory literally named ${HOME}

codefromthecrypt:
`globals.DefaultHomeDir` is used in text description, "" is the value to use for default


mathetake (approved):

codefromthecrypt:
verified this works (and also the wasm build not just loads but works)

PR #467, opened 2025-07-15
Cleans up api.Run and backfills ephemeral admin server when not in config

codefromthecrypt:
This makes three changes that improve the quality of func-e from a user's perspective:
 - Change `api.Run` to call library code directly instead of routing through the CLI library (urfave)
   - This stops pollution of func-e arg and env parsing as well simplifies lifecycle in callers such as EG
 - Add `api.RunOptions` for `EnvoyOut` and `EnvoyErr`
   - This prevents callers from being unable to control subprocess logging, e.g. was going to `os.Stdout/Stderr`
 - Make admin server config optional from the user's POV
   - When a caller didn't pass an admin server, errors would go into the logs as func-e assumed it would be there

---

Other changes were tech debt clearing to clear the path for this. There is still more to do, but this is the largest bit.

- Made a new consolidated e2e suite that can be used regardless of if func-e is compiled.
  - Specifically, this allows use of a debugger to hunt down state issues.
- Refactored a new, realistic fake Envoy binary for all test scenarios.
  - This actually crashes like envoy does on no config and starts listeners our tests use.
- Removed legacy test helpers and fakebinary code, consolidating test logic and sources.
  - There were traps like main code copy/pasted into tests, esp in signal handling
- Added config parsing code for admin and static listeners, and backfilled tests
  - Config parsing now hits edge cases from envoy's actual source, and tests use the listeners.
- Updated Envoy to reflect latest version and updated test tools
  - Formatting and linters have drifted and this is all sorted now
- Removed inner timeout in e2e tests
  - Having context timeouts inside e2e tests make a race between the go test timeout and the inner one


api/func-e.go
```diff
@@ -61,6 +60,27 @@ func Out(out io.Writer) RunOption {
 	}
 }
 
+// EnvoyOut sets the writer for Envoy stdout
+func EnvoyOut(w io.Writer) RunOption {
```
codefromthecrypt, line 64: api changes here.. only 2 of them, and they are additional. In general, this package is a bomb as it ties api to impl, but we can shave that yak in a later PR.

internal/test/e2e/testrun.go
```diff
@@ -0,0 +1,271 @@
...
+
+	"github.com/shirou/gopsutil/v3/process"
+	"github.com/stretchr/testify/require"
+
+	"github.com/tetratelabs/func-e/internal"
+	"github.com/tetratelabs/func-e/internal/tar"
+	"github.com/tetratelabs/func-e/internal/test/morerequire"
+)
+
+var envoyStartedLine = "starting main dispatch loop"
```
codefromthecrypt, line 35: here's the file with the new consolidated test cases which cover what was in e2e (now reusable with both a built binary and library calls)

api/func-e.go
```diff
@@ -77,20 +104,39 @@ type runOpts struct {
...
+			EnvoyPath: ro.envoyPath,
+			EnvoyOut:  ro.envoyOut,
+			EnvoyErr:  ro.envoyErr,
+		},
+	}
+	if err := api.InitializeGlobalOpts(o, ro.envoyVersionsURL, ro.homeDir, ""); err != nil {
+		return nil, err
 	}
 
-	funcECmd := cmd.NewApp(&o)
```
codefromthecrypt, line 92: before, this routed library logic through urfave which was creepy and made it near impossible to understand

api/func-e_run_test.go
```diff
@@ -0,0 +1,38 @@
...
+package api
+
+import (
+	"context"
+	"testing"
+
+	"github.com/tetratelabs/func-e/internal/test/e2e"
+)
+
+func TestRun(t *testing.T) {
```
codefromthecrypt, line 24: example of "e2e" run tests applied to the library mode, which works in a debugger

Tools.mk
```diff
@@ -1,8 +1,9 @@
 # Copyright 2021 Tetrate
 # Licensed under the Apache License, Version 2.0 (the "License")
 
-golangci_lint := github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.6
-goimports     := golang.org/x/tools/cmd/goimports@v0.5.0
+golangci_lint := github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.2.1
+gofumpt       := mvdan.cc/gofumpt@v0.8.0
+gosimports    := github.com/rinchsan/gosimports/cmd/gosimports@v0.3.8
```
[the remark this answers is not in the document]
codefromthecrypt: probably, but I'm about 50hrs over time on what I was looking to accomplish. let's do more of this sort of thing in another PR

codefromthecrypt:
PS I ran with envoy gateway using go.mod replace and it is source compatible.  internal/infrastructure/host/proxy_infra_test.go finished with no errors

Tools.mk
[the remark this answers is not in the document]
codefromthecrypt: https://github.com/tetratelabs/func-e/pull/471/

PR #471, opened 2025-07-16
Updates to latest hugo and switches to "go tool"

codefromthecrypt:
Before, we used "go run" to invoke tools like linters in our Makefile. This switches to "go tool" which is the latest way to do this. The main side-effect is a larger go.mod.

As a part of this, I updated Hugo which drifted our theme which is abandoned. Hence, I made a patch until we switch to something else. See https://github.com/panr/hugo-theme-hello-friend


codefromthecrypt:
reverted the go tool stuff and focused on updates and backfilling doc

PR #473, opened 2025-07-18
Add ctrl+c twice support with shutdown hook panic recovery

codefromthecrypt:
This implements ctrl+c twice functionality where the first interrupt starts graceful shutdown and any subsequent interrupt forces immediate exit via os.Exit(1).

The implementation handles normal shutdown scenarios gracefully. The known limitation on Darwin (envoy orphaning on SIGKILL) is documented and tested.

Fixes #456


codefromthecrypt:
incomplete.. there's a flake

PR #474, opened 2025-07-20
Remove Shutdown Hooks and Dump Admin State on Startup

codefromthecrypt:
This simplifies func-e's lifecycle by removing shutdown hooks and collecting admin state on startup instead. Essential xDS configuration data is captured when Envoy's dispatch loop starts, making it available immediately when needed. This makes things more coherent and allows us to leverage `CommandContext` for signal propagation.

This also aligns exit behavior with native Envoy (code 0 on graceful shutdown/interrupt). This allows `api.Run` to return nil on effective success (context canceled or exit code zero), increasing the signal of the error return.

## Impact

* **Breaking:** No archived run dirs or shutdown files; `Run` returns nil on interrupts (vs. prior errors).
* **Benefits:** Reduced complexity/leaks; reliable cross-platform; smaller binary; essential xDS data available when running.

## Key Changes

The key improvement is that essential xDS data is now captured reliably during startup when Envoy's admin API becomes available, rather than attempting complex shutdown coordination. This provides the necessary configuration visibility with significantly less complexity and better cross-platform reliability

* **Startup:** Collect config dump with comprehensive xDS data (LDS/RDS/CDS/EDS/SDS) via admin API after "main dispatch loop" starts. Essential configuration state available immediately.
* **Shutdown:** Remove hooks (admin/node data archiving). Return nil on clean exits/interrupts.
* **Interrupts:** Handle multiples (e.g., Ctrl+C x2) gracefully; add e2e test.
* **Run Dir:** Now holds `stdout.log`, `stderr.log`, `envoy.pid`, `config_dump.json` with full xDS snapshot. No archiving.
* **Platform:** Simplify proc attrs (Pdeathsig on Linux only); fix macOS zombie issues.
* **Testing:** Add access logging to configs which allows us to test envoy's stdout; verify logs/run dir in e2e. Rename `TestRun_MinimalListener` to `TestRun_RunDirectory`.
* **Cleanup:** Remove `golang.org/x/sync`; update docs/help (no hooks/archiving mentions).
* **Version:** Update to 1.34.3 in code/docs.


internal/cmd/run_test.go
```diff
@@ -24,53 +26,50 @@ import (
 	"github.com/tetratelabs/func-e/internal/version"
 )
 
+func init() {
+	// Don't let urfave quit the current test process on cancel!
+	cli.OsExiter = func(code int) { log.Printf("urfave called exit: %d", code) }
```
codefromthecrypt, line 31: I'm sure @mathetake and @anuraaga could guess the rage level when I finally figured out what was quitting the whole test process.

codefromthecrypt:
ps after this change I will look into allowing the user to supply their own StartupHook or ability to remove the default one.

mathetake (approved):

PR #476, opened 2025-07-21
Public API Separation

codefromthecrypt:
Previously, an api package was added, but it also put the implementation there. This causes package cycle problems and subverts the idea of an API.

This migrates `api.Run` (impl) to `api.RunFunc` (api) and `func_e.Run` (impl). This also backfills a test to make sure the code works.

To do this, we migrate `/main.go` -> `/cmd/func-e/main.go`, which allows us to re-use the root package as a library, instead of a binary (main) one.

There is impact, but to only one user: envoy gateway. The fix is mechanical and allows us to clear tech debt before making more features.

```diff
diff --git a/internal/infrastructure/host/proxy_infra.go b/internal/infrastructure/host/proxy_infra.go
index f384b6854..229fdcd1c 100644
--- a/internal/infrastructure/host/proxy_infra.go
+++ b/internal/infrastructure/host/proxy_infra.go
@@ -12,7 +12,8 @@ import (
        "os"
        "path/filepath"

-       funcE "github.com/tetratelabs/func-e/api"
+       func_e "github.com/tetratelabs/func-e"
+       "github.com/tetratelabs/func-e/api"
        "k8s.io/utils/ptr"

        egv1a1 "github.com/envoyproxy/gateway/api/v1alpha1"
@@ -92,7 +93,7 @@ func (i *Infra) runEnvoy(ctx context.Context, out io.Writer, name string, args [
                defer func() {
                        exit <- struct{}{}
                }()
-               err := funcE.Run(pCtx, args, funcE.HomeDir(i.HomeDir), funcE.Out(out))
+               err := func_e.Run(pCtx, args, api.HomeDir(i.HomeDir), api.Out(out))
                if err != nil {
                        i.Logger.Error(err, "failed to run envoy")
                }
diff --git a/internal/infrastructure/host/proxy_infra_test.go b/internal/infrastructure/host/proxy_infra_test.go
index 2d82cc689..36bc997f0 100644
--- a/internal/infrastructure/host/proxy_infra_test.go
+++ b/internal/infrastructure/host/proxy_infra_test.go
@@ -13,7 +13,8 @@ import (
        "testing"

        "github.com/stretchr/testify/require"
-       funcE "github.com/tetratelabs/func-e/api"
+       func_e "github.com/tetratelabs/func-e"
+       "github.com/tetratelabs/func-e/api"

        egv1a1 "github.com/envoyproxy/gateway/api/v1alpha1"
        "github.com/envoyproxy/gateway/internal/crypto"
@@ -91,7 +92,7 @@ func TestInfraCreateProxy(t *testing.T) {
 func TestInfra_runEnvoy_stopEnvoy(t *testing.T) {
        tmpdir := t.TempDir()
        // Ensures that all the required binaries are available.
-       err := funcE.Run(context.Background(), []string{"--version"}, funcE.HomeDir(tmpdir))
+       err := func_e.Run(context.Background(), []string{"--version"}, api.HomeDir(tmpdir))
        require.NoError(t, err)

        i := &Infra{proxyContextMap: make(map[string]*proxyContext), HomeDir: tmpdir}
```


api/run.go
```diff
@@ -0,0 +1,77 @@
...
+	return func(o *opts.RunOpts) {
+		o.EnvoyErr = w
+	}
+}
+
+// RunOption is a configuration for RunFunc.
+//
+// Note: None of these default to values read from OS environment variables.
+// If you wish to introduce such behavior, populate them in calling code.
+type RunOption func(*opts.RunOpts)
```
codefromthecrypt, line 68: this uses an internal type which is a trick to effectively hide the exported fields

codefromthecrypt:
once this is merged and after next release we need to fix [homebrew](https://github.com/Homebrew/homebrew-core/blob/983c1bff73a733f1b325307a11f8bf974585702e/Formula/f/func-e.rb) to `system "go", "build", *std_go_args(ldflags:), "./cmd/func-e"` (add trailing `, "./cmd/func-e"`)

PR #480, opened 2025-09-18
deps: updates to go 1.25 and latest linters and deps

codefromthecrypt:


codefromthecrypt:
reverted Hugo until we upgrade the netlify runner to latest ubuntu

mathetake (approved):

PR #491, opened 2026-04-29
Adds api.HTTPTransport and transient retry of network glitch

codefromthecrypt:
**api.HTTPTransport**: is a new `RunOption` for tests, allowing control over all HTTP tasks such as getting versions or hitting the admin server. This is threaded through everything and internally used to port tests to truly in-memory via `synctest.Test`

**Transient retry**: `httpGet` (used for downloading envoy versions and envoy) now retries once on dial/TLS failures so a single network glitch doesn't fail the whole operation.

**Undeprecate FUNC_E_HOME**: XDS is great and all, but it is annoying to have to set several at a time in scenarios like tests or where you just need a temporary root.

**Renovations**: Along with above, Go 1.26 and all library deps are updated as well as the lint process. 

Here's an example synctest

```go
// TestRun shows func-e works inside synctest.Test without real network I/O.
func TestRun(t *testing.T) {
	synctest.Test(t, func(t *testing.T) {
		// Pipe-backed server keeps all I/O in-process, compatible with synctest.
		var actualURL *url.URL
		ts := httptest.NewServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			actualURL = r.URL
			w.WriteHeader(http.StatusNotFound)
		}))

		// Route func-e's HTTP traffic through the test server via its transport.
		versionsURL := ts.URL + "/envoy-versions.json"
		err := func_e.Run(t.Context(), []string{"--config-yaml", "foo"},
			api.HomeDir(t.TempDir()), api.EnvoyVersionsURL(versionsURL), api.HTTPTransport(ts.Client().Transport))
		require.Error(t, err)
		require.Equal(t, "/envoy-versions.json", actualURL.String())
	})
}
```


internal/test/httptest/http.go
```diff
@@ -0,0 +1,138 @@
...
+
+	internalapi "github.com/tetratelabs/func-e/internal/api"
+)
+
+// Server is re-exported so callers importing this package don't also need to
+// import the stdlib net/http/httptest just to name the return type.
+type Server = httptest.Server
+
+// NewServer starts an httptest server backed by net.Pipe.
+func NewServer(t *testing.T, handler http.Handler) *httptest.Server {
```
codefromthecrypt, line 21: this is the pipe replacement for httptest.NewServer used internally.

internal/envoy/http_test.go
```diff
@@ -4,26 +4,165 @@
...
+
+		defer res.Body.Close() //nolint:errcheck
+		require.Equal(t, http.StatusOK, res.StatusCode)
+		require.Equal(t, http.MethodGet, actualMethod)
+		require.Equal(t, "/", actualPath)
+		require.Equal(t, devUserAgent, actualUserAgent)
+	})
+}
+
+func TestHttpGet_RetryDecisions(t *testing.T) {
```
codefromthecrypt, line 38: new test that is interesting as you can do polling stuff in synctest

codefromthecrypt (commented):
notes

internal/cmd/run_test.go
```diff
@@ -9,169 +9,199 @@ import (
...
-	go func() {
-		scanner := bufio.NewScanner(stderrReader)
-		for scanner.Scan() {
-			if strings.Contains(scanner.Text(), "starting main dispatch loop") {
-				cancel() // interrupts the child func-e process
-				return
+	synctest.Test(t, func(t *testing.T) {
+		o := setupTest(t)
+
+		// Use a fake Envoy binary
```
codefromthecrypt: synctest.Test waits for bubble goroutines to exit, not for idle, so real subprocess I/O is fine.

internal/envoy/http.go
```diff
@@ -5,34 +5,63 @@ package envoy
...
+		resp, err := client.Do(req)
+
+		if resp != nil || err == nil || ctx.Err() != nil || !isNetError(err) || attempt > 0 {
+			return resp, err
+		}
+
+		// Budget the retry delay from the remaining deadline to avoid wasting
+		// it on sleep (capped at 1s when there's no deadline).
+		delay := time.Second
+		if deadline, ok := ctx.Deadline(); ok {
```
codefromthecrypt: rewrote the retry to deadline bounded

codefromthecrypt:
this also added a deadline bounded retry because archive-envoy flakes (you can see it in tests) which are annoying and rather than make the tests retry, just do one here.

e2e/func-e_test.go
```diff
@@ -63,14 +65,14 @@ func readOrBuildFuncEBin() error {
 }
 
 // funcEExec is a temporary adapter for e2e tests except run.
-func funcEExec(ctx context.Context, args ...string) (string, string, error) {
+func funcEExec(ctx context.Context, args ...string) (stdout, stderr string, err error) {
```
anuraaga, line 68: The new named returns don't seem to be used. Nor do I see any reason to edit this function unless my eyes are failing me

internal/cmd/app.go
```diff
@@ -4,156 +4,227 @@
...
-		args := c.Args()
-		if args.Present() {
-			for _, cmd := range c.App.Commands {
-				if cmd.Name == args.First() {
-					cli.HelpPrinter(c.App.Writer, cmd.CustomHelpTemplate, cmd)
-					return nil
-				}
+// parse creates a Kong parser and parses the CLI args. When Kong requests an
+// exit (--help, --version, or a fatal parse error), the returned error is an
+// *ExitError; Kong has already printed any output. We translate Kong's exit
```
anuraaga: Not doing any research but this sort of trick does make me wonder about kong as a library choice

internal/cmd/man_test.go
```diff
@@ -4,23 +4,95 @@
...
-	expected, err := app.ToMan()
-	require.NoError(t, err)
+	expected := generateManPage()
 
 	actual, err := os.ReadFile(siteManpageFile)
 	require.NoError(t, err)
 	require.Equal(t, expected, string(actual))
 }
+
+func generateManPage() string {
```
anuraaga: Actually testing the man page is definitely an improvement. I was still a surprised by generating it in the test - can't we inline a single string? I may have missed it but don't see anything dynamic in that flags loop

internal/cmd/app_test.go
```diff
@@ -62,12 +63,12 @@ func TestHomeDir(t *testing.T) {
 	tests := []testCase{ // we don't test default as that depends on the runtime env
 		{
 			name:     "default",
-			args:     []string{"func-e"},
-			expected: filepath.Join(u.HomeDir, ".local/share/func-e"),
+			args:     []string{"which"},
```
anuraaga: Just to confirm, previously it defaulted to which, now it's explicit?

internal/cmd/app.go
```diff
@@ -4,156 +4,227 @@
...
+		clear func()
+	}{
+		{cli.ConfigHome, func() { o.ConfigHome = "" }},
+		{cli.DataHome, func() { o.DataHome = "" }},
+		{cli.StateHome, func() { o.StateHome = "" }},
+		{cli.RuntimeDir, func() { o.RuntimeDir = "" }},
+		{cli.Platform, func() { o.Platform = "" }},
+	} {
+		if p.cli != "" {
+			p.clear()
```
anuraaga: Maybe too opinionated but this loop doesn't look Go to me.

If inlining several ifs is too bad, I would create a helper accepting a `flag string, opt *string`

internal/cmd/app.go
```diff
@@ -4,156 +4,227 @@
...
-		// Emit deprecation warning if $FUNC_E_HOME is set
-		if homeDir != "" {
-			fmt.Fprintln(c.App.ErrWriter, "WARNING: $FUNC_E_HOME (--home-dir) is deprecated and will be removed in a future version.") //nolint:errcheck
-			fmt.Fprintln(c.App.ErrWriter, "Please use --config-home, --data-home, --state-home or --runtime-dir instead.")             //nolint:errcheck
-		}
+// ExitError signals that Kong requested process termination with the given
+// code (e.g. --help, --version, or a parse error Kong already printed). The
+// caller should translate it into a process exit code.
+type ExitError struct {
+	Code int
```
anuraaga: Not sure if there is a doc or link that could help understand what this actually gets set to

anuraaga (approved):
FWIW don't think I see the boilerplate improvement from kong but will assume it's there.

codefromthecrypt:
ps thanks for the feedback @anuraaga I owe a cleanup and a better description as Kong was more about urfave upgrade adding more deps etc than about boilerplate.. also I didn't do a great job at cruft sweep either!

e2e/func-e_test.go
codefromthecrypt: lint fail basically. have more linters and they don't like not naming this

codefromthecrypt:
@anuraaga ps I have reverted kong as honestly I think it isn't worth doing something like that in this PR. we need the real before after of CURRENT urfave not conjecture. almost done with the urfave update.

internal/envoy/http.go
```diff
@@ -5,34 +5,63 @@ package envoy
...
+			if half := time.Until(deadline) / 2; half < delay {
```
anuraaga: I think I'm reading right, that the delay gets effectively exponentially halfed each attempt. In practice there should be enough request delay to make progress but being the opposite of most retries it's frying my brain.

If keeping things simple, how about a fixed delay by just dividing deadline by attempts with cap before the loop?

Otherwise no shame in reaching for the backoff dep

e2e/func-e_test.go
anuraaga: Ah got it I guess it must be when there are multiple returns with the same type

api/run.go
```diff
@@ -125,6 +126,19 @@ func EnvoyErr(w io.Writer) RunOption {
 	}
 }
 
+// HTTPClientFunc creates HTTP clients used during a run.
+type HTTPClientFunc = api.HTTPClientFunc
+
+// DefaultHTTPClient returns http.DefaultClient.
+func DefaultHTTPClient() *http.Client { return http.DefaultClient }
```
anuraaga: Just a double check `http.Client` is better than `http.RoundTripper` for this?

anuraaga (approved):

api/run.go
codefromthecrypt: no definintely you are right. just because it is needed internally doesn't mean the caller should think in terms of http.Client. Minimally yeah we need the round tripper, and also we don't need to re-create the client so many times either.

internal/envoy/http.go
codefromthecrypt: TL;DR; I can just make it sleep 1s regardless and if it overruns it does.. It is unlikely any task can complete if there's only a second or two total deadline.

---

I over solved it because I was like looking at cool code ;)

So this was doing a "single retry" but I got carried away with being fancy and that made it both look like it was a delay loop (it only allows once), and hid the why of the halving thing, which I admit was maybe too clever. ( don't sleep the entire remaining time, leave some time for the actual try)

Saving the better commented code here for a rainy day, but keeping the next change set dumb.

```go
// httpGet GETs rawURL with a User-Agent header and one retry on transient network error.
func httpGet(ctx context.Context, client *http.Client, rawURL, ua string) (*http.Response, error) {
	// A loop avoids duplicating the request setup when a retry is needed.
	for attempt := 0; ; attempt++ {
		// #nosec -> url can be anywhere by design
		req, err := http.NewRequestWithContext(ctx, http.MethodGet, rawURL, http.NoBody)
		if err != nil {
			return nil, err
		}
		req.Header.Add(userAgentHeader, ua)

		resp, err := client.Do(req)

		// Return unless this is the first attempt and it hit a transient network error.
		if resp != nil || err == nil || ctx.Err() != nil || !isNetError(err) || attempt > 0 {
			return resp, err
		}

		// Reserve half the remaining deadline for the retry itself.
		retryBudget := time.Second
		if deadline, ok := ctx.Deadline(); ok {
			if delay := time.Until(deadline) / 2; delay < retryBudget {
				retryBudget = delay
			}
		}
		if retryBudget <= 0 {
			return resp, err
		}

		// Wait for the retry budget to elapse, or bail if the context is canceled.
		select {
		case <-ctx.Done():
			return nil, ctx.Err()
		case <-time.After(retryBudget):
		}
	}
}
```

anuraaga: Hah got it, yeah simplifying sounds good. That loop quacks way too much like a retry loop, completely missed the check of attempt more than 0 ;)

internal/admin/admin.go
```diff
@@ -21,43 +21,79 @@ import (
...
+	ServerAddr      = "127.0.0.1:9901"
+	AddressPathFlag = "--admin-address-path"
+	live            = "live"
+)
 
 // NewAdminClient creates an AdminClient by polling for the admin port at
 // adminAddressPath.
-func NewAdminClient(ctx context.Context, adminAddressPath string) (internalapi.AdminClient, error) {
-	// Block until admin port is available
+func NewAdminClient(ctx context.Context, client *http.Client, adminAddressPath string) (internalapi.AdminClient, error) {
```
anuraaga, line 32: Heh thanks for finding the proper solution accepting the client here

internal/envoy/http.go
```diff
@@ -5,34 +5,49 @@ package envoy
...
-	// #nosec -> url can be anywhere by design
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
-	if err != nil {
-		return nil, err
+const userAgentHeader = "User-Agent"
+
+// httpGet GETs rawURL with a User-Agent header and one retry on transient network error.
+func httpGet(ctx context.Context, client *http.Client, rawURL, ua string) (*http.Response, error) {
+	// A loop avoids duplicating the request setup when a retry is needed.
+	for attempt := 0; ; attempt++ {
```
anuraaga: Mmm, I was definitely hoping to lose the loop. If a private helper can't help with setup concern, then we chalk this up to lack of do while in Go I guess

anuraaga (approved):

codefromthecrypt:
@anuraaga thanks for the feedback, you especially saved 2 things

1. suspicious motivation to switch to kong. I freaked out about urfave adding a dep but really it isn't a big deal it is just being more modular. Not worth it.
2. while internally http.Client is the right thing to pass around, externally, the only thing required really is http.RoundTripper and now the example shows why!

internal/envoy/http.go
codefromthecrypt: I was just annoyed at having to extract helpers and yeah maybe this is not fighting the right battle. I'll extract

codefromthecrypt: unrolled!

internal/envoy/http.go
```diff
@@ -5,34 +5,51 @@ package envoy
...
-func httpGet(ctx context.Context, client *http.Client, url string, p version.Platform, v string) (*http.Response, error) {
+const userAgentHeader = "User-Agent"
+
+// httpGet GETs rawURL with a User-Agent header and one retry on transient network error.
+func httpGet(ctx context.Context, client *http.Client, rawURL, ua string) (*http.Response, error) {
 	// #nosec -> url can be anywhere by design
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
-	if err != nil {
-		return nil, err
+	get := func() (*http.Response, error) {
```
anuraaga, line 20: I often forget about internal functions like this looks great. Thanks!

anuraaga (approved):

PR #493, opened 2026-05-06
admin: better find the Envoy child process

nacx:
The code in the admin client that looked for the Envoy child process always returned the first child. This causes problems when func-e is used as a library in e2e tests or similar, and other processes are started as well (I faced issues with this while writing tests in the Built On Envoy project).

Since func-e always sets the `--admin-address-path` when starting Envoy, we can look for a child process that has that flag instead of always picking the first one.


internal/admin/admin.go
```diff
@@ -248,24 +248,35 @@ LOOP:
...
-			// Assume the first child is the Envoy process
-			envoyProc = children[0]
+			// Find the first child whose command line contains the
+			// --admin-address-path flag, which is guaranteed to be present
+			// in every func-e-managed Envoy process.
+			var envoyCmdline []string
+			for _, child := range children {
+				cmdline, cmdErr := child.CmdlineSliceWithContext(ctx)
+				if cmdErr != nil {
+					continue
```
anuraaga, line 258: Not sure it's ok to squash this error

anuraaga (commented):
I noticed boe has a workaround to start extprocs after envoy

https://github.com/tetratelabs/built-on-envoy/blob/24fc7bf2359ee1f5724d3bb4bb63d16d4e17c694/cli/internal/envoy/runner.go#L125

Is the intention to remove that restriction, or is it a different case? If so, can you share the actual case you're running into?

> func-e is used as a library in e2e tests or similar

If I read this as using `Run` directly in an e2e test, I'd expect it to use a startup hook to get the admin client, not poll.

codefromthecrypt:
I agree that func-e is using a simplistic prod path for docker, which in normal case shouldn't collide, even in tests.

--admin-address-path is not exactly the best way as that's a valid arg someone can pass to envoy with or without func-e

If we do something different, it would be a selector you can use in tests and then normal HEALTHCHECK can avoid using.

func-e historically used Envoy’s `--` ignore-rest behavior to add markers. Anyone calling in can add their own markers after `--`. So, if we need precision, esp for non-prod scenarios, I might suggest we just pass in a cmd expression that can be overridden, and not assume it is needed in normal case. So, in your tests, you can pass in `-- boe-test-foo` as the trailing args or some other signature?

How does that sound?

nacx:
>I noticed boe has a workaround to start extprocs after envoy
https://github.com/tetratelabs/built-on-envoy/blob/24fc7bf2359ee1f5724d3bb4bb63d16d4e17c694/cli/internal/envoy/runner.go#L125
Is the intention to remove that restriction, or is it a different case? If so, can you share the actual case you're running into?

Those lines are to overcome a side effect of how func-e is being used in boe and its e2e tests. `boe` now has support for ext_proc extensions, and when running `boe run --extension <extproc extension>`, the BOE process will start:
* The extproc as a child process.
* func-e.Run to start Envoy.

In the `boe` e2e tests, we run the `boe` command to configure stuff and actively wait on Envoy to be ready. To do that, we instantiate the AdminClient with the PID of the "boe" process (the only info we have at that time), and this is where the issue codes are, because that AdminClient would find that the first child process of "boe" is the extproc, not Envoy.

We workaround this by making sure Envoy starts first, and deferring the start of the extproc (inside boe) to the adminclient startup hook. So we're not hitting this problem anymore. I agree that this is a particular setup and probably something that is not a func-e concern, but ultimately the goal was to provide a more robust way of identifying the Envoy process to make it more easily "just work" in any setup.

>func-e historically used Envoy’s -- ignore-rest behavior to add markers. Anyone calling in can add their own markers after --. So, if we need precision, esp for non-prod scenarios, I might suggest we just pass in a cmd expression that can be overridden, and not assume it is needed in normal case. So, in your tests, you can pass in -- boe-test-foo as the trailing args or some other signature?
How does that sound?

Let me move this PR to draft and give a try to this approach.

codefromthecrypt:
func-e already has a "run id" (fixed on docker) so I think this is probably the best and least strange way out. AdminClient can be specific to that. I can help get this to work.

codefromthecrypt:
https://github.com/tetratelabs/func-e/pull/505 should do it I think

PR #494, opened 2026-05-06
chore: use go tools instead of building on-demand

nacx:
Uses native `go tool` instead of using `go run` for tools used during the build, which allows for better version pinning.
The main caveat was polluting the main `go.mod`, but this PR addresses that by separating the tools in a separate `tools/go.mod`.

Opening as a draft to discuss, happy to dismiss this PR if it is preferred to keep the current approach.


nacx:
Failing tests are fixed in https://github.com/tetratelabs/func-e/pull/493

codefromthecrypt:
not really keen on this, `go tool` has a long paper trail of issues. I'm not precisely sure what we are trying to achieve on this? Is the latency of the first-time run the goal?

https://github.com/bufbuild/buf/issues/3957

codefromthecrypt:
I'll contribute a RATIONALE.md on this as I think we need to know the difference between "wow it works today, take a photo!!" and what we do when the deps conflict again, and why all that. then, it is kindof immortalized, not just the change but what happens later.

codefromthecrypt:
@anuraaga I filed the rationale here. I think for now, we're good. It is just a little dancing to make sure `go.work`... works. wdyt?

RATIONALE.md
```diff
@@ -49,5 +49,23 @@ their own directories and co-mingle its configuration and logs with those
 of func-e when it runs Envoy (the gateway process). It also allows Docker to
 export `FUNC_E_RUN_ID=0` to aid in location of key files.
 
+## Why tools/go.mod?
```
anuraaga, line 52: I guess this isn't clear on why use go tool, it is only a passing mention of the sums.

RATIONALE.md
```diff
@@ -49,5 +49,23 @@ their own directories and co-mingle its configuration and logs with those
...
+`go tool` lets us run things like linters and hugo without a platform install.
+We keep them in `tools/go.mod` instead of the main `go.mod` because func-e is
+also imported as a library; tool dependencies should not leak into that graph.
+
+This replaces the former `go run package@version` process, where versions
+lived in Makefile commands instead of a normal checked-in module graph.
+
+The biggest tradeoff from our old process is tools are not isolated from each
+other. Hugo, golangci-lint, nfpm, etc. share one dependency graph and can
+revlock each other in the future.
```
anuraaga, line 63: Maybe the operation for this case, i.e., either "just don't update" or split off another go.mod for the specific tool.

codefromthecrypt (approved):
I'm going to approve so the lockup with the other PR works out. Happy for any post merge suggestions.

anuraaga (approved):
As an anti-Go tool guy, I wouldn't mind something like the Buf docs about the implications of using different dependencies than the actual released tool, but also understand if omitting it ;)

https://buf.build/docs/cli/installation/#protoc-and-plugins-not-included

PR #504, opened 2026-05-07
fix 'make site' when the theme submodule has never been cloned

nacx:
Fixes the `make site` target when the theme submodule has never been cloned.


codefromthecrypt (approved):
thx

PR #505, opened 2026-05-08
Integrates `RunID` and `HTTPTransport` to `NewAdminClient`

codefromthecrypt:
This integrates `api.RunID()` to disambiguate sub-processes in `NewAdminClient`.

Before, we looked for the first subprocess, which could be anything, such as an ext-proc process. This is a routine use case of built-on-envoy (BOE). Moreover, there's nothing preventing func-e from starting multiple envoys. It was designed to do this. If you can't know which subprocess is envoy or which envoy you are talking about, any AdminClient functionality is thwarted.

To solve this, we add a marker `--run-id` after an `ignore_after` sentinel envoy supports (`--`). This doesn't conflict with envoy args and anyway envoy args precede (`--`). We use `--run-id`, not `--func-e-run-id` both for simplicity and to not brand embedders like ai-gateway as func-e. There's intentionally no config for this to prevent sprawl and also not permanently commit to the marker should we want to change it later.

With this marker in, we have a good idea both what is envoy and what is our envoy. If no-one adds `api.RunID()` it is already stronger than before as we at least scan for the named marker. If they do, it narrows. So, BOE for example, should be safe without even using `api.RunID()` until such time they do like mini-fleet testing or something where multiple envoys are launched.

However, it could be tightened now, with no negative impact:
```go
type Healthcheck struct {
	RunID string `name:"run-id" env:"BOE_RUN_ID" default:"0" hidden:""`
}

func (h *Healthcheck) Run(ctx context.Context) error {
	return healthcheck(ctx, io.Discard, os.Stderr, h.RunID)
}

func doHealthcheck(ctx context.Context, boePid int, runID string, logger *slog.Logger) error {
	adminClient, err := admin.NewAdminClient(ctx, boePid, api.RunID(runID))
	if err != nil {
		logger.Error("Failed to find Envoy admin server", "error", err)
		return err
	}
	return adminClient.IsReady(ctx)
}
```

As usual some cleanupts were warranted. First, even though this function is experimental, it is already coded to multiple projects, so breaking `NewAdminClient` wouldn't be helpful. That's why we accept `options ...api.RunOption`. Even though some are not honored, doing this gives you control of `HTTPTransport` for free, helpful in tests and obs.

Closes #493

### DIFF
--- a/internal/admin/admin.go
+++ b/internal/admin/admin.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
@@ -24,8 +25,12 @@ import (
 const (
 	ServerAddr      = "127.0.0.1:9901"
 	AddressPathFlag = "--admin-address-path"
+	runIDFlag       = "--run-id"
 	live            = "live"
+	pollInterval    = 50 * time.Millisecond
 )
+
+var errMultipleEnvoyProcesses = fmt.Errorf("multiple Envoy processes found; set %s to disambiguate", runIDFlag)
 
 // NewAdminClient creates an AdminClient by polling for the admin port at
 // adminAddressPath.
@@ -192,7 +197,7 @@ func (c *adminClient) Get(ctx context.Context, path string) ([]byte, error) {
 // pollAdminAddressPathForPort polls for the admin-address.txt file.
 // It returns the admin port number or an error if the timeout is reached.
 func pollAdminAddressPathForPort(ctx context.Context, adminAddressPath string) (int, error) {
-	ticker := time.NewTicker(50 * time.Millisecond)
+	ticker := time.NewTicker(pollInterval)
 	defer ticker.Stop()
 
 	var adminAddr string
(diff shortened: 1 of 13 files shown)


internal/admin/admin.go
```diff
@@ -221,65 +226,166 @@ LOOP:
...
 
 	// Shell wrappers expose the wrapped command as one argv entry. Keep this
 	// fallback after the argv-preserving scan so direct args can contain spaces.
 	if len(cmdline) >= 3 && cmdline[1] == "-c" {
 		fields := strings.Fields(cmdline[2])
-		for i, arg := range fields {
-			if arg == flag && i+1 < len(fields) && fields[i+1] != "" {
+		for i := range len(fields) {
+			arg := fields[i]
+			if arg == internalapi.ArgsIgnoreRest {
```
anuraaga, line 268: Not sure the variable is making things clearer vs just inlining `--`. Or otherwise, maybe `EndOfEnvoyArgs` would be a clear name, ignore rest is in terms of Envoy but here we're func-e

internal/admin/admin.go
```diff
@@ -221,65 +226,166 @@ LOOP:
...
+		}
+	}
+
+	return "", fmt.Errorf("%s not found in command line", AddressPathFlag)
+}
+
+// extractRunID returns the last match, so the func-e-appended value after [internalapi.ArgsIgnoreRest] wins.
+func extractRunID(cmdline []string) (string, error) {
+	var runID string
+	for i := 0; i < len(cmdline); i++ {
```
anuraaga, line 288: Optional, can consider an internal function `func extract(fields []string) string` to call in both cases though I guess there may be other places like that too

anuraaga (approved):

codefromthecrypt:
oops sorry I shouldn't have done auto-merge on this @anuraaga I'll weave the feedback into a next

internal/admin/admin.go
codefromthecrypt: https://github.com/tetratelabs/func-e/pull/506 should get it

internal/admin/admin.go
codefromthecrypt: https://github.com/tetratelabs/func-e/pull/506 should get it

PR #507, opened 2026-05-08
Installs go so that release e2e can run

codefromthecrypt:


codefromthecrypt:
macOS release e2e of the release binary failed as go wasn't installed.

anuraaga (approved):

PR #509, opened 2026-05-15
Use yesterday's envoy main build via the "dev" version

codefromthecrypt:
Before this, testing pre-release Envoy with func-e meant compiling it yourself or using Docker, which only worked on Linux. Now you can install and run dev builds on all platforms the same way you would any tagged version.

```bash
$ ENVOY_VERSION=dev func-e run --version                                                              
downloading https://archive.tetratelabs.io/envoy/download/dev/envoy-dev-darwin-arm64.tar.xz
starting: /Users/codefromthecrypt/.local/share/func-e/envoy-versions/dev/bin/envoy with logs in /Users/codefromthecrypt/.local/state/func-e/envoy-runs/20260518_170024_807

/Users/codefromthecrypt/.local/share/func-e/envoy-versions/dev/bin/envoy  version: abbadd905fa486ff1085cf3bbe4f3b73eb6dd8e0/1.39.0-dev/Clean/RELEASE/BoringSSL

$ func-e versions -a                                                                                  
  dev 2026-05-18 (a8d396eb)
  1.38.0 2026-04-23
  1.37.2 2026-04-10
-- snip--
```

This works because envoy's version manifest was recently updated with a special "dev" key backed by a [daily pipeline job](https://github.com/tetratelabs/archive-envoy/pull/70), which archives linux from envoy's docker image, and builds macos from the same SHA.

## Updating the "dev" build with the "dev-latest" alias.

`dev` installs on demand like all other versions and stays put once pulled. The alias `dev-latest` compares the remote release date against the local install and re-downloads only when the build has changed. `func-e use dev-latest` persists "dev" in the version file, so it is a one-shot refresh, not a stored preference.

## Why Use An Envoy "dev" Build?

Envoy releases are worth the wait, but they land about once a quarter. The dev build is for the gap between “merged” and “released”: you can try current Envoy mainline without building Envoy yourself or relying on a Linux-only Docker workflow.

For example, Envoy 1.39-dev already has new dynamic-module support for request-aware upstream selection. A dynamic module can parse a request, write `selected_backend` or `selected_endpoint` into request state, and have the cluster read it while choosing the upstream host. That removes the need to pass routing decisions through synthetic headers  or route re-selection.

Use dev when you need to test an unreleased feature, validate upgrade impact early, or give feedback before the next Envoy release is cut. Use tagged versions for normal production runs.

### DIFF
--- a/e2e/main_test.go
+++ b/e2e/main_test.go
@@ -6,20 +6,9 @@ package e2e
 import (
 	"context"
 	"fmt"
-	"io"
-	"net/http"
-	"net/http/httptest"
 	"os"
-	"strings"
 	"testing"
 )
-
-const (
-	envoyVersionsURLEnvKey = "ENVOY_VERSIONS_URL"
-	envoyVersionsJSON      = "envoy-versions.json"
-)
-
-var expectedMockHeaders = map[string]string{"User-Agent": "func-e/dev"}
 
 // TestMain ensures the "func-e" binary is valid.
 func TestMain(m *testing.M) {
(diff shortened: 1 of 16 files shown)


e2e/main_test.go
```diff
@@ -29,55 +18,13 @@ func TestMain(m *testing.M) {
...
 	os.Exit(m.Run())
 }
 
 func exitOnInvalidBinary(err error) {
 	fmt.Fprintf(os.Stderr, `failed to start e2e tests due to an invalid "func-e" binary: %v\n`, err)
 	os.Exit(1)
 }
-
-// mockEnvoyVersionsServer ensures envoyVersionsURLEnvKey is set appropriately, so that non-release versions can see
-// changes to local envoyVersionsJSON.
```
codefromthecrypt, line 55: no longer needed really and was rarely used except to use a pre-release version of envoy

anuraaga (approved):

PR #513, opened 2026-05-19
Reverts to Tools.mk for go run instead of go tool

codefromthecrypt:
Before this, tools lived in tools/go.mod via `go tool`. This forced all tools into one dependency graph where hugo, golangci-lint, and nfpm could revlock each other, and required `GOWORK=off` on every invocation because Go rejects `-modfile` in workspace mode. Dependabot also raised PRs against transitive deps of tools (aws-sdk, otel, go-git) that had nothing to do with func-e.

Now, tools are pinned in Tools.mk and run via `go run package@version`, the approach used before PR #494. Each tool resolves its own deps independently, tools/go.sum (426 lines of transitive tool deps) is gone, and the Makefile no longer needs the GOWORK workaround.


anuraaga (approved):

codefromthecrypt:
ironic that transitive go.mod showed up on this PR basically the tools interfered and now when separate the linting is more strict ;)

PR #514, opened 2026-05-19
Automates PR to update last known Envoy version

### LINKED ISSUE #25: Automate a PR to update func-e
On release of next version, we should be able to do an update to func-e. Notes:

* Not all versions will change the last known version. For example, if 1.20.0 is latest and a release here adds a patch to 1.19, there's noop and should be no pull request.
* There is some variable substitution needed ex. https://github.com/tetratelabs/func-e/pull/399
* Testing such a change would be best done against someone's fork, not func-e itself, and it may be hard to test anyway.

codefromthecrypt:
Before this, bumping the last known Envoy version required a manual PR touching 5 files every time archive-envoy published a new release. Now a scheduled workflow polls envoy-versions.json and opens a PR when the latest version changes. A numeric version comparison safeguards against the last known version accidentally going backwards.

Fixes https://github.com/tetratelabs/archive-envoy/issues/25


codefromthecrypt:
tested via https://github.com/codefromthecrypt/func-e/pulls?q=is%3Apr+is%3Aclosed

PR #521, opened 2026-08-02
fix: stage Envoy downloads and only move verified installs into place

CodePrometheus:
If a download is interrupted — a dropped connection, Ctrl-C, a full disk — func-e leaves a partial Envoy binary at the install path, because the tarball is unarchived directly there and the SHA-256 sum is only checked afterwards. 
Every later run stats that path, reports "<version> is already downloaded", and hands the truncated binary to the caller, which fails at exec time with no indication that the install is the problem. There is no way to recover short of manually deleting the version directory, since func-e never re-downloads a version it believes it has.

Downloads now unarchive into a staging directory next to the install path and are moved into place only after the checksum matches, so the install path either holds a complete install or nothing at all. A failed download is cleaned up and the next run downloads again.


CodePrometheus:
Gentle ping~ @codefromthecrypt  mind taking a look when you have time?

nacx (approved):
LGTM. Thank you!

